PR #1, opened 2016-05-22
Crypto

Theodus:
Add an RFC proposing the addition of a Cryptography package to the standard library

### DIFF
--- a/text/0000-crypto.md
+++ b/text/0000-crypto.md
@@ -15,11 +15,11 @@ Cryptography is a vital tool in the areas of authentication, confidentiality, da
 
 *Security must be the first priority of this package.*
 
-The package will include the following interfaces:
+This package will use the OpenSSL crypto library through the C FFI. The following API will be used:
 
 - HashFunc
 
-	provides a hash() function that produces a fixed-length byte sequence based on the input sequence.
+    interface providing a hash() function that produces a fixed-length byte sequence based on the input sequence.
     
     Example use:
     	
@@ -28,67 +28,26 @@ The package will include the following interfaces:
     	
         [80 216 88 224 152 94 204 127 96 65 138 175 12 197 171 88 127 66 194 87 10 136 64 149 169 232 204 172 208 246 84 92]
 
-- Cypher
 
-	provides encrypt() and decrypt() functions that both produce a byte sequence from the input sequence. If the result of encrypting a given byte sequence is then passed into the decrypt function, the result should be the given byte sequence.
-    
-    primitives must be provided for the various Block cypher modes of operation: CBC, ECB, CFB, OFB, and CTR. These will be passed into the constructors of any block cyphers along with the encryption key.
-    
-    Example use:
-    	
-        let des = DES("01234567", ECB)
-        let enc = des.encrypt("abcdefgh")
-        des.decrypt(enc) // returns "abcdefgh"
-
-- Asymmetric
-
-    An Asymmetric constructor will take in a cryptographically secure pseudorandom number generator. It can then produce a private and a public key.
-    
-    The PrivateKey will have the ability to encrypt data, decrypt data, sign data, verify a signature, and create a public key.
-    
-    The PublicKey will have the ability to encrypt data and verify a signature.
-    
-    Example use:
-    	
-        let rand = CryptoRandom()
-       	let rsa = RSA(rand)
-        let private_key = rsa.private_key()
-        let public_key = rsa.public_key()
-        
-        let enc = public_key.encrypt("abcdefgh")
-        private_key.decrypt(enc) // returns "abcdefgh"
-        
-        hash = SHA256.hash("abcdefgh")
-        let signature = private_key.sign(hash, "")
-        public_key.verify(hash, signature) // returns true
-        
-- Symmetric
-    
-    A Symmetric constructor will take in a cryptographically secure pseudorandom number generator. It can then produce a single key that may encrypt and decrypt data.
+- ConstantTimeCompare
+ 
+    primitive where the apply function takes in two byteseq arguments and returns a Bool, true if they are equal and false otherwise.
 
     Example use:
     	
-    	let rand = CryptoRandom()
-       	let key = AES(rand).key()
-       	let enc = key.encrypt("abcdefgh")
-        key.decrypt(enc) // returns "abcdefgh"
-       	
-The package must also allow the user to do the following:
-
-- Perform constant time comparisons of byte sequences
-- Create a cryptographically secure pseudorandom number generators
-
+    	let s1 = [U8(1), U8(2), U8(3), U8(4), U8(5)]
+    	let s2 = [U8(5), U8(4), U8(3), U8(2), U8(1)]
+    	ConstantTimeCompare(s1, s1) // returns true
+    	ConstantTimeCompare(s1, s2) // returns false
+   
+   see also: [Golang implementation](https://golang.org/src/crypto/subtle/constant_time.go?s=490:531#L2)
 
 # How We Teach This
 
 This package should be used with some knowledge of basic concepts such as:
 
 - Hash functions - can be used to calculate the checksum of some data. It can be used in digital signatures and authentication. e.g. SHA-256 or MD5
-- Encryption algorithms - take some byte sequence as input and produce a cipher byte sequence using a variable key. You have 2 types of ciphers: block and stream.
-  - Block ciphers work on blocks of a fixed size (8 or 16 bytes). e.g. DES
-  - Stream ciphers work byte-by-byte. Knowing the key, you can decrypt the cipher. e.g. XOR
-- Public-key algorithms - there are two different keys: one to encrypt and one to decrypt. You only need to share the encryption key and only you can decrypt the message with your private decryption key. e.g. RSA
-- Symmetric-key algorithms - the same key is used for both encryption and decryption. The keys maintain a private information link between two or more parties. e.g. Twofish or Blowfish
+- Constant time comparison - used to prevent [timing attacks](http://crypto.stanford.edu/~dabo/papers/ssl-timing.pdf) with the use of constant-time algorithms
 
 # Drawbacks
 
@@ -100,5 +59,4 @@ The impact of not providing cryptographic security to various Pony applications 
 
 # Unresolved questions
 
-- Should the package be implemented in pure Pony or through FFI with an established library such as OpenSSL? 
-- Inclusion of secret-key encryption
+None
--- a/text/0000-crypto.md
+++ b/text/0000-crypto.md
@@ -17,13 +17,13 @@ Cryptography is a vital tool in the areas of authentication, confidentiality, da
 
 This package will use the OpenSSL crypto library through the C FFI. The following API will be used:
 
-- HashFunc
+- HashFn
 
-    interface providing a hash() function that produces a fixed-length byte sequence based on the input sequence.
+    interface providing an apply() function that produces a fixed-length byte sequence based on the input sequence.
     
     Example use:
     	
-    	SHA256.hash([101 120 97 109 112 108 101])
+    	SHA256([101 120 97 109 112 108 101])
     Return value:
     	
         [80 216 88 224 152 94 204 127 96 65 138 175 12 197 171 88 127 66 194 87 10 136 64 149 169 232 204 172 208 246 84 92]
--- a/text/0000-crypto.md
+++ b/text/0000-crypto.md
@@ -44,14 +44,14 @@ This package will use the OpenSSL crypto library through the C FFI. The followin
    
 - Digest
 
-    class that produces a hash from chunks of input. The hash can be calculated at any time and updated with the update method.
+    class that produces a hash from chunks of input.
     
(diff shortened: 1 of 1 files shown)


Theodus:
I'll close this PR if it's too early to be submitting RFCs :)

SeanTAllen:
@Theodus Go ahead and leave it up. We might change the template or process a little on Wednesday. But hopefully nothing to big. If we change anything, I doubt there would be much for you to change.

jemc:
Haven't done a full review of this yet, but a few early thoughts:

First, this proposal doesn't mention whether this package would be implemented in pure Pony or in using some established shared librar(y/ies) via FFI.  There are pros and cons to weigh here when it comes to encryption in the standard library, so I think it's important to know what we're talking about.

Something seems off in the way that public key cryptography is presented.  First of all, we need to make a distinction between asymmetric and symmetric encryption - this proposal seems to focus only on asymmetric and call it `PublicKey`, leaving no room for a symmetric encryption option on equal footing.  Symmetric encryption requires two keys for every encrypt/decrypt operation (your own private key as well as the public key of the other party), so it needs a different interface to think about - we should have both interfaces in mind up front so that they can be idiomatic together instead of having the other one will be a glued-on afterthought.  Also, it seems like even the `PublicKey` example given is a bit fishy, as it shows encrypting and decrypting with the same key object.  In asymmetric encryption, you should be encrypting with the public key and decrypting with the associated private key, but in this example, that distinction is either missing or its hidden in some way to obscure it from the developer (which is still probably wrong from a developer experience point of view).

jemc:
There's also secret-key encryption, relying on the secrecy of a shared secret between two parties instead of relying on the one-way transformations of private-to-public keys.  Secret key encryption is definitely still relevant for some applications, given that the unidirectionalness of private-to-public key transformations is being [called into question by quantum computing](https://tonyarcieri.com/imperfect-forward-secrecy-the-coming-cryptocalypse).

Theodus:
I will include symmetric-key encryption and make some changes so that the example for public-key encryption is more clear.

tmmcguire:
Some questions:
1. Would this be a top-level package? (I assume it would be part of the Pony distribution.) 
2. Assuming the cryptographic primitives are implemented by a C library, which one? Or, which ones? Is it reasonable to support multiple backends? OpenSSL's [libcrypto](https://www.openssl.org/docs/man1.0.2/crypto/), [NaCl](https://nacl.cr.yp.to/), and [libsodium](https://download.libsodium.org/doc/) are the three that come to mind (and the last two are approximately the same thing). (Then there's [TweetNaCl](https://tweetnacl.cr.yp.to/software.html), from which i can only run screaming.)
   
   The Wikipedia [Comparison of Cryptography Libraries](https://en.wikipedia.org/wiki/Comparison_of_cryptography_libraries) lists what I think is a fairly limited selection.
3. I suspect the API examples in the detailed design are incomplete. See the [Java Cryptography Architecture Reference Guide](http://docs.oracle.com/javase/6/docs/technotes/guides/security/crypto/CryptoSpec.html) for a good way to give yourself nightmares for the rest of your life.

Theodus:
@tmmcguire
1. I think that it should be a top-level package 
2. OpenSSL's crypto is probably the preferred choice since ponyc already depends on libressl
3. This is absolutely true. I am not an expert on Cryptography by any stretch and any suggestions would be appreciated.

jemc:
Since libsodium was mentioned, I'll add that [I've already written a 3rd-party libsodium package for Pony](https://github.com/jemc/pony-sodium).  Granted, it needs some sprucing up, but it works nicely for what it does.

> I am not an expert on Cryptography by any stretch and any suggestions would be appreciated.

Honestly, when it comes to the more complicated side of the cryptography APIs, I would kind of prefer to have someone who is really comfortable with these concepts (and has solid use case "stories" for their usage in mind) to be the one who designs them.  If someone who meets that description is willing to step in and collaborate here, that's great, but if not then perhaps this RFC could make better progress by being reduced in scope to cover a just part of the current scope.

For example, it could be reduced to cover just cryptographic hashing functions.  Then the rest of the features could be added to a "Request for Request for Comment".  I think Rust does this by adding issue tickets to the RFCs repo, then someone comes in to write the RFC that eventually closes the ticket?  That might be an appropriate way to deal with the parts you're not comfortable with - but maybe @SeanTAllen can confirm that this would be the process he had in mind for that workflow.

SeanTAllen:
@jemc I hadn't considered that sort of possibility. It certainly sounds like a good use of issues on this repo.

Theodus:
I think that reducing the scope of this RFC to hashing functions and constant time comparisons would be a good idea. Should I change the RFC to describe just these features and call for the use of OpenSSL?

SeanTAllen:
@Theodus Yes. That makes sense.

Theodus:
Would anyone be opposed to changing byte sequences to String? Performing hash functions on Strings is the most common use case and creating byte sequences from Strings seems a bit unnecessary.

jemc:
@Theodus - they should be able to accept a `ByteSeq` and work on either `String` or `Array[U8]`.

Theodus:
An example of what an implementation may look like can be found at github.com/theodus/pony-crypto

text/0000-crypto.md
```diff
@@ -0,0 +1,62 @@
...
+
+Cryptography is a vital tool in the areas of authentication, confidentiality, data integrity, and nonrepudiation.
+
+# Detailed design
+
+*Security must be the first priority of this package.*
+
+This package will use the OpenSSL crypto library through the C FFI. The following API will be used:
+
+- HashFunc
```
jemc: I think the name `HashFn` would be more idiomatic with some of the types we already have in the standard library.

text/0000-crypto.md
```diff
@@ -0,0 +1,62 @@
...
+
+    interface providing a hash() function that produces a fixed-length byte sequence based on the input sequence.
```
jemc: Given that this interface only requires one method, why not make it use an `apply` function so that a simple `lambda` can easily be substituted?

Theodus:
I think that this RFC is ready for the final comment period.

SeanTAllen:
@Theodus can you send an email to the dev list with a short overview of the RFC as well as a link to?

Thanks.

malthe:
What would be equivalent of (in Python):

``` python
h = hashlib.sha256(a)
h.update(b)
```

Note that `h` is an object and:

``` python
h == hashlib.sha256(a + b)
```

Also, in Python it was decided to use lowercase for these methods (since they're named constructors, returning a generic `HASH[function]` object). I think lowercase would make sense for Pony, too.

Theodus:
What benefits would creating an object have over a single primitive?

malthe:
The benefit is that you can use the `update` method to compute a digest of a string chunk by chunk.

In Java and presumably most other object-oriented languages, a hash output comes from a `MessageDigest` object – which is not to say that the generated native code necessarily reflects this in all situations.

Theodus:
I like the idea of having a single object for all hashes provided by OpenSSL, but I think that @SeanTAllen and/or @jemc should weigh in on this.

jemc:
I think the streaming of chunks is probably a good motivation for using a stateful object that does the hashing.

It looks like OpenSSL provides two different APIs for this - one with a stateful object and chunk support, and another with a simple function using a single input and output buffer.  @Theodus has modeled his API after the simple function OpenSSL APIs, and these will actually have less overhead in the case that you are not _using_ chunked support.  So I wouldn't say we should have chunked support _only_ as @malthe may be suggesting - as that increases the cost of the more common operation of hashing a single buffer.

Also, if we do add a chunked-support API, I wouldn't want to see it be implemented by a single class.  The implementation of that class would have to switch internally on the which hash was specified at creation in practically every function, which should really be a code smell that these should be different classes (or maybe structs, actually, since they would mirror the OpenSSL structs!) which implement a common interface.

malthe:
@jemc I would think that the object abstraction is zero-cost in the case where we can "optimize away the object". That's because the simple function API can be invoked in the constructor when a string is passed such that the digest value is readily available while at the same time be able to support chunked operation via `update`.

I don't know what Pony does today about that situation but eventually I could imagine that it would be efficient about such temporary objects (effectively doing away with them in an optimization pass).

sylvanc:
The pony compiler stack allocates objects where possible. That's determined by looking at the LLVM `DominatorTree`, which isn't perfect, but is pretty good. However, providing both a chunked and an "all at once" API would be nice, I think.

The chunked version might be able to take a hash algorithm as a type parameter.

malthe:
@sylvanc note that the "object version" is an all-at-once solution:

``` pony
let hash = SHA256(string).digest()
```

This should be as efficient as any non-object version with the stack allocation optimization.

jemc:
@malthe - It'll still be more FFI calls to shared library functions (which LLVM probably can't inline).

Theodus:
I think that @sylvanc's idea of providing both APIs would be the best approach.

malthe:
@jemc, if the EVP interface is used then the calls look like this:

``` c
EVP_MD_CTX_init(&mdctx);
EVP_DigestInit_ex(&mdctx, md, NULL);
EVP_DigestUpdate(&mdctx, mess1, strlen(mess1));
EVP_DigestUpdate(&mdctx, mess2, strlen(mess2));
EVP_DigestFinal_ex(&mdctx, md_value, &md_len);
EVP_MD_CTX_cleanup(&mdctx);
```

That is, it doesn't really matter if you provide an option to `update` the digest context with additional bytes. There is no "all-at-once" option in the library. I don't know if there is such an option at a lower-level but the EVP interface is the recommended layer.

jemc:
@malthe - @Theodus' example implementation [demonstrates the simple all-at-once interface exposed in libcrypto](https://github.com/Theodus/pony-crypto/blob/master/hashfunctions.pony#L102).

Theodus:
While trying to use the EVP interface I have found that the digest can only be updated before it is finalized, unlike Python's hashlib which allows the hash to be returned at any point. Is there any reason to use openssl's EVP rather than implementing it in Pony and generating the digests through the existing primitives?

Theodus:
The example at github.com/theodus/pony-crypto has been updated to include the new API.

text/0000-crypto.md
```diff
@@ -41,6 +41,17 @@ This package will use the OpenSSL crypto library through the C FFI. The followin
...
    
    see also: [Golang implementation](https://golang.org/src/crypto/subtle/constant_time.go?s=490:531#L2)
+   
+- Digest
+
+    class that produces a hash from chunks of input. The hash can be calculated at any time and updated with the update method.
+    
+    Example use:
+
+        let d = Digest.sha1()
```
jemc, line 51: Seems like this could take a primitive as a type argument to a `create` constructor, as I think @sylvanc was suggesting.  This would clear up my concern about having a big `match` statement in the implementation of `Digest`, since it could delegate to methods of the primitive.

This would look like:

``` pony
let d = Digest[SHA1]
// ...
```

Theodus:
If you look at the implementation, there is no match statement since a reference to the corresponding HashFn is stored as a tag in the state of Digest.

jemc:
> Is there any reason to use openssl's EVP rather than implementing it in Pony and generating the digests through the existing primitives?

I think the difference is that you'll have to allocate a new string to hold the composite, and copy all of the chunks into the composite before sending to the simple function.  I think this kind of defeats the point of having a chunked API, where the advantage is avoiding the performance cost of the extra allocate/copy operations.

> If you look at the implementation, there is no match statement since a reference to the corresponding HashFn is stored as a tag in the state of Digest.

This works if you're doing the allocate/copy described above, but not if you're using the true chunked API underneath.

Theodus:
Fair enough, I'll use the EVP.

jemc:
To be clear, I still think the one-call interface should be used in the one-call hash functions.

Theodus:
I agree.

Theodus:
The example now uses the EVP API.

Theodus:
I think that this should be merged if there are no further questions/concerns.

text/0000-crypto.md
```diff
@@ -0,0 +1,73 @@
...
+   
+- Digest
+
+    class that produces a hash from chunks of input.
+    
+    Example use:
+
+        let d = Digest.sha1()
+        d.update("message1")
+        d.update("message2")
```
jemc: I'm not sure about the name `update` for this method - in Pony, `update` methods usually take two arguments and use the `receiver(a) = b` sugar.

A better name for this might be `append`, since that's the conceptual operation going on here.

jemc:
@sylvanc - How do you feel about the digest constructors (as opposed to using a type parameter)?  I'm seeing in @Theodus' example implementation that the implementation is not too hairy without type parameters, and they may not actually end up bringing much benefit in this case, but wanted to check in with you first about it.

SeanTAllen:
is this ready for an up/down vote at the next sync call?

Theodus:
I would say yes.

sylvanc:
The `Digest` constructors look rather elegant to me. I think there might be a more efficient way to handle constructing the `_hash` array, but that's an implementation thing, not an RFC thing :)

PR #2, opened 2016-05-27
Add zero-copy String and Array trimming.

jemc:
This change adds zero-copy `trim` and `trimmed` methods to the `Array` and `String` classes in the `builtin` package.

EDIT: updated names to `trim` and `trim_in_place`.

### DIFF
--- a/text/0000-string-array-trim.md
+++ b/text/0000-string-array-trim.md
@@ -96,7 +96,7 @@ I propose the following API changes to the `builtin` package.
 
 * We should extend the Pony pattern about [limiting string allocations](https://github.com/ponylang/pony-patterns/blob/master/performance/limiting-string-allocations.md) to include a section about using `trim` and `trimmed`.
 
-We should lead by example, using `trim` and `trimmed` wherever applicable in the standard library and official Pony examples.
+* We should lead by example, using `trim` and `trimmed` wherever applicable in the standard library and official Pony examples.
 
 # Drawbacks
 
--- a/text/0000-string-array-trim.md
+++ b/text/0000-string-array-trim.md
@@ -104,7 +104,9 @@ I propose the following API changes to the `builtin` package.
 
 # Alternatives
 
-What other designs have been considered? What is the impact of not doing this?
+Other designs considered:
+
+* Use a wrapper object that is a `ByteSeq` and can apply as a virtual "slice" of any other `ByteSeq`, acting as an intermediary and erroring if the range is violated.
 
 # Unresolved questions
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I'm in favor of these changes and others of the sort. Array, String et al are used everywhere and we need them to be as high performance as possible. I'd like to redo Array and String entirely at some point. I think we need to take a really hard look at why each exists, the odd differences between them, their relationship to ByteSeq and Seqs in general. That however is a very large undertaking and we can make incremental improvements such as thus until such time as someone (perhaps  me) pulls together a RFC to kick off a full rethink of "Strings" in Pony.

jemc:
As a side-note, this is a design we half-sketched-out together in a sync call one or two weeks ago, but now I'm formalizing it here.

SeanTAllen:
My one concern is that `trim` & `trimmed` could be a little confusing. That said, I can't think of better names at this point in time

jemc:
To explain a bit of the naming thinking, I chose `trim` to distinguish from the existing `Array.slice`, and I chose `trimmed` because I like using the past-tense form of verbs for persistent-data-structure operations - as in, here is the array after having been "trimmed", but the original is still intact.  Hopefully that is as intuitive to others as it is to me.

SeanTAllen:
It would be nice if the names somehow conveyed the "borrowing" nature of `trimmed`.

Random thought. Probably an awful thought. But ruby has `!` methods that indicate the method changes the underlying object. Perhaps such a naming convention could be useful for Pony for knowing if a method creates a copy or not.

`trim` and `trim!` if it was ruby.

0joshuaolson1:
@SeanTAllen Julia has such a naming covention too. But Rust uses `!` for macros, which Pony may get one day...

SeanTAllen:
@jemc i'd really like to see this move forward. what can i do to help?

jemc:
Well, I was waiting to see if anyone had any other naming suggestions, but since no one has chimed in, let's move to final comment period.

SeanTAllen:
@sylvanc any comments?

Praetonus:
Maybe this could use the `in_place` naming already used in some places of the standard library: `trim` to create a new object and `trim_in_place` to modify the existing object.

SeanTAllen:
Are we ready for final vote on this @jemc ?

jemc:
@SeanTAllen - yes, right after I modify the RFC to incorporate @Praetonus' suggestion about `trim_in_place` which seems to be popular.

jemc:
Okay, updated based on @Praetonus' suggestion.

PR #3, opened 2016-05-30
Use named methods instead of "apply" and "update"

malthe:
This is a proposal to use named methods when possible, leaving the use of the "apply" syntax sugar to function-like objects and dropping the "update" sugar altogether. This directly affects a number of classes in the standard library which currently use "apply" and "update" for methods such as getting or setting an item.

[Rendered](https://github.com/malthe/pony-rfcs/blob/8f68d898cb9d81cb8e2203765de4ac57a23aa0d2/text/0000-use-named-methods.md)

### DIFF
--- a/text/0000-use-named-methods.md
+++ b/text/0000-use-named-methods.md
@@ -1,16 +1,13 @@
-- Feature Name: Use named methods when possible
+- Feature Name: Use named methods instead of apply/update
 - Start Date: 2016-05-30
 - RFC PR:
 - Pony Issue:
 
 # Summary
 
-This is a proposal to use named methods when possible, leaving the use
-of the "apply" syntax sugar to function-like objects and dropping the
-"update" sugar altogether. This directly affects a number of classes
-in the standard library which currently use "apply" and "update" for
-methods such as getting or setting an item.
+This is a proposal to use named methods throughout the standard library as opposed to using the "apply" and "update" methods that are generally chosen because they map to the callable and assignment sugar.
 
+Some objects have just a single "apply" method which supports an inline implementation as a lambda function (an object literal). This will still be possible, but it should no longer be a requirement that the single method be named "apply".
 
 # Motivation
 
@@ -23,41 +20,25 @@ the like may result in less precise object interfaces that map poorly
 to the underlying algorithms. The use of named methods throughout the
 codebase helps steer clear of this pitfall.
 
-In addition, the term "apply" in particular is typically not very
-meaningful in the context of an action such as getting an item from a
-collection. We can note that most objects are likely to have a default
-or primary action that is not well described with the term "apply".
-
+In addition, the terms "apply" and "update" are often not particularly meaningful. For example, a method to obtain an HTTP status code would be better named "code" (instead of "apply"), and the method to send a request from an HTTP client might be "request" (instead of "apply".)
 
 # Detailed design
 
 There are some 200 implementations in the standard library of an
-"apply" method and about a dozen "update" methods.
-
-If this RFC is admitted into the repository there needs to be a
-process of deciding new names. For example, unit tests currently have
-an "apply" method which could be renamed to "run", while the logger
-object could have its "apply" method renamed to "append".
-
+"apply" method and about a dozen "update" methods which would all need to be renamed. The process of coming up with suitable names is out of scope for the RFC process and should be discussed during implementation.
 
 # How We Teach This
 
-The documentation needs to be updated.
-
+This proposal represents a more simple library design and so there will be less to teach.
 
 # Drawbacks
 
-The "apply" and "update" sugar make some operations more compact.
-
+The "apply" and "update" sugar makes the code more compact.
 
 # Alternatives
 
-The bracket notation is a common syntax choice for item assignment and
-retrieval. It would be an option to introduce additional sugar to map
-special methods to this notation.
-
+No alternative is currently proposed.
 
 # Unresolved questions
 
-A complete list of suggested renames has not been developed at this
-time.
+It's currently unresolved whether the "update" sugar should still be supported (it's clear that "apply" has obvious uses such as having an object pass for a function).
--- a/text/0000-use-named-methods.md
+++ b/text/0000-use-named-methods.md
@@ -1,0 +1,63 @@
+- Feature Name: Use named methods when possible
+- Start Date: 2016-05-30
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This is a proposal to use named methods when possible, leaving the use
+of the "apply" syntax sugar to function-like objects and dropping the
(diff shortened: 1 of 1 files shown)


jemc:
So far, I'm not convinced that this would add more value than it removes.

I quite like the current use of both `apply` and `update` in the standard library and in my own code.  I see them as being quite similar to the operator method sugar (for example, `+` calls the `add` method), and as with those, I don't think they're terribly confusing, once you're familiar with the process.  If new learners are having trouble with this in practice, maybe we can make improvements to the documentation or the tutorial to make it clearer.

If there are classes where `apply` makes for a confusing name for a method when used explicitly, I wouldn't necessarily be opposed to adding "alias methods" with more descriptive names that just call `apply` (or vice-versa).

Regarding using square brackets instead for `update` and `apply` sugar, I don't think that's a viable option.  This would be ambiguous with type parameter syntax, especially once we have dependent types, where we can put compile-time values in type parameter brackets.

SeanTAllen:
I'm not convinced at this point in time and am leaning quite strongly against this. I would like to hear more concrete examples of:

```
The syntax sugar that some languages have for member assignment and
the like may result in less precise object interfaces that map poorly
to the underlying algorithms. The use of named methods throughout the
codebase helps steer clear of this pitfall.
```

In particular, as it relates to Pony's use of update/apply sugar that you are proposing replacing.

malthe:
The problem that I have with the "apply" and "update" sugar on collections is that it covers so little of the interface and that the sugarless interface is just as good (and arguably more powerful).

For example, the "update" interface actually returns a previous value if there was one and we often want a "setdefault" operation that takes a lambda to create a default value. The sugar suggests that we usually want to use the simple version of getting or setting an item, but really what you want is to make the most precise choice of operation – both to make the code readable and to get the best performance.

I think that `map.get(key)` is just as succint as `map(key)` and it just makes the language easier to learn and more uniform to use.

0joshuaolson1:
I agree with @jemc that this looks more like a naming concern for the standard library. Also, `map(key)` is analogous to array accesses, which don't use square brackets either.

SeanTAllen:
This seems to have stalled. Before this makes it to or out of final comment period, at the least I would like to see more conversation around what needs to change in all the pony examples, the documentation, what the level of breakage on existing code etc would be.

I'd also be interested in seeing how this impacts lambda/class interoperability.

Theodus:
Considering how much time has passed, I think that this PR should either be closed or moved to the final comment period.

sylvanc:
I don't see how `map.get(key)` can be considered as succinct as `map(key)`.

I also don't see how `map.set(key, value)` is as succinct as `map(key) = value`.

I think this is problematic for arrays, as indexing will no longer work (a feature very much expected in programming languages). Nor do I think think eliminating the ability to have indexing-like features on user defined classes improves the language.

As @SeanTAllen points out, this also would require lambdas to have named entry points. So instead of:

``` pony
let f = lamda(): U32 => 3 end
let x = f()
```

We would instead have:

``` pony
let f = lambda get(): U32 => 3 end
let x = f.get()
```

Similarly, partial application would have to change its syntax entirely, as it produces a lambda. That lambda would now have to have a name. I can't think of any useful way to express this syntactically.

Also, as @SeanTAllen points out, this would make lambda/class interoperability require significant planning.

Effectively, this RFC proposes to remove many core functional features from Pony. That's not something I'd like to do.

malthe:
I think `map(key)` misses the point that a map is an object that encapsulates a hash map and provides a number of useful operations. With `map(key)` it looks like a function, but it's actually an object.

And since the method name is _apply_, it reads as an operation that will possibly change the object. I think this works well for a callable object but not so much for item retrival on a map or an array.

Just to be clear, this proposal is not about removing the apply-sugar. It's about:
1. Removing the update-sugar;
2. Using named methods for objects in the standard library.

A lambda will still be an object that's callable because the method is attached via the name `apply`.

I suppose bracket syntax might be an option if classes are required to be capitalized and variables not.

SeanTAllen:
I don't think there is support for this change based on the comments herein, that said, I'd like to take this to final comment period and then to a vote. Are you ok with that @malthe?

malthe:
@SeanTAllen yeah I think there are enough comments in the thread to have a vote on it.

malthe:
I think the following examples help illustrate how named methods are more readable:
1. The documentation says that promises can be fullfilled or rejected, but in the implementation, fullfilling a promise is done through _calling_ it (via the "apply" method) instead of the obvious `fullfill()` – or `satisfy()` – which would be a direct counterpart to `reject()`.
   
   screen shot 2016-09-22 at 09 28 24
2. The HTTP client allows you to schedule a request by calling it. I think a more intuitive API would simply be `client.request(...)`.
   
   screen shot 2016-09-22 at 09 29 00
3. The status primitives in `net/http` – whether one agrees or disagrees with them – use `apply()` instead of the more obvious `code()`. And `string()` produces the status reason, which is also somewhat arbitrary. This method could simply be `reason()` instead. An HTTP status is exactly that, a code and a reason. It might be that `string()` should produce instead the two together, e.g. `"200 OK"`.

The gist here is that named methods allows us to express most clearly what a method does. The "apply" sugar is not a convenience in this context, it's a source of confusion.

SeanTAllen:
There's nothing in this RFC as it stands that assuages my concerns about the impact on ease of use for lambdas as they now exist. I think there would be a dramatic effect where I would now have to pass a lambda with not just the correct type signature but interface as well. That would likely have a massive impact. The impact on partial application is also not addressed at all.

I think there is a small win in one area with this RFC for large losses in others. As such, as it stands I'm not in favor of this RFC.

malthe:
@SeanTAllen, it's not mentioned in the RFC proposal because it's not related.

This RFC is not about removing the "apply" sugar, it's about not using it as a sort of "default" or primary method on an object.

I can pull out the part about removing the "update" sugar. That's all things considered orthogonal to the above concern.

SeanTAllen:
Its very much related because the existing standard allows you to use objects or object literals interchangeably for "primary" actions. This would remove that as a possibility.

malthe:
I think I understand what you mean.

In Java 8 you can satisfy an interface with a single method by passing a lambda:

``` java
Collections.sort(names, (a, b) -> b.compareTo(a));
```

The method here is actually called `compare` but you can still make an object literal for it because there is just a single method in the interface.

hibnico:
I confirm, Java is more flexible in that regard. This would make a nice new RFC for Pony.

SeanTAllen:
@malthe are you ready for a vote on this?

malthe:
@SeanTAllen, yes.

SeanTAllen:
We voted on this at the sync and have decided not to accept this RFC. Thank you for the time and effort that you put into this. Even if we aren't accepting this, we found the exercise of going through the RFC process valuable.

PR #4, opened 2016-05-30
Sugar for method chaining

malthe:
Lots of APIs support method chaining, but in Pony (and most languages) this needs explicit support in the class implementation. This is a proposal to introduce a syntax sugar for invoking a method with no return value and have the object returned.
- [Rendered](https://github.com/malthe/pony-rfcs/blob/50996c11ce338e50517fba164391fa804581da39/text/0000-sugar-for-method-chaining.md)

### DIFF
--- a/text/0000-sugar-for-method-chaining.md
+++ b/text/0000-sugar-for-method-chaining.md
@@ -1,0 +1,55 @@
+- Feature Name: Sugar for method chaining
+- Start Date: 2016-05-30
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Lots of APIs support method chaining, but in Pony (and most languages) this needs explicit support in the class implementation. This is a proposal to introduce a syntax sugar for invoking a method such that the return value is discarded and replaced with the object itself allowing further operation.
+
+
+# Motivation
+
+It can be useful to invoke multiple methods on the same model. An example is a configuration objct where the methods change or setup a part of the configuration.
+
+Rather than requiring API designers to wire this capability into the method return value in the form of returning ``this`` explicitly, this proposal is to use a special syntax to make it work in the general case.
+
+Methods such as ``Map.compact`` can be simplified by not having to declare and return ``this`` and some methods can return more useful return values without breaking the option to do method chaining.
+
+Note that this change also makes it clear that the returned object is in fact not a modified copy of the original object, but the same object. This helps promote the idea that in Pony, objects are immutable only by reference capability and not inherent design.
+
+
+# Detailed design
+
+The proposed syntax to invoke a method for chaining is ``object.>method``.
+
+When using the method chaining sugar, the existing return value is discarded and replaced with a reference to `this`.
+
+The following semantics apply:
+
+* If the receiver type can alias as itself, the chained type doesn't change;
+* If the receiver type is a non-ephemeral unique type (iso, trn or #send), the call was necessarily recovered so it is safe to chain the same type;
+* If the receiver type is an ephemeral unique type:
+  * If the call was recovered, if the method capability is the non-ephemeral type, or if the non-ephemeral type is a subtype of the method capability (i.e. tag method for iso object or box method for trn object), the chained type is the ephemeral type.
+  * Else, the chained type has the same capability as the method capability.
+
+# How We Teach This
+
+In the tutorial there will be a new section under "Methods" that motivates and explains this syntax.
+
+
+# Drawbacks
+
+This adds additional syntax to the language. Users will have to understand the meaning of the new notation.
+
+This is also a breaking change although existing code can easily be updated with the new syntax.
+
+
+# Alternatives
+
+Other syntax options were discussed such as ``object->method`` which has the problem that this syntax is already used for view adaptation and ``object:method`` which was considered to be visually hard to distinguish from a regular method call.
+
+
+# Unresolved questions
+
+None.


jemc:
I definitely like the idea of method chaining being done  with syntax on the caller side instead of cluttering up the API of the receiver with extra return values that the caller may or may not even want to use. As you say, the generated code is trivial for the caller, and it removes a lot of complexity from the receiver object's API.

Coupled with automatic receiver recovery, this would also have the potential to clear up a major headache when it comes to `iso` and `trn` objects.

Automatic receiver recovery lets you (in certain circumstances) call a `fun ref` method on an `iso` or `trn` receiver. The current convention for method chaining on `fun ref` methods is to have the method return `this` as a `ref`.  Together, this means you can't really chain `fun ref` methods with an `iso` or `trn` at all.  With method chaining as a special caller-side construct instead (as you describe in this RFC), it seems like you could make it work with `iso` and `trn` objects (again, under certain safe circumstances).  You'd have to carefully handle ephemerality during this chaining, but at first glance it seems like it could be made to work correctly.

The fact that this feature would let you chain methods that are not usually chainable on `iso` or `trn`, as well as methods that weren't written with chaining in mind (and don't have `this` as a return value), seems to make this feature rather compelling to me.

Regarding syntax, I'm not sure what would be best.  I'll try to give it some thought, but will also wait to see what others think about the syntax.

SeanTAllen:
Assuming this became a Pony feature, would we go through the existing standard library and remove instances of `this` being returned? 

Can this be applied to all methods with no return type? Behaviors are methods and they have no return type because they are async. Is the intent to have this apply to behaviors as well?  The semantic meaning between doing this for a behavior vs a function feels different to me.

Praetonus:
@SeanTAllen A behavior call returns the receiver, so it is already possible to do something like `myactor.mybe1().mybe2()`.

Regarding the RFC, I agree with @jemc, this could be great combined with automatic recovery.

SeanTAllen:
@Praetonus i didn't know that. interesting. i'd never tried that. i think that complicates this matter then as there is already a bit of implicit return here. that makes me more strongly favor either an implicit chaining (change default return of functions to this) or changing it on behavior. All of this is based on the assumption that we adopt this RFC in some form.

SeanTAllen:
This seems to have stalled. Is there anything that needs to be done to restart the conversation around this?

jemc:
Does anyone have any alternative syntax ideas other than the colon?

Has anyone done any analysis to know whether using a colon would be ambiguous in our LL(1) grammar?

Perelandric:
Is there a reason this is limited to methods that return nothing? Seems like if Pony is going to spend some of its limited resource of syntax options, it should be made as useful as possible.

We can already ignore a return value, so why not have this just be another way of saying _"I don't care what's returned"_? Then developers may be more likely to add a usable return value, knowing that the caller can ignore it if they want to chain a method off the original receiver instead. 

It also would seem easier to teach since there are no caveats to the context of its use.

As to the syntax, I'd prefer something like `..` over `:`. Google's Dart language uses that for its [cascade notation](https://www.dartlang.org/docs/dart-up-and-running/ch02.html#cascade), which also replaces the return value with the original receiver.

jemc:
@Perelandric - The way I read the description, "invoking a method with no return value and have the object returned", I just read it to mean that we would be ignoring the return value (but I can see why this language is confusing - it could probably be improved).  There is other language in the RFC that implies that it applies whenever you don't care about the return value.

Thanks for the link to the Dart `..` cascade notation - it's good to know that other languages have this concept, and see how they represent and teach it.  It also seems like `..` would be less likely to be an ambiguity problem than using a colon.

Perelandric:
Thanks @jemc. I'm (perhaps incorrectly) reading the detailed design as describing qualifications for its availability. If that's not the case, and it's instead describing where a developer may desire to use it but without any real restriction, then I think this is nice.

Agreed about the `:` too.

SeanTAllen:
I like the idea of a chaining operator quite a bit.

I don't like `..` because I think it will be visually hard to distinguish from `.`

Too easy to mistake what is going on.

jemc:
For syntax, @sylvanc recommended taking a look at the F# function composition operators.

I found this blog post, but haven't read it yet: https://fsharpforfunandprofit.com/posts/function-composition/

I'm going to read it and come back with some thoughts later.

sylvanc:
The docs for F# function operators:

https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/functions-%5bfsharp%5d#function-composition-and-pipelining

Specifically, I think `|>` is something Pony should have (possibly with that exact syntax?). Perhaps chaining could use some similar syntax? It would be nice to be able to do both pipelining and chaining, so that we could chain a call on the same receiver and the argument would be the result of the first call. I'm not sure what that would look like.

For pipelining, we need to think about:
1. If the result is a tuple, is it passed along as a tuple, or is it destructured into multiple arguments?
2. Is the pipelined result the last argument to the function on the right-hand side? That's how it works in functional languages. I don't think it should be a named argument, as that means a function has to be written in a specific way in order to be pipelined to.

A simple pipelining example:

``` pony
class Foo
  var x: U32
  new create(x': U32) => x = x'
  fun apply(): U32 => x
  fun m(y: U32): U32 => x + y

let a: Foo(1)
let b: Foo(2)
let x = a() |> b.m() // x = 3
```

If chaining were, for example, `||` then "chain and pipeline" could be `||>`. Perhaps that's too ugly.

``` pony
obj.method1(a, b)
  || method2(c)
  || method3(d, e, f)
```

Which is equivalent to:

``` pony
obj.method1(a, b)
obj.method2(c)
obj.method3(d, e, f)
```

And:

``` pony
obj.method1(a, b)
  ||> method2()
  ||> method3(d, e)
```

Which is equivalent to:

``` pony
obj.method3(d, e, obj.method2(obj.method1(a, b)))
```

On the other hand, `||` looks like `or` or even `parallel operation` (depending on what other language the reader is thinking of). Smalltalk uses `;` for method cascading (which obviously we don't want to use).

https://en.wikipedia.org/wiki/Method_cascading

http://news.dartlang.org/2012/02/method-cascades-in-dart-posted-by-gilad.html

hibnico:
Continuing the syntax brainstorming, what about keeping a dot around for chaining, reminding that it is about the call site. Like for instance:
- pipelining: `|>`
- chaining: `.>`
- pipelining and chaining: `|.>`

It has a kind a strange look, but has then the advantage to not look like something widespread in other languages. And it is coherent: _op+`>`_ is about special function calls, then pipe for pipes and dot for objects.

Maybe the strange feeling comes from the syntax resemblance to smileys, probably the last one would be a mad one-eyed elf! :)

Perelandric:
Just to throw a couple other ideas into the mix...

---
- `<>` for chaining
- `<|>` for chain and pipe

I think it conveys taking the receiver from the left (`<`) and using it on the right (`>`). Plus the non-pipe version seems very fast to type on many keyboard layouts.

---
- `+>` for chaining
- `+|>` for chain and pipe

The `+` conveys that you wish to _add_ another call off the previous receiver.

sylvanc:
Excellent syntax brainstorming! I worry about `<>` because it means the same as `!=` in many languages.

I was experimenting with good ol' antlrworks 1.5.2, and I think the following are unambiguous:

```
->
|>
-|>
```

The key is to use `->` in the same place as `.` and `~` (i.e. it must be on an `atom` and be followed by an `ID`) rather than as a binary operator. As a binary operator, it is ambiguous with the `->` in viewpoint adapted type.

This may be a Bad Idea, since we are reusing the `->` symbol here, but `->` does have previous usage as "another form of lookup" due to C/C++.

jemc:
@sylvanc - If we're using `->` for some form of chaining/piping _and_ for viewpoint adaptation, I think it would be a pretty confusing overload for users.

Maybe we should change the symbol for viewpoint adaptation?  What about `::`?

SeanTAllen:
Changing viewpoint adaptation is a rather large breaking change. Is it really worth it rather than coming up with something that is less potentially confusing for chaining et al?

Theodus:
I think that changing viewpoint adaptation should be avoided. How about the following:
`>>` for chaining
`|>` for pipelining
`|>>` for both

Would this cause any ambiguity?

sylvanc:
Currently, `>>` is "shift right", i.e. sugar for `shr`.

SeanTAllen:
Given the c roots of Pony, I think `>>` should stay with its "c meaning".

jemc:
@sylvanc, regarding your questions about pipeline operator behavior:

>  If the result is a tuple, is it passed along as a tuple, or is it destructured into multiple arguments?

Yes, I think destructuring would be cool, unless there is a reason not to do this.

> Is the pipelined result the last argument to the function on the right-hand side? That's how it works in functional languages.

Maybe this is how it works in F#, but that's not how it works in Elixir (it's the first argument).  Pretty unfortunate that we have conflicting precedents for the same operator :/.

Personally, pipelining into the first argument seems most useful and intuitive to me, given that the pipelined argument is coming "from the left".  However, I think either way we go, people from either F# or Elixir backgrounds are going to think it's "backwards".

SeanTAllen:
Clojure had two different methods for doing "first" or "last" when doing "pipelining" which they call "threading".

Perelandric:
I'm a little confused by the pipelined example in https://github.com/ponylang/rfcs/pull/4#issuecomment-232617000.

>
```pony
obj.method1(a, b)
  ||> method2()
  ||> method3(d, e)
```
>
>Which is equivalent to:
```pony
obj.method3(d, e, obj.method2(obj.method1(a, b)))
```

If the result of `method1(a, b)` is being used as an argument in the next call `method2()`, then what exactly is being chained? 

The equivalent shows `obj` being the receiver for `method2()`, but is that because `method1()` returned the same object which is also being used as the argument for `method2()`? Or could `method1()` be returning a `String` and `method2()` is actually "looking back" to what the previous receiver was in the `method1()` call to find out what its receiver should be?

If it's looking back, then is the behavior the same for the non-pipelined example? If so, then that would seem to be a bit different from the behavior described in the RFC, since the RFC seems to describe the LHS operand being "moved ahead" in order to be a substitute for the RHS method's return value.

SeanTAllen:
@sylvanc @jemc thoughts?

SeanTAllen:
I feel like this is a really important RFC but its stalled. I'm going to get this on a sync call agenda so we can hash it out. Hopefully this weeks as I am unable to attend next week's.

jemc:
So far I think @nlalevee's [syntax suggestion](https://github.com/ponylang/rfcs/pull/4#issuecomment-232831865) seems to have the fewest problems, from my perspective.

I also find it to be relatively intuitive in terms of symbols meaning things that they look like (`|` means there is pipelining involved, `.` associated with method calls, etc).

I would vote for an RFC that used these symbols.

---

However, I still find the [first-argument / last-argument dispute](https://github.com/ponylang/rfcs/pull/4#issuecomment-238042760) over pipeline operator meaning to be pretty unsettling - or rather, it's a big reason why I wouldn't consider this settled.  To me, first-argument meaning is a significant preference but that may be due to my Elixir-bias (and complete ignorance of F#).

But I think this point is far from being a minor aesthetic difference - it makes a major difference for how useful the operator is, and in what context.  It also makes a difference in how APIs will be designed that are intended for use with this operator (will they plan for the first argument or the last argument to be supplied "via pipe"?).

As I see it, the first-argument pipeline operator helps you create functions that use the first argument as a pseudo receiver, which is totally useful for applying primitive functions over objects instead of having to get the author of the class to define that method on that object (see some discussion around using this in https://github.com/ponylang/rfcs/pull/30#issuecomment-246183356).

For example, `foo |> Bar(1, 2, 3)` is a conceptual emulation of `foo.bar(1, 2, 3)`, without needing to add the `bar` method to `foo` itself (as long as `bar` can be implemented in terms of the existing public API for `foo`).

The last-argument meaning of the pipeline operator is more like currying, and I don't find it to be very compelling when we have partial application already (with the `~` operator).

malthe:
I think there is an aesthetics or perhaps readability concern that the first step shouldn't be different than the next ones.

This syntax (variations are possible) addresses that:

``` java
using x
  .foo()
  .bar()
end
```

It's also acceptable to use inline-form: `using x .foo() .bar() end`.

Note that `using` would be a keyword – or it could be another term.

I think the pipe operator et al are out of scope for this RFC.

SeanTAllen:
We discussed this on the sync call and came to the conclusion that too many things have become conflated in this RFC and we should break it apart.

We have here...

the original chaining idea
and a couple different forms of pipelining.

We ask for folks who have been involved in the pipelining conversation to open a new RFC for pipelining.

@sylvanc is going to be looking at our grammar in order to find something nice and parseable and report back with ideas on what we can do specifically for chaining. We can add various pipeline operators that have been discussed here in response to other RFCs.

We all agree that we would like basically everything that has come up in this RFC to be in Pony but that having them all together in this one RFC is making it hard to move forward and that each can be added independently.

jemc:
Sorry I wasn't able to make the sync call this week.

I definitely agree that the other operators are outside the original scope of this RFC, and are distracting from the original intent of it.  However, at the same time, I think we're most likely to come out with a coherent syntax if we try to plan for these "similar" kinds of operators together, so that we can be sure to leave some "room" in the BNF for the other ones we want to add.

SeanTAllen:
@sylvanc you were going to look into the grammar to see what could make sense. have you had a chance to do that?

Praetonus:
Is there anything we can do to make this move forward?

There are performance concerns with the `return this` pattern as it adds jitter in the optimiser alias analysis and it can hinder some really important optimisations.

Maybe we can discuss it at the next sync? I think I'll also start an implementation with the various symbols proposed here to see if there is any actual problem with some of them.

SeanTAllen:
We were waiting on @sylvanc. 

@Praetonus can you take over doing:

> @sylvanc is going to be looking at our grammar in order to find something nice and parseable and report back with ideas on what we can do specifically for chaining. We can add various pipeline operators that have been discussed here in response to other RFCs.

SeanTAllen:
@Praetonus I'm very in favor of moving this forward. Let's discuss at the next sync.

jemc:
As I mentioned in an earlier comment, I like @nlalevee's [suggestion](https://github.com/ponylang/rfcs/pull/4#issuecomment-232831865) best of the ones I've heard so far.

I find `.>` to be quite intuitive (`.` to make the method call, `>` to send the receiver forward).

I don't see why this wouldn't work fine with our grammar.

Praetonus:
Basic implementation done. I've pushed it at [my fork](https://github.com/Praetonus/ponyc/tree/chaining). It wasn't extensively tested so there might be bugs left.

The complicated part was the refcap stuff. I've implemented the following rules, I'm not sure if they have to be more or less strict.
- If the receiver type can alias as itself, the chained type doesn't change.
- If the receiver type is a non-ephemeral unique type (`iso`, `trn` or `#send`), the call was necessarily recovered so it is safe to chain the same type.
- If the receiver type is an ephemeral unique type
  - If the call was recovered, or if the non-ephemeral type is a subtype of the receiver type (i.e. `tag` method for `iso` object or `box` method for `trn` object), the chained type is the non-ephemeral type.
  - Else, the chained type has the same capability as the method receiver capability.

jemc:
@Praetonus - those semantics sound correct to me at first blush, though it would be good to go over it at the sync call.

---

I think we should also mention in the RFC that the grammar must allow for whitespace around the operator, to allow uses like:

``` pony
my_array
  .> push(4)
  .> push(5)
  .> append([6, 7, 8])
```

malthe:
FWIW I still think the originally proposed (by me) colon syntax is preferable over `.>`.
1. It looks similar enough that it's easy to grok what it could mean.
2. It looks different enough that you can tell them apart visually.
3. It can be made required that there be no whitespace between the symbol ":" and the method name. This is – presumably – different in the case of `.>`. With the space after `.>` it becomes less clear that `push` is a method on `my_array`.

``` pony
my_array
  :push(4)
  :push(5)
  :append([6, 7, 8])
```

That said, I don't think `.>` is a terrible choice.

Praetonus:
The grammar is agnostic to whitespace. I'm not sure if it would be a good thing to add constraints on that.

jemc:
General consensus on the sync call was that `.>` is a bit more clear for a reader than using `:`.

@malthe - if you were to update the RFC to include @Praetonus' notes about semantics, and to use the `.>` syntax, it should be quickly accepted.

malthe:
I think it makes sense to require that the method has a return value of `None` and to perhaps consider changing the semantics of behaviors such that they don't return themselves.

It could lead to subtle bugs to automatically discard the return value.

Thoughts?

Praetonus:
> require that the method has a return value of `None`

I don't see any compelling reason to do that. What kind of bugs do you have in mind?

> changing the semantics of behaviors such that they don't return themselves

I'd second that, but I think it's outside of the scope of this RFC.

jemc:
@malthe - as I see it part of the appeal of this RFC is that methods that currently return `this` can be free to change to more interesting return values that can be helpful to the user in different situations.  Restricting to methods that return `None` would prevent those kinds of changes, and also really limit the usefulness of this feature.

SeanTAllen:
Very very much with @jemc on that. Restricting to methods returning None takes away a ton of the interesting parts of this to me. It could turn out to be a bad idea, but its one I'm really interested in trying out.

malthe:
I have updated the proposal.

Note that I have suggested that `object~>method` is valid syntax and means method chaining with partial application.

Praetonus:
What would chaining on partial application mean? Partial application returns a new object without doing anything to the receiver so chaining would discard that object and essentially be a no-op.

jemc:
@malthe - what is the motivating use case for `object~>method`?

That is, a partial application can never have side effects, and the only purpose of it is to produce a callable object, which is the result of the expression.  Since the method chaining form will discard the original result of the expression and put the receiver in its place, it seems like there would be no way to retrieve the callable object that was produced by the partial application.

malthe:
@jemc I am not sure there is a use-case other than consistency.

The partial callable object returned is still a new object, but if/when you eventually complete the call, the result will be the original object (when applicable). It should be possible since there is still a reference to it in the partial object.

``` pony
let foo = object~>method(1)
foo(2).method(3, 4)
```

That is, `foo(2)` returns `object` back.

But I think it might be pretty useless. I'll remove it so that we can get on with it.

jemc:
@malthe ah, I understand how it would work now, thanks for explaining.  But like you I also can't think of a compelling use case at the moment :wink:.

Are you ready to move to final comment period?

malthe:
@jemc yep let's do it.

Praetonus:
I think the refcap rules for ephemeral types can be even more permissive. Since the return value is always discarded, there can be no alias outside of the isolation boundary if the call was recovered. It looks like it would be safe to chain the ephemeral type in this case. It would allow things like

``` pony
var x: A iso
x = (consume x).>foo().>bar()
```

Basically, the "borrow" many people have been talking about. We could even add some sugar to allow it to work with `let`.

Praetonus:
So, to sum it up
- If the receiver type can alias as itself, the chained type doesn't change.
- If the receiver type is a non-ephemeral unique type (iso, trn or #send), the call was necessarily recovered so it is safe to chain the same type.
- If the receiver type is an ephemeral unique type
  - If the call was recovered, if the method capability is the non-ephemeral type, or if the non-ephemeral type is a subtype of the method capability (i.e. tag method for iso object or box method for trn object), the chained type is the ephemeral type.
  - Else, the chained type has the same capability as the method capability.

malthe:
You're saying "If the receiver type is an ephemeral unique type [...] if the receiver type is the non-ephemeral type".

Did you mean to format this differently?

Praetonus:
"Receiver type" is meant to designate the type outside of the function in the outer bullet point, and the type inside of the function in the inner bullet point. I'll reformulate.

Praetonus:
Updated.

jemc:
This looks to be ready for a vote tomorrow, so I'm switching the label.

Please speak up if it's not actually ready for some reason.

SeanTAllen:
@Theodus i'm really busy right now, can you create the issue for this?

Theodus:
@SeanTAllen sure!

SeanTAllen:
@Theodus thanks

PR #5, opened 2016-05-31
Optimisation inhibition

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/opt/text/0000-optimisation-inhibition.md)

### DIFF
--- a/text/0000-optimisation-inhibition.md
+++ b/text/0000-optimisation-inhibition.md
@@ -1,0 +1,103 @@
+- Feature Name: optimisation-inhibition
+- Start Date: 2016-05-27
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add various Pony `compile_intrinsic` functions preventing some compiler optimisations such as dead code removal.
+
+# Motivation
+
+In some very specific domains such as benchmarking, some compiler optimisations are really bad. In particular, we do not want the benchmarked code to be eliminated even if it has no side-effects from the optimiser standpoint.
+
+# Detailed design
+
+Since we're talking about benchmarking and performance analysis, this new construct must
+
+- allow really fine-grained control: keep as little code as possible.
+- not introduce any additional cost: no function call, no register/memory assignment, nothing.
+
+In C, this is possible by using inline assembly marked `volatile`. For example:
+
+```c
+void escape(void* p)
+{
+    asm volatile("" : : "g"(p) : "memory");
+}
+
+void clobber()
+{
+    asm volatile("" : : : "memory");
+}
+
+int main(void)
+{
+    int a = 0;
+    escape(&a);
+    a = 42;
+    clobber();
+}
+```
+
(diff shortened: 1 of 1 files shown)


jemc:
Are there examples of other languages that provide features like this?  Rust, maybe?  If so, what do they call them?  I'd be more likely to be in favor of this RFC if there is some precedent for it that users might be familiar with and we could use similar language to describe the features.

Praetonus:
Rust has direct support of inline assembly so it doesn't need compiler support for this, as in C. That said, it looks like the standard benchmark library of Rust has a function named `black_box` wrapping the same assembly. Google's benchmark library (C++) uses `DoNotOptimize`. These two functions have the same goal as `escape` in the RFC: keeping a computation producing an arbitrary result untouched by the optimiser.

jemc:
I must admit I still don't have a good idea of what `clobber` is doing.

Praetonus:
`clobber` is basically a "touch all memory" function. This is useful with `escape`d pointers. `escape` is saying "I'm reading this input and doing something in memory" and `clobber` is saying "I'm doing something in memory". From the compiler standpoint, it could mean that `escape` stored the pointer in some global state and then `clobber` read the pointed value from the global state (but nothing happened in reality).

To sum it up, any write to an `escape`d memory location before a call to `clobber` must stay in the program.

jemc:
I'm not sure about the name `Optimiser`, given that what this primitive actually does is _prevent_ optimization - the name definitely leads you to believe otherwise.

Praetonus:
What do you think about `DoNotOptimise` for the primitive? It's not a noun, I don't know if that's problematic. For the functions, I'm thinking `apply` (for the sugar) and `observe` (because it's observing potential side effects) for `escape` and `clobber` respectively. I'll update the RFC if there are no objections against these names.

jemc:
That name is definitely clearer, and I can't think of anything better, so I don't think I have any objections to it.  But we can hear what others think.

Praetonus:
Since there hasn't been any new elements, I'd like to push this RFC to the final comment period.

SeanTAllen:
@Praetonus can you send a short email to pony+dev with a link to this RFC, a short description and a request for any final comments? Thanks!

Praetonus:
Sorry, I had no time to do it during the last days. I see you've sent the mail, thanks.

SeanTAllen:
@Praetonus not a problem.

sylvanc:
I like `DoNotOptimise` as it results in code like:

``` pony
actor Main
  new create(env: Env) =>
    var x = U32(0)
    DoNotOptimise[U32](x)
    x = 42
    DoNotOptimise.observe()
```

That reads nicely to me.

SeanTAllen:
@Praetonus are you ready for this to move to an up/down vote at the next sync call?

Praetonus:
@SeanTAllen Yes.

PR #6, opened 2016-06-01
Standard library network implementation

malthe:
The standard library should include support for common network protocols and follow an established set of implementation guidelines. There needs to be a principle of clean layer separation and the codebase organization should reflect this.

Read [complete proposal](https://github.com/malthe/rfcs/blob/standard-library-network-implementation/text/000-standard-library-network-implementation.md).

### DIFF
--- a/text/000-standard-library-network-implementation.md
+++ b/text/000-standard-library-network-implementation.md
@@ -1,0 +1,50 @@
+- Feature Name: Standard library network implementation
+- Start Date: 2016-06-01
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The standard library should include support for common network protocols and follow an established set of implementation guidelines. There needs to be a principle of clean layer separation and the codebase organization should reflect this.
+
+
+# Motivation
+
+The current codebase is somewhat disorganized and lacks a clean layer separation.
+
+For example, the HTTP server codebase is closely coupled with both the TCP and SSL transport layer. The HTTP communication channel needs to separated from the HTTP message layer. And further, in the interest of supporting protocols such as WebSocket and HTTP/2, abstractions need to be devised such that a clean layer separation is possible.
+
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


Theodus:
I am very much in favor of this change. It will make the standard library net package more maintainable and extensible.

jemc:
Honestly, until looking at https://github.com/ponylang/ponyc/issues/899, I wasn't really sure what exactly was being proposed, so this RFC might have been lacking a little in details.

As I understand it, the only _user-facing_ impact of this RFC would be the separation of the `net` package into smaller subpackages - is that correct?  If so, I think I'm on board with this.

However, as I mentioned in https://github.com/ponylang/ponyc/issues/899, I have some disputes about how files are named, but I also have plans to reform the file names in all of the builtin and standard library packages, so I think we can have that discussion separately.  Also, renaming file names inside packages won't have any user-facing impact, so we don't have to worry about two "big naming changes" for users to adapt to if we leave off my file naming proposal for a separate discussion.

malthe:
@jemc – the RFC proposal goes a bit further to say that some of the subpackages should be reworked to have a more clean layer separation such that we can better support newer protocol versions and replace lower layers. In particular, this concerns the HTTP server implementation.

SeanTAllen:
The http server should be replaced.

jemc:
@malthe - I don't know how to judge those other aspects without seeing a proposed design.  I definitely agree in principle with the concept of decoupling where it makes sense.

SeanTAllen:
@malthe are you available to do a zoom call to discuss this RFC?

malthe:
Closing this proposal (for now at least).

The reason why "ssl" and "http" are in subdirs is that both have a hard dependency on an external SSL library.

The file organization is somewhat orthogonal to the implementation design and the coupling level of each protocol and this currently affects mostly the HTTP server which anyway needs to be rewritten.

PR #7, opened 2016-06-04
Ponybench

Theodus:
This proposes a microbenchmarking tool for the standard library.

### DIFF
--- a/text/0000-ponybench.md
+++ b/text/0000-ponybench.md
@@ -13,48 +13,44 @@ This package adds to the tooling around pony. Good tools support the language ec
 
 # Detailed design
 
-The API of this package is similar to that of ponytest for ease of use. The following is an example of a comlete program with 2 trivial benchmarks:
+The PonyBench.apply behavior has the following signature:
+```pony
+be apply[A: Any #share](name: String, f: {(): A ?} val, ops: U64 = 0)
+```
+The amount of ops may be set manually, but by default it is handled automatically with the benchmark run for a minimum of 1 second by default. If the second has not elapsed when the Benchmark function returns, the amount of ops is increased in the sequence 1, 2, 5, 10, 20, 50, … and the function is run again. A simple average (total execution time over ops) is used to calculate the execution time in nanoseconds.
 
 ```pony
-
-actor Main is BenchmarkList
+actor Main
   new create(env: Env) =>
-    PonyBench(env, this)
+    let bench = PonyBench(env)
+    bench[USize]("fib 5", lambda(): USize => Fib(5) end)
+    bench[USize]("fib 10", lambda(): USize => Fib(10) end)
+    bench[USize]("fib 20", lambda(): USize => Fib(20) end)
+    bench[USize]("fib 40", lambda(): USize => Fib(40) end)
+    bench[String]("fail", lambda(): String ? => error end)
+    bench[USize]("add", lambda(): USize => 1 + 2 end, 1_000_000)
+    bench[USize]("sub", lambda(): USize => 2 - 1 end, 1_000_000)
 
-  fun tag benchmarks(bench: PonyBench) =>
-    bench(_BenchmarkAdd)
-    bench(_BenchmarkSub)
-
-class iso _BenchmarkAdd is Benchmark
-  fun name(): String => "addition"
-
-  fun apply(b: BenchmarkRunner) =>
-    let n1: USize = 2
-    let n2: USize = 2
-    for i in Range[USize](0, b.n()) do
-      b.discard(n1 + n2)
-    end
-
-class iso _BenchmarkSub is Benchmark
-  fun name(): String => "subtraction"
-
-  fun apply(b: BenchmarkRunner) =>
-    let n1: USize = 4
-    let n2: USize = 2
-    for i in Range[USize](0, b.n()) do
-      b.discard(n1 - n2)
+primitive Fib
+  fun apply(n: USize): USize =>
+    if n < 2 then
+      n
+    else
+      apply(n-1) + apply(n-2)
     end
 
 ```
-
-Each benchmark is run for a minimum of 1 second by default. If the second has not elapsed when the Benchmark function returns, the value of b.n() is increased in the sequence 1, 2, 5, 10, 20, 50, … and the function is run again. A simple average (total time to run the benchmark function over b.n()) is used to calculate the execution time in nanoseconds.
-
 Output:
 ```
-addition      10000000	        28 ns/op
-subtraction   10000000	        28 ns/op
+fib 5     50000000          33 ns/op
+fib 10     5000000         352 ns/op
+fib 20       30000       42951 ns/op
+fib 40           2   646773866 ns/op
+**** FAILED Benchmark: fail
+add        1000000           2 ns/op
+sub        1000000           2 ns/op
 ```
-This shows that each operation was executed 10,000,000 times and took an average of 28 ns/op (note that the accuracy is poor at low execution times until compiler optimization can be properly restricted).
+For example, this shows that the add function was executed 1,000,000 times and took an average of 2 ns/op).
 
 The current proof of concept implentation can be found at [https://github.com/Theodus/ponybench](https://github.com/Theodus/ponybench)
 
--- a/text/0000-ponybench.md
+++ b/text/0000-ponybench.md
@@ -1,0 +1,67 @@
+- Feature Name: ponybench
+- Start Date: 2016-06-04
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+Ponybench provides a microbenchmarking facility that can be used to examine the performance of pony code. This gives users an opportunity to easily profile their programs, remove bottlenecks, and make them faster.
+
+# Motivation
+
+This package adds to the tooling around pony. Good tools support the language ecosystem and increase adoption.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


Praetonus:
I don't like `Runner.discard`. It is adding bias to the benchmarking and this is really bad. A function preventing compiler optimisations should be zero-cost. Since this is the object of [this RFC](https://github.com/ponylang/rfcs/pull/5), I suggest you to remove `discard` altogether and to mention `DoNotOptimise` in the benchmark package docs.

Theodus:
I agree completely, Runner.discard is just a placeholder for that feature. I am unsure how the optimization inhibition would be used in practice though.

SeanTAllen:
@Theodus how are you feeling about this? do you want/need more feedback before final comment period?

Theodus:
I'm waiting for #5 to go through so that the compiler optimization doesn't get in the way of the benchmark results.

SeanTAllen:
@Theodus ok, let me know when you want to go to final comment with this.

@jemc @sylvanc any comments before then?

jemc:
I think I'd like to see some more description of the intended operation of tool - what does the output look like when it's running?  How does it choose/configure how many iterations to run?  What stats does it report on the result (mean, std dev, percentile, etc)?

SeanTAllen:
@Theodus averages can hide a lot of variation. do you imagine this would eventually store to something like HDR histogram if someone finishes porting it to Pony?

Theodus:
@SeanTAllen various features may be added after the original implementation. Adding extra analysis for the results would be beneficial, especially if memory usage could also be benchmarked.

SeanTAllen:
So that wouldn't be extra analysis. That impact on how data is collected.

Theodus:
I think that this is ready for the final comment period

SeanTAllen:
@Theodus updated Theo. Can you send an email to pony+dev announcing it?

SeanTAllen:
@Theodus as you are very active with RFCs, I gave you collaborator access to this repo so you can update labels for final comment etc in the future.

SeanTAllen:
@Theodus can I add discussing this to the sync call agenda for next week? I'm not really familiar with a microbench tool of this sort and would be interested in discussing the topic of benchmarking tools for Pony.

Theodus:
@SeanTAllen That sounds good to me.

Theodus:
@SeanTAllen Should we talk about this in the next sync meeting, or should it be moved to `ready for vote`?

jemc:
Sorry for being late to the party with this comment, but I noticed that this proposed design doesn't seem to have a mechanism for benchmarking asynchronous operations.

I wouldn't necessarily use that as a reason to delay acceptance of this RFC - it can always be added in a future RFC.  For example, a future RFC could add `bench_async`, where the given lambda is expected to return a `Promise[X]` instead of returning `X` directly.

SeanTAllen:
@Theodus thoughts on the async issue raised by @jemc?

Theodus:
I think that the ability to benchmark asynchronous operations is important and should be part of the initial RFC. It was an oversight on my part.

text/0000-ponybench.md
```diff
@@ -0,0 +1,78 @@
...
+fib 40           2   646773866 ns/op
+**** FAILED Benchmark: fail
+add        1000000           2 ns/op
+sub        1000000           2 ns/op
+```
+For example, this shows that the add function was executed 1,000,000 times and took an average of 2 ns/op).
+
+There will also be an async behavior as follows:
+```pony
+be async[A: Any #share](name: String, f: {(): Promise[A] ?} val, timeout: 1_000_000_000, ops: U64 = 0)
```
jemc: Will the `timeout` param be the timeout for just one operation or for all of them?
Will the operations be run in series or in parallel?

Also, note that this is invalid syntax - `timeout` is missing its type.

I'm also not totally sold on the method name `async` when `bench` is the name of the synchronous version of the method - seems unintuitive to me at least.

Theodus:
@jemc The timeout param is only for the benchmark given since only one benchmark will run at a time. Also the synchronous version is `apply`, but I still agree that `async` is a bit off. I'm open to suggestions for a different name.

Theodus:
This is ready for a vote unless there are any more questions/concerns.

jemc:
Why did you remove the timeout entirely?  My questions were oriented toward making sure the parameter was well-documented, but I don't necessarily think it should be removed.

Theodus:
The timeout conflicts with the auto benchmarking, which uses its own timeout for determining how many iterations to call the given function.

jemc:
How so?  It seems that these serve two different purposes - the user-specified _timeout_ for distinguishing a failed run from a valid one, and a _timing_ for determining the expected duration of future runs.  I don't understand how these would conflict with eachother, so perhaps you can explain further?

Theodus:
What action should the timeout affect? If each iteration of the function, then setting up the timer would be counted in the average run time. If the total time taken for all iterations, then the timeout should take the auto benchmarking algorithm into consideration.

jemc:
The point of a timeout in async tests in `ponytest` is to halt a test that would otherwise never end (never finish its work).  So if I'm testing something broken that would never finish its asynchronous action (in this case, never fulfill or reject the `Promise`), I don't want to have to sit in front of a blank screen that never completes its work.

In other words, the timeout is a way of saying "this operation should never take more than _N_ seconds under any circumstances, so if it did, something is wrong and you can expect that it would never stop - therefore, halt it now and tell me about the failure".

In this context, applying the timeout to the average run time doesn't make any sense, because if your benchmarked operations were going to fail the timeout, you would expect them _never to finish running_.  So if you check the timeout against the average run time, you already know that the operations finished and you're not serving the real purpose of the timeout.

What I would expect is that the timeout acts as a halting mechanism for the tested operation, and you cancel the timer when the operation finishes.

However, it's also possible that this could affect the benchmark results for the operation (CPU time would be spent in the `Timers` actor), so it might be best to only use the timeout during the "exploratory" phase of the benchmark (the pre-iterations when you are trying to determine the number of iterations to perform).

Theodus:
Ok, so I think that the timeout should always run first to test if the promise is fulfilled before the timer fires. Then normal benchmarking operations will follow. Does that seem reasonable?

jemc:
@Theodus - yes, but don't you already have an "exploratory" phase to determine the number of operations already?

Theodus:
@jemc - not if a non-zero value for ops is passed to the function (since zero triggers the auto benchmarking).

jemc:
Right, makes sense.  I suppose it's also possible that the user doesn't care about guarding against timeouts, and might want to give a zero timeout value, signifying indefinite.  It definitely seems like _some_ users will want to give a timeout as a sanity check, though.

text/0000-ponybench.md
```diff
@@ -54,9 +54,9 @@ For example, this shows that the add function was executed 1,000,000 times and t
 
 There will also be an async behavior as follows:
 ```pony
-be async[A: Any #share](name: String, f: {(): Promise[A] ?} val, ops: U64 = 0)
+be async[A: Any #share](name: String, f: {(): Promise[A] ?} val, timeout: U64 = 0, ops: U64 = 0)
 ```
-which will calculate the average time until the Promise is fulfilled.
+which will calculate the average time until the Promise is fulfilled. If timeout is greater than 0, the benchmark will fail if the promise is not fulfilled within the time given.
```
jemc: I think it would be worth specifying the discussion we had that the timeout is only enforced the first time, before the benchmarking iterations begin.

sylvanc:
@jemc and I are voting yes!

jemc:
We do still have reservations about the name `async` for the asynchronous benchmark method, but still had no better suggestions, and it can always be changed again by a future RFC.

rurban:
I dislike the name and think **BenchMark** is better. 

ponybench should rather be an external benchmarking tool to benchmark many pony compilers against a standard set of benchmarks. To see regressions and improvements in the compiler/run-time itself. ponybench tests pony, Benchmark tests user code.

lisael:
Incidentally, I've found this RFC while creating my own benchmark framework RFC. I've already made a PoC, FWIW  https://github.com/lisael/pony-bench . It's much smaller than @Theodus ponybench but has interesting approaches that may be (or not) discussed here.
1. ponytest-like interface : all benchmarks are registered, pre-processed and then run with a helper
   1. It's consistent with ponytest, if you know how to unit-test, you almost know how to benchmark
2. best of 3 runs. That's how Python's `timeit` do. It mitigates the case: a kernel thread and a cron and 3 daemons reclaim CPU while I run the benchmark
3. The calibrating phase (called "exploratory" in this comments) is... different (Though, I'm not sure it's any better, I'll play with ponybench a bit to get arguments :) )

I've got more general question. How will we use the microbenchmarks, in ponyc stdlib?

I mean, shall we add benchmarks in all packages? If so, how do we run them, and what is a pull-request blocking result? Any refactoring should include benchmarks to prove it doesn't slows down the code? Do we answer these questions in a future RFC?

Theodus:
The original plan was to use an interface resembling that of ponytest. I changed to the current interface because the loops were being optimized out by the LLVM backend of the Pony compiler. Also, I find passing a lambda function as an argument to be easier than creating a separate class for each benchmark. I do like the simplicity of the implementation by @lisael, but the algorithm used here is based on that of [Go's benchmarking tool](http://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go). The final question about the use of benchmarks in the standard library is an important one and should be discussed in [issue 1137](https://github.com/ponylang/ponyc/issues/1137). @lisael would you mind moving the discussion over there?

PR #8, opened 2016-06-06
Try, catch, else, finally

malthe:
This RFC advocates for a rename of "else" to "catch", "then" to "finally", and introduces an "else" clause which is reached if no error was raised (like in Python).

See [full proposal](https://github.com/malthe/rfcs/blob/try-catch-else-finally/text/0000-try-catch-else-finally.md).

### DIFF
--- a/text/0000-try-catch-else-finally.md
+++ b/text/0000-try-catch-else-finally.md
@@ -1,0 +1,55 @@
+- Feature Name: Try, catch, else, finally
+- Start Date: 2016-06-06
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This RFC advocates for a rename of "else" to "catch", "then" to
+"finally", and introduces an "else" clause which is reached if no
+error was raised.
+
+# Motivation
+
+The "catch" and "finally" terms are common in programming
+languages.
+
+Using "catch" frees "else" up for a different purpose which is the
+meaning from the Python language where the "else" clause exists to
+guard precisely the code that we can expect to sometimes raise an
+error and not the code that comes after.
+
+The term "then" is commonly used in promise terminology to mean "on
+success". This is why this RFC recommends "finally" instead.
+
+
+# Detailed design
+
+The design of the new "else" clause is straight-forward. After the
+"try" clause finishes we set a flag and outside the "try" block this
+flag is tested to see if we should enter the "else" clause.
+
+
+# How We Teach This
+
+The documentation will need to be updated.
+
+
+# Drawbacks
+
+The main drawback is that we already have lots of code that uses the
+current syntax.
+
+That said, we can improve the code in many places using the new "else"
+clause. This will result in better code with less bugs.
+
+
+# Alternatives
+
+Currently none are proposed.
+
+
+# Unresolved questions
+
+The reuse of "else" leaves a question of how to transition code safely
+to the new syntax.


Praetonus:
##### New `else` clause

This looks similar to the construct proposed [here](https://github.com/ponylang/ponyc/issues/291).

I don't really like using `else` here. To me, the meaning of `try/else` is "try to do this, or else do that if this fails". An `else` block executed on success doesn't sound right to me.

Regarding the proposed implementation, setting and checking flags would be bad for performance. It would be better to simply merge the `try` and `else` blocks at code generation and let the LLVM exception facilities do the work for us.

##### Renaming
- `else` to `catch`: This sounds strange since we're not really "catching" things when raising Pony exceptions. It's more like "we couldn't produce a coherent result so we're doing something else instead".
- `then` to `finally`: I don't have a strong opinion for either one here.

jemc:
As I understand it (correct me if I misunderstand your proposal), you're advocating changing `else` to mean "code that happens if there was no error".  I think this is counter-intuitive, as my brain doesn't want to correlate that meaning with the keyword `else`.  I wasn't aware that Python uses the `else` keyword this way, but nevertheless I find it counter-intuitive. Not only that, but it's also basically the opposite meaning that it has now, which is really bad for compatibility with everyone's existing Pony code.  You did note that this is backwards incompatible, but I think it's hard to overestimate how much code that this would disrupt.

Note that choosing to avoid the `then` keyword doesn't mean it can be freed up for using in method calls like `Promise.then()`, since it will still be a keyword in `if` and `unless` blocks. I'd instead think we should try to reuse the `if` block keywords as much as it makes sense to do (more on this below) - so that we avoid creating more keywords as much as possible, as more keywords means more words that users can't use in their own method and field names, exactly the problem you're trying to address with `then`.

However, keyword choices aside, I think I agree with the _spirit_ of this proposal.  In fact I think this is quite similar to [something I proposed a while back](https://github.com/ponylang/ponyc/issues/291), long before the RFC process was here.  That is, I think it was close to the same idea (a way to guard only the code that can raise an error, but not the code that comes after), but I had some different preferences for the keywords than you put forward here.

Specifically, I was advocating for:

``` pony
try
  // code that might raise an error (calls to partial functions or explicit error)
then
  // code that depends on the try code, and can access local variables introduced there
else
  // code that executes when something in the try block raises an error
finally
  // code that always executes no matter what
end
```

The advantage of these keyword choices means that all of the keywords that are shared with `if` and `unless` blocks will have now roughly the same meaning in `try` blocks.  That is `then` and `else` can be used in an analogous way to create a `if`/`unless`-style conditional where the branching depends on whether the predicate raises an error, instead of the boolean value of the predicate.  Making `try` blocks act more like `if` and `unless` blocks would be good for the language's consistency.  Compare the two examples below to note the similarities:

``` pony
if _boolean_fn() then
  _had_success()
else
  _had_failure()
end
```

``` pony
try _partial_fn() then
  _had_success()
else
  _had_failure()
end
```

My proposed keyword choices also have the benefit of being fully compatible with all existing Pony code that uses `try`/`else`/`end`, since it would not _require_ using a `then` to distinguish the predicate from the statically error-free block.

The only existing code that would break is code that uses the _current_ meaning of `then` in a `try` block, which would have to be rewritten to use `finally`.  I think this is not a huge deal, because I'm not aware of any code anywhere that is actually using the `then` keyword in a `try` block (although someone might be using it somewhere). Whatever growing pains we might have, we could mitigate this by first deprecating the use of `then` in a `try` block (and giving a helpful compiler message to steer folks in the right direction).

jemc:
> This looks similar to the construct proposed [here](https://github.com/ponylang/ponyc/issues/291).

Heh, yep!  You beat me to it, @Praetonus!

> `else` to `catch`: This sounds strange since we're not really "catching" things when raising Pony exceptions.

Yeah, this is an excellent point.  Compared to other languages, in Pony, there's nothing to "catch" because exceptions do not have an associated object / type / value.

malthe:
@jemc I agree – that's a way superior design and choice of keywords.

I had to reopen a [new pull request](https://github.com/ponylang/rfcs/pull/9).

PR #9, opened 2016-06-07
Try, then, else, finally

malthe:
This RFC advocates for a rename of "then" to "finally", and introduces a "then" clause which can be placed immediately after the "try" block to program the case where no error was raised. Local variables introduced in the "try" block can be used here.


text/0000-try-then-else-finally.md
```diff
@@ -0,0 +1,80 @@
...
+that "then" is a reserved keyword already and cannot be used a method
+name).
+
+Instead, "finally" is used in mainstream languages such as Java and
+Python and communicates its semantic quite clearly.
+
+
+# Detailed design
+
+The "then" clause is syntax sugar that could expand to:
```
jemc: As @Praetonus [noted in the other PR](https://github.com/ponylang/rfcs/pull/8#issuecomment-224120374), the most efficient and simplest implementation is to simply have the compiler generate the LLVM IR as if the `then` block was tacked on to the end of the `try` block, but to have the Pony compiler enforces that the `then` block cannot raise an error.

text/0000-try-then-else-finally.md
```diff
@@ -0,0 +1,80 @@
...
+construct and vice versa.
+
+
+# Alternatives
+
+The unsugared syntax exhibited in the detailed design notes can be
+thought of as an alternative. It has the drawback that a "try" block
+needs to be named (in the sense that a "let" symbol is used) and
+naming is hard. This leads to common names such as "success" except
+this obviously can't be reused in the same scope.
```
jemc: This also has the drawback that the compiler's `consume` tracking of branches cannot accurately know that the `else` block and the `then` block are mutually exclusive.  So, for example, you could not `consume` a reference in both branches as you otherwise could if you were dealing with `then` and `else` blocks on an `if` statement.

malthe:
I updated the text to reflect the preference on an LLVM IR-based implementation.

jemc:
As I said in #8 (and in https://github.com/ponylang/ponyc/issues/291), I like this change and these keywords.

I think it will go a long way toward making me feel more confident about my code.  As it stands now, I'm constantly worrying about whether I'm accidentally including more error-able operations in a `try` block than what I intended - so why not let the compiler worry about that for me instead, if I can express my intent that the `then` block should be error-free?

One of the arguments against this in https://github.com/ponylang/ponyc/issues/291 is that it was "optional type checking", so it was likely that "most programmers will just not use it".  However, I think the fact that @malthe came to a similar conclusion that this feature was needed says a lot about it actually being needed.

However, it's worth asking around to others - Have you felt this need before?  Would you use this?

malthe:
I'm not sure I agree that it's "optional type checking" because it shouldn't be optional in the sense that we should try to reduce the `try` block to the minimum required.

It is simply bad form to put safe code in a `try` block unless it's completely trivial such as the subtraction in `unsafe_method(size - 1)`. Why? Because it is an unnecessary cognitive load on the developer who has to read the code and understand what's going on.

While it is true that you can always refactor your code to use a separate function, this often doesn't happen – perhaps because [naming is hard](http://martinfowler.com/bliki/TwoHardThings.html).

In Python the curiously named `else` block – the equivalent of `then` in this RFC – is used all the time.

jemc:
> While it is true that you can always refactor your code to use a separate function, this often doesn't happen – perhaps because [naming is hard](http://martinfowler.com/bliki/TwoHardThings.html).

It also introduces a different type of cognitive load - searching back and forth between method definitions to try to piece together an understanding of what's going on.

---

And yes, I also agree with the rest of your points.

SeanTAllen:
Can I get some sample code of the problem this solves? I'm not understanding the problem this solves and I think that is purely from how I currently write Pony. I'm sure there is a valid reason for this, I just don't understand at the moment.

malthe:
The proposed `then` block is actually pretty awkward in terms of scope – it's not intuitive that a symbol defined in `try` can be used in `then`.

In addition, it was mentioned on the zoom call that `finally` – the current `then` – is really orthogonal to exception-handling (it is contingent on leaving a code block).

It seems like a good idea to try and develop a syntax akin to Go's `defer` – but explore the possibilities in not just using the entire method as the scope. The quintessential example is opening a file and eventually close it regardless of the code path taken. Another is obtaining a lock and releasing it subsequently.

I think there might be a different design that could alleviate the need for a `then` block. To reiterate, the problem is to identify where an error occurs whether it be when you read or run the code. The strategy proposed in this RFC was to reduce the amount of code in a `try` block while maintaining a sensible code flow, but we can probably do better.

Praetonus:
> It seems like a good idea to try and develop a syntax akin to Go's defer – but explore the possibilities in not just using the entire method as the scope. The quintessential example is opening a file and eventually close it regardless of the code path taken. Another is obtaining a lock and releasing it subsequently.

For things like this, we already have `with` blocks.

SeanTAllen:
I think it makes sense to close this RFC with some comments as to why after a new RFC that people can be pointed to is PR'd. Thoughts @jemc, @malthe, @sylvanc ?

SeanTAllen:
@Praetonus we do have with blocks but this would be more inclusion. A block of code that after a scope is guaranteed to run. The semantics are different than `with`. An RFC for that will hopefully clear up the differences more and I think an important part of that RFC would be how it impacts on the existing `then` with `try` and also on `with`.

@malthe is summarizing a conversation re: this PR that occurred during yesterday's sync call. @malthe, @jemc, @sylvanc and I all agreed that this RFC wasn't going to accomplish its goal. But expanding it out to something like Go's defer could be very interesting and is something we should explore in a future RFC.

jemc:
The approach we discussed on the sync call was to proceed with discussing this strategy for the new meaning of `try`/`then`, removing the current meaning of `then` in a `try` block in favor of something else.  Since these two features are orthogonal, I think it makes sense to discuss them in separate RFCs.

That is, I think this RFC should be kept open, but amended to say that the current behaviour of `then` will be _removed_ instead of renaming as `finally`.

Then someone can open a separate RFC to discuss what a block-that-always-runs-at-the-end-of-a-block-even-in-case-of-early-return might look like.

malthe:
@jemc I decided to close this RFC and open a new since the present one here represents a path to extended the syntax and what we probably want is to actually remove some of it.

jemc:
The new RFC does not address the need for the new meaning of `then`.  I guess I'll have to open yet another RFC for that change?

malthe:
It was my understanding that we didn't really see the need for this
proposed "then" block — a departure from mixing error handling with flow
control. So I left it out from the new RFC.

jemc:
[This was my understanding of what was discussed.](https://github.com/ponylang/rfcs/pull/9#issuecomment-224923969)

PR #10, opened 2016-06-08
Optional 'return'

hibnico:


SeanTAllen:
Can you give an example 

`Not having to put the 'return' keyword on purely functional function like traditional C-like language does bring clarity to the code. But for longer piece of code, for functions which unfortunately cannot be displayed entirely on the screen, its a little harder to see quickly where the function is actually ending and returning the value. It is especially true since Pony doesn't have some braces to mark the end of a function.`

hibnico:
See the _scanToken function there:
https://github.com/nlalevee/pony-yaml/blob/master/src/yaml3/scanner.pony#L116

SeanTAllen:
I don't see how a return at the end (which changes return from meaning "early return") helps.

Unrelated but are you familiar with docstrings in Pony? You have a lot of multiline comments like:

```
  /*
   * Eat whitespaces.
   *
   * Tabs are allowed:
   *
   *  - in the flow context;
   *  - in the block context, but not at the beginning of the line or
   *  after '-', '?', or ':' (complex value).
   */
```

that would make excellent docs strings.

hibnico:
Many times I was scrolling from the top of the function to search for the end, and I was ending way below. When reading carefully the code, like when reviewing code, indeed, it would not be missed. But when browsing I often get lost, since it is far from the function declaration and it is just an expression with no apparent side effect.

About comment style: I know. Most of the code was copy-pasted from the C version of the libyaml library. Since transforming C comments into Pony doc strings is painful, I'll do it when this proto will start to work.

malthe:
I think an alternative to this would be an explicit `end` after a method definition.

This is problematic for one-line methods, but perhaps they're overrated anyway.

jemc:
> I think an alternative to this would be an explicit `end` after a method definition.

If you're talking about a required one, I'm definitely opposed.  If it's an optional one, I'm probably opposed to that too - it seems way weirder than allowing an optional `return`.

> one-line methods, but perhaps they're overrated anyway.

Nah, one-line methods are great, especially for constant-value methods mimicking things like C enums.

Perelandric:
In my short time using Pony, I've had one situation where I felt having an explicit `return` at the end, after a rather long `match`, would've enhanced clarity. Was a tiny bit disappointed to find that this was forbidden.

Even if the value of the `match` itself was to be returned, I could see it being useful to have an explicit `return` statement before a long `match` expression as an immediate reminder that this is the final result of the function body.

Not crucial, but would be nice here and there. Though I wonder how often it would be abused.

jemc:
For whatever it's worth, I personally am not against allowing an explicit `return` on the final expression of a method, as I don't think it causes any problems for the implementation, and it doesn't cause any problems for folks who don't want to use it - as far as I'm concerned developers could just use their best judgment about when it makes things more clear.

In general, I'm opposed to dictating style for other people's projects from the top-down, and I see this as simply a style concern which doesn't make a big difference one way or another, so I don't see a problem with allowing it.  But I also wouldn't really consider myself a strong advocate for this change.

SeanTAllen:
Personally I would like to know how this specifically would be taught (tutorial? if yes, where in the tutorial?)

Also, no drawbacks, I don't believe. Every new addition of code comes with some carrying cost. I'm not sure what the cost is for this one.

Perelandric:
Since this is just removing the restriction on using `return` with the value of the final expression of the function, I'm not sure what more would need to be taught, except perhaps to describe idiomatic use of `return` in the [methods](http://tutorial.ponylang.org/expressions/methods.html) section of the tutorial, which doesn't explicitly describe this restriction at the moment.

hibnico:
[RFC](https://github.com/nlalevee/rfcs/blob/master/text/0000-optional-return.md) updated about where the documentation should be modified as suggested by @Perelandric.

There are mixed feelings triggered by this RFC, so I suggest to enter the final comment period to resolve its status one way or the other.

SeanTAllen:
@nlalevee please send an email to pony+dev list announcing that this has entered the final comment period.

cquinn:
I'm generally against this change, as it introduces multiple styles. I think it is valuable to have a single, simple idiomatic style for a language. This makes it easier to read each other's code, share code, contribute to projects, etc. Golang does a good job of this, with basic syntax and formatting being dictating with gofmt, and a pretty well described idiomatic style described here: https://golang.org/doc/effective_go.html

I think Pony would benefit from an approach like this, and introducing multiple alternate ways to do things is counter to that.

jemc:
For whatever it's worth, I'm pretty against the Go approach of machine-formatting code style.  When you strip away the programmer's ability to impose a context-specific style on a piece of code, you get codebases that look like a desert, barren and devoid of all expression.

Some programmers (I'm one of them) like to play with and adjust formatting for a particular snippet so that the important bits of information encoded are maximally apparent, and the unimportant or repetitive bits are correspondingly de-emphasized.  This makes for codebases that are much more readable, and the ideas behind the code jump out at the reader in more obvious ways.  This in turn makes it easier for new readers to quickly become familiar with a codebase and the ideas behind it.

As far as I can see the argument against such practices is that having style be subjective and context-dependent creates the possibility for style disputes.  While this is true, to those of us who value the freedom to look for more eloquent ways to express an idea will gladly accept style disputes over codebases that are devoid of any expressive style.  I think it comes down to seeing programming as a fundamentally creative/artful/expressive activity, rather than a purely mechanical one.  In my opinion, programming is an art and should be done by code artists, and should not be delegated to a "Mechanical Turk".

I don't think that this particular opinion of mine is very relevant to this particular RFC, but I just wanted to mention it since @cquinn brought up the Go approach as something for Pony to strive toward, while I disagree in general with that goal.

Perelandric:
> _"Golang does a good job of this, with basic syntax and formatting being dictating with gofmt..."_

Except that it isn't dictated. The Go team understood that it was important to make it an optional tool instead of enforcing it with compiler errors.

There will always be _some_ style differences. In Go, some people prefer this:

``` go
var foo, bar int = 4, 2

if condition {
    return foo
}
return bar
```

While others prefer this:

``` go
var foo int = 4
var bar int = 2

if condition {
    return foo
} else {
    return bar
}
```

`gofmt` makes no change to these different styles. The `return` syntax in the former example was needed early on because the compiler wasn't able to determine if all branches returned a value. This [has been fixed](https://golang.org/doc/go1.1#return), and now either way is fine.

I see this RFC as no longer preventing a rather innocuous syntax that can add clarity to the code in some situations. I doubt it'll have much impact on people's style choices.

That said, while I still find it to be a minor annoyance when the compiler barks at me because I've changed something and forgot to delete the `return`, which is now at the end of the function, it is just that... a _minor_ annoyance that is seldom encountered.

cquinn:
True, gofmt only adjusts whitespaces and doesn't modify code syntax. But the Effective Go doc does recommend your first example over your second one.

I do find that the Go community's embracing of a generally common idiomatic form to be a uniting factor. Of course there are variations within it that each programmer will prefer, but an overall consistency across the standard and popular libraries (I think) does improve the ecosystem and contributes to the language's success. I'd like to see Pony grow a similar community, and so would like to borrow some of the good ideas from Go.

This change is not a big deal, but could begin a slippery slope of adding alternate syntaxes for doing the same thing that might just make Pony harder to learn. Just my 2¢.

SeanTAllen:
@nlalevee are you ready for this to go up for a vote?

hibnico:
Yep @SeanTAllen , let's go for the vote so we get a decision about this proposal.

cquinn:
👎

sylvanc:
I find @cquinn 's description of the problems compelling (the part about different ways to do the same thing). 👎

SeanTAllen:
We voted and decided not to adopt this RFC.

We think any convenience is relatively small, it introduces more than 1 way to do to things which isn't a reason to reject everything but it is an issue unless you can show a large win. All in all the carrying cost of this new code to support is something we don't feel is merited by the possible benefits from this change.

PR #11, opened 2016-06-09
Guidelines for source file names in the standard library.

jemc:
[Rendered](https://github.com/jemc/pony-rfcs/blob/file-naming-guidelines/text/0000-file_naming_guidelines.md).

### DIFF
--- a/text/0000-file_naming_guidelines.md
+++ b/text/0000-file_naming_guidelines.md
@@ -1,0 +1,43 @@
+- Feature Name: file_naming_guidelines
+- Start Date: 2016-06-08
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add guidelines for naming Pony source files in the standard library.
+
+# Motivation
+
+We have style guidelines for Pony source code in the standard library, but we don't yet have any guidelines for naming Pony source files. Having a consistent, reproducible scheme for naming files is a natural extension of code style, and key to good organization.
+
+# Detailed design
+
+I propose the following rules for naming Pony source code files in the standard library:
+
+- The *file name* of Pony source files should be based on the name of the *principal type* defined in that file.
+    + The *principal type* in a file is the type that makes up the bulk of the significant lines of code in the file or is conceptually more important or fundamental than all other types in the file. For example, if a file defines a trait type and a group of small class types that all provide that trait, then the trait type should be considered the *principal type*.
+    + If there are multiple types defined in the file which all have equal significance and a shared name prefix, then the shared prefix should be used as the *principal type name*. For example, a file that defines `PacketFoo`, `PacketBar`, and `PacketBaz` types should use `Packet` as the *principal type name*, even if no `Packet` type is defined.
+    + If there are multiple significant types defined in the file which do not have a shared name prefix, then this should be taken as a hint that these types should probably be defined in separate files instead of together in one file.
+- The *file name* should be directly derived from the *principal type name* using a consistent reproducible scheme of case conversion.
+    + The *file name* should be the "snake case" version of the *principal type name*. That is, each word in the *principal type name* (as defined by transitions from lowercase to uppercase letters) should be separated with the underscore character (`_`) and lowercased to generate the *file name*. For example, a file that defines the `ContentsLog` type should be named `contents_log.pony`.
+    + If the *principal type* is a private type (its name beginning with an underscore character), then the *file name* should also be prefixed with an underscore character to highlight the fact that it defines a private type. For example, a file that defines the `_ClientConnection` type should be named `_client_connection.pony`.
+    + If the *principal type* name contains an acronym (a sequence of uppercase letters with no lowercase letters between them), then the entire acronym should be considered as a single word when converting to snake case. Note that if there is another word following the acronym, its first letter will also be uppercase, but should not be considered part of the sequence of uppercase letters that form the acronym. For example, a file that defines the `SSLContext` type should be named `ssl_context.pony`.
+
+# How We Teach This
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I prefer `-` to `_` in file names but, I'm good with this. +1 to this RFC.

hibnico:
I really like the idea of mapping the types declared in a file and the file name. I often struggle with that reading the standard library code.

Then, is there a rationale for the file name to be snake cased rather than camel cased ?

Maybe it is my long years of Java coding which makes me biased, but I find really no brainer to search of a type and just look for a name.ext. I needs a mental transformation. And it can be painful if the tools doesn't support that transformation. For instance, if I am reading some code, there is some type I don't know about, I would really like to just copy the name and make my favorite editor search for that name + "*.pony". Having to transform the name to a snake cased one would be painful.

Also, not sure yet if it is a good pattern for the standard library, but I have found that it works great where files which contain main types are camel cased, and for the files which contain no particular principal type, I just use a lower case name which tries to describe what's in there. See for instance that package:
https://github.com/nlalevee/pony-mysandox/tree/master/src/async_compute

malthe:
I prefer "" to "_", e.g. `contentslog.pony`.

It's true that this makes a module that map to a private principal type unrecognizable but it can be argued that top-level private classes shouldn't be principal (e.g. `_ClientConnection` from the proposal).

The fact that the underscore syntax, e.g. `contents_log.pony` follows the method naming convention is a bad thing in my view. And the underscore syntax is not sufficient anyway for a machine to convert the name to a camel-case class name. For example, `ssl_connection` can't be trivially turned into `SSLConnection`.

jemc:
> is there a rationale for the file name to be snake cased rather than camel cased?

I think the only rationale would be that files with uppercase letters tend to be uncommon in most source trees, but this is a pretty weak one.  I think I could be easily convinced to follow the convention of using the exact name of the principal type for the file name, case included (for example, `ContentsLog.pony` and `_ClientConnection.pony`.  I would probably +1 that idea, but I would expect others to be a bit more resistant - maybe I'm wrong, though.  It would be interesting to hear others' thoughts on that.

> it can be argued that top-level private classes shouldn't be principal

If there are no public types in a file, and we can't consider private types to be principal types, then how would we name the file?  If we name the file after something else (I have no idea what), then how do we find the file that defines a given private type when looking for it?  Some private types (like `_ClientConnection`) are extensive and important, and in my opinion there's no reason to suggest they can't be principal types.

malthe:
@jemc, Java has inner classes which arguably have pretty good utility as _strategies_ or _behaviours_ used internally in a method. But I think principal private classes are mostly an anti-pattern. We need to be able to understand what objects are in the system and putting an underscore in front of it is like a big warning sign saying "no access".

hibnico:
@malthe if we should compare to Java, since in Pony the file has no visibility meaning then private class in Pony are the Java class with default package visibility. So even in Java a principal type of a file may not be exposed in an API, not useable for third party code.
Also, the discussion here is just about a file name convention, not rules for an API design. I "just" see it as an improvement of browsing code.

SeanTAllen:
@jemc i think this is ready for final comment, agree?

jemc:
Sure, let's do that.

For the record, I think the main remaining point of disagreement is about the word separation formatting in file names.  We've got advocates for the following options:
1. `contentslog.pony`
2. `contents-log.pony`
3. `contents_log.pony` (my preference, and the one currently described in this RFC)
4. `ContentsLog.pony`

I think we should try to come to some consensus here within the comment period (or, if necessary, at the end when we vote).

For the record, I actually really like option 4 due to its simplicity and consistency with the type name (I think I even like it better than option 3), but I see it as being controversial for others (maybe I'm wrong?) due to the relative uncommonness of using uppercase letters in file names.  Option 1 is my least favorite, because there is no indication of word separation at all, but it is the closest to what we have now, so that's worth noting.

SeanTAllen:
I prefer 2 but find 3 to be completely reasonable and don't feel strongly enough to argue for 2.

I dislike 4 because of annoying issues on OSX with case sensitivity or lack thereof. Whenever I start using capital letter in source code file names on OSX eventually some sensitivity edge case crops up. Thus a preference for 1 to 3 over 4.

In general my top 2 are

3 and 2

followed by

4

followed by 

1

Theodus:
I personally prefer 2, but 3 is a close second.

SeanTAllen:
ready for final vote on this @jemc ?

jemc:
@SeanTAllen - close, but one more question:

@Theodus, @SeanTAllen, since you both said you prefer 2 - I was about to change the RFC to reflect this, but the question hit me - what about the case of the private principal type?  The RFC advocates for distinguishing these with a preceding underscore to match the one on the type, but it's not clear how this would look to me in a "dasherized" scheme.  That is, I'm having trouble liking either `_client-connection.pony` or `-client-connection.pony`, and I think that's a good reason to prefer underscores over dashes (so that we avoid a mixture or a conflation of underscores and dashes).  Any other ideas?

SeanTAllen:
i find dashes to be of a more civilized age.

but i'm good with _ as many languages do that.

Theodus:
The underscoreat the beginning would be my preference for sure.

jemc:
@Theodus - so `_client-connection.pony`?  I don't really know if I can bring myself to like that.

Theodus:
@jemc, I got two other opinions and they both agreed with you, so I suppose the dash is more appealing.

jemc:
But I'm still saying `_client_connection.pony` - so I don't see how you're agreeing with me! :ghost:

Theodus:
Like I said, that option was a close second.

PR #12, opened 2016-06-10
Function to remove unneeded space in arrays and strings

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/array-string-shrink/text/0000-array-string-compact.md)


jemc:
Seems good to me. :+1:

SeanTAllen:
As presented, this would make me believe the memory will be returned to the OS, which wouldn't be the case. I think as part of teaching this, we need to discuss in at least a cursory fashion how the pool allocator works and what this might actually accomplish.

SeanTAllen:
I think this RFC needs to include how this interacts with the pool allocator and if there are any changes that would need to happen with it in order to make this useful.

Praetonus:
@SeanTAllen I don't think it is really interacting with the pool allocator since it is only making the memory available for garbage collection by reallocating the necessary space. It could be integrated directly in the allocator by splitting chunks but I'm not sure it would be a good thing (I'll add that in the alternatives). But you're right, the current language isn't specific enough. I think saying that the old storage is made available for garbage collection would be sufficient.

I'll also remove the bit about not knowing object sizes in `Array` since it is possible to compute them with `Pointer._offset`.

SeanTAllen:
Sounds good @Praetonus

SeanTAllen:
This is looking good to me.

Praetonus:
I'd like to push this RFC to the final comment period.

SeanTAllen:
@Praetonus 

Can you send a short email to pony+dev mailing list with:
- a link to this RFC
- a short description of the RFC
- request for any final input/comments?

Thanks,
Sean

sylvanc:
Looks good. Should `shrink_unused` possibly be called `compact` ?

Praetonus:
`compact` seems good to me. I'll update the RFC.

SeanTAllen:
I feel like this is ready for a vote up or down at the next Sync and am labeling as such.

Praetonus:
I've modified the bit about the computation of object sizes in the implementation of the function. I know the final comment period has ended but this is an important detail and the previous proposal for this can lead to bugs.

text/0000-array-string-compact.md
```diff
@@ -35,7 +35,14 @@ fun ref compact(): String ref^
 
 The function will reallocate sufficient storage for the existing elements in the collection and will copy them to this new storage. The old storage can then be garbage collected.
 
-As said in the docstrings the function may not do anything, specifically for small arrays. This is because small allocations (realised by `pony_alloc_small`, i.e. <= 512 bytes) are always rounded to a power of two. We can compute stored object sizes with `Pointer._offset` so we'll always know when compacting is impossible.
+As said in the docstrings the function may not do anything, specifically for small arrays. This is because small allocations (realised by `pony_alloc_small`, i.e. <= 512 bytes) are always rounded to a power of two. Because of this, we want to know when compacting is impossible to avoid unnecessary reallocations. Since `Array` is generic, `Array`s of different types could store objects of different sizes, which requires us to compute stored object sizes. This is possible with `Pointer._offset` but it is unclear and can lead to bugs caused by overflow in pointer arithmetic. Therefore, we'll add the following function to `builtin.Pointer`:
+
+```pony
+fun tag _elt_size(): USize
```
jemc: Is there a compelling reason for using this shorthand for the word "element"?  You're only saving four letters here, so I'm not sure it's worth the extra confusion when we could just use the full word and be more clear and readable.  We definitely have longer method names than this elsewhere in the API.

Praetonus: No compelling reason. I'll change that.

PR #13, opened 2016-06-10
Drop "then" from "try/else" clause

malthe:
Drop "then" clause from "try/else" clause – see [proposal text](https://github.com/malthe/rfcs/blob/try-else-and-drop-then/text/0000-try-else-and-drop-then.md).

See [PR #9](https://github.com/ponylang/rfcs/pull/9) for an extended conversation on the topic.

### DIFF
--- a/text/0000-try-else-and-drop-then.md
+++ b/text/0000-try-else-and-drop-then.md
@@ -26,6 +26,8 @@ and how the "with" statement conceivably solves the problem. Since in
 some languages exception handling is used for flow control it might
 also be good to talk about this.
 
+The "Try-then blocks" section in the tutorial can be removed.
+
 
 # Drawbacks
 
--- a/text/0000-try-else-and-drop-then.md
+++ b/text/0000-try-else-and-drop-then.md
@@ -19,6 +19,14 @@ addition the "with" statement supports multiple such resources at the
 same time without additional nesting.
 
 
+# How We Teach This
+
+In the documentation we need to talk about what resource management is
+and how the "with" statement conceivably solves the problem. Since in
+some languages exception handling is used for flow control it might
+also be good to talk about this.
+
+
 # Drawbacks
 
 The main drawback is that code that uses "try/else/then" today will
--- a/text/0000-try-else-and-drop-then.md
+++ b/text/0000-try-else-and-drop-then.md
@@ -1,0 +1,31 @@
+- Feature Name: Try, else, no "then"
+- Start Date: 2016-06-10
+- RFC PR:
+- Pony Issue: #291
+
+# Summary
+
+Drop "then" clause from "try/else" clause.
+
+
+# Motivation
+
+The behavior is orthogonal to error handling. It applies in general
+when leaving a code block.
+
+The "with" statement already provides the necessary functionality for
+obtaining and giving up a resource such as a lock or file handle. In
+addition the "with" statement supports multiple such resources at the
+same time without additional nesting.
+
+
+# Drawbacks
+
+The main drawback is that code that uses "try/else/then" today will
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@malthe you removed sections from the RFC, for example "how do we teach this".

As part of the removal of the feature I think in "how do we teach this", we need to have a strategy for communicating the change and what if any existing documentation will need to be changed.

malthe:
@SeanTAllen, added.

SeanTAllen:
@malthe which documentation? tutorial? or something else? are there materials/examples etc that need to change because of this?

I'm in favor of this but I think it behooves us to know what needs to be updated so that when a PR arrives we can make sure we don't have dangling documentation etc.

malthe:
@SeanTAllen I have added a note that the relevant [Try-then blocks](http://tutorial.ponylang.org/expressions/exceptions.html#try-then-blocks) documentation section can be removed.

I couldn't tell how much the construct is used in the repository. I think the easiest is to simply remove the parser code that implements the syntax and have the compiler list the errors. This is a to-do that should be done before we move this RFC into the final stage I think.

SeanTAllen:
@malthe are you ready to move this to final comment period?

Perelandric:
If the `then` would run in the case where the code in the `else` errored or returned, it would feel a little like a more flexible version of Go `defer` in that it isn't bound to the entire function body. Then again, if that behavior is desired, a different construct that doesn't _require_ a potential error may be better.

jemc:
> a different construct that doesn't require a potential error may be better.

@Perelandric - that's more or less what we discussed on the last call where we talked about this - the concept is somewhat orthogonal and could use its own form that isn't tied to `try`.

malthe:
@SeanTAllen, I think it's ready but I wanted to get an idea of the extent to which the "then" clause is used in the current codebase (standard library and/or examples) and tried removing the parse rule from ``parser.c``. But even I try to also remove ``then_clause`` from the various calls to ``ast_get_children`` I always get an assertion error: ``Assertion 'p != NULL' failed``.

SeanTAllen:
@malthe ok. let me know when you feel this is ready to go into final comment period by dropping an @ here.

SeanTAllen:
This seems to have staled. Is there any reason @malthe not to take this to final comment period?

malthe:
@SeanTAllen – yes and no: the difficulty is that actually implementing this change is not so straight-forward since `with` is implemented as sugar that uses `then` to dispose its objects.

But the good news is that `then` is not currently used anywhere in the standard library, so it should be relatively smooth sailing once removed.

The reason this RFC has stalled is that ideally we would first have a general `defer` statement that we could use to implement `with` and then be free to remove `then` from `try`.

Or alternatively somehow disable `then` from direct usage so that it's just used as an implementation detail. I couldn't figure out how to do that in a better way than renaming it to e.g. `then_disabled` or `__then`.

SeanTAllen:
@malthe i think with input from @Perelandric, @jemc and maybe @sylvanc we should be able to get this to a place where we have an agreed upon solution that could be voted on even if it isn't immediately implementable.

sgebbie:
In addition to dropping `then` from `try...else...then..end` maybe a construct like:

```pony
after
  // some block
always
  // code that is executed irrespective of `error`
end
```

Might be useful? Then one could in theory write:

```pony
after
  try
    h.assert_true(FileInfo(tmp)?.directory)
  end
always
  h.assert_true(tmp.remove())
end
```

Or even the follow "partial" code:

```pony
after
  h.assert_true(FileInfo(tmp)?.directory)
always
  h.assert_true(tmp.remove())
end
```

SeanTAllen:
Closing as stale.

PR #14, opened 2016-06-11
Unique promises

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/unique-promise/text/0000-unique-promise.md)

### DIFF
--- a/text/0000-unique-promise.md
+++ b/text/0000-unique-promise.md
@@ -1,0 +1,83 @@
+- Feature Name: unique-promise
+- Start Date: 2016-06-09
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add a `UniquePromise` to the `promises` package mirroring the `Promise` API for sendable objects (`iso`, `val`, `tag`) instead of shareable ones (`val`, `tag`).
+
+# Motivation
+
+The `Promise` API is great but can only be used to produce immutable or opaque results. Having a way to produce mutable isolated results would be really useful.
+
+# Detailed design
+
+A `Promise` can be fulfilled and share its result with an arbitrary number of chained `Promise`s, or be rejected and reject every chained `Promise`. An `UniquePromise` can deal with isolated results so this is not applicable here. A given `UniquePromise` can accept a chained `UniquePromise` only once and further attempts will be rejected. Because that uniqueness is part of `UniquePromise` semantics, trying to chain multiple `UniquePromise`s to the same `UniquePromise` is a logical error from the programmer and we do not need an `already_chained` special case on the notifier.
+
+## Pony interface
+
+`UniquePromise` actor:
+
+```pony
+actor UniquePromise[A: Any #send]
+  """
+  A promise to eventually produce a result of type A. This promise can either
+  be fulfilled or rejected.
+
+  A unique promise can be chained after this one.
+  """
+
+  be apply(value: A)
+    """
+    Fulfill the promise.
+    """
+
+  be reject()
+    """
+    Reject the promise.
+    """
+
+  fun tag next[B: Any #send](fulfill: UniqueFulfill[A, B],
+    rejected: UniqueReject[B] = UniqueRejectAlways[B]): UniquePromise[B]
+    """
+    Chain a promise after this one. If there is already a chained promise, the
+    new promise is rejected.
+    """
+```
+
+Fulfill and reject interfaces and convenience classes:
+
+```pony
+interface iso UniqueFulfill[A: Any #send, B: Any #send]
+  fun ref apply(value: A): B^ ?
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
There's nothing right now that really teaches promises. I think as part of adding anything to Promises that we need to come up with a decent strategy for teaching them.

SeanTAllen:
What are the situations where we would want to be able to send iso's as part of a Promise?

Praetonus:
To teach promises, maybe a Pony pattern would be good?

Getting `iso` from promises can be useful when we want to mutate the result during/after the promise chain.

SeanTAllen:
I think we need a chapter in the tutorial on Promises as a minimum.

I don't really follow the mutate results during/after. When would you want to do that? Can you give examples of situations where that is needed and/or desire-able? I understand this in theory but code demonstrating the value would be great.

SeanTAllen:
Another question. The argument so far for this is to allow mutable values from a promise but, this allows for tag and val as well. If the only reason for this is mutability then it shouldn't be allowing any sendable item. If allowing val and tag has value here then we need to be able to teach people when to use Promise vs Unique Promise which needs to be addressed as part of the RFC.

Praetonus:
A (quick and unpolished) example of this could be a database ORM package. Asynchronous queries would return objects via promises. These objects need to be mutable to allow modifying them directly and committing the modifications into the database.

Regarding `iso` vs `#send`, I've put `#send` here because it was possible. Allowing `val` and `tag` doesn't add anything so I'm not opposed to restricting unique promises to `iso` if it reduces confusion and collisions in promises vs unique promises use cases.

SeanTAllen:
@Praetonus Can you provide example code? I think example code for how this would be used is very important to the RFC. Example code of different use cases will help greatly clarify.

Praetonus:
Here's the simplified [ORM example](https://gist.github.com/Praetonus/915c4e7eede703b874d67d2d83963b08). The promise is used with asynchronous queries and not with behaviours but this is still a form of message sending so this would be the same principle with pure Pony idioms.

Additionally, I have some more thoughts on `iso` vs `#send`. With `iso` only, you can't do things like this:

``` pony
let lbd = recover lambda(x: MyObj iso) =>
  dosomething()
end

uniquepromise.next[None](consume lbd)
```

Since `None` is always `val`, we can't use it for convenience in `next` when using lambdas or functions returning `None`. I think this is a serious drawback because the user will have to return an isolated useless object from every function filled by an unique promise.

SeanTAllen:
Does a promise rather than a notifier make sense in the context of the ORM example? It feels odd compared to the normal notification pattern that is normally used for async operations of that sort.

With the iso vs #send example of None you give, that is an interesting point. I feel like there is a core of an idea but it needs to be fleshed out. I'm still grasping at the ramifications of it and how we would explain to people when to use which type of promise or if it should be two different kinds of promises rather than something about how you set them up.

Another question, can I mix and match Promise and UniquePromises together if they are in fact different types of actors?

Praetonus:
I think a promise makes sense here since it is a request/response pattern and not random exterior events you are notified of.

It should be possible to chain promises after unique promises and conversely. This would be good, I'll have to think about the implementation.

SeanTAllen:
@Praetonus could this be additional methods on the existing Promise or does it need to be a different types?

SeanTAllen:
@Praetonus any updates to this?

jemc:
For whatever it's worth I've been thinking more about promises lately.  I think we need to try to come up with a generic interface for promises to be interoperable with eachother.  Not only is there the distinction between `Promise[A]` and `UniquePromise[A]`, but I've also seen a need to have a promise that returns an `Error` object instead of raising an anonymous `error` (so, `OtherPromise[A,E]`).  It would be great if we had a common interface so that all of these (including other kinds of promises a user might create) were interoperable to the greatest extent that makes sense (basically, being able to chain them to eachother).

Praetonus:
I think the [generic subtyping](https://github.com/ponylang/rfcs/pull/62) could help solve the common interface and chaining problems. I'll come back to this RFC after the other one is implemented.

patternspandemic:
As a new user of Pony, and when I hadn't understood the capabilities requirement of the existing promises API (#share), I tried to use it with the intention of fulfilling a promise with an iso object. Obviously this didn't work out for reasons I now understand, but it may be interesting to note that I attempted to use it in a way that a UniquePromise would allow.

I solved my problem by fulfilling the promise with the shareable data from which I could construct the objects I needed. Basically instead of creating and fulfilling with the object, I created the object inside a fulfillment handler, and passed it along via a behavior call on an actor closed over by the handler.

SeanTAllen:
@Praetonus Should this RFC be closed?

SeanTAllen:
Closing as stale.

PR #15, opened 2016-06-16
Safe Local Variable

hibnico:
As presented in the user and dev mailing list, I think there is an opportunity to improve the type system regarding the proof of some data moves, hence this RFC.
This RFC is incomplete just presenting the motivation, to gather any early comment, in case there are already work on that topic.
Next will be about writing down the ideas I have started to have.

### DIFF
--- a/text/0000-safe-local-variable.md
+++ b/text/0000-safe-local-variable.md
@@ -1,0 +1,86 @@
+- Feature Name: Safe Local Variable
+- Start Date: 2016-06-15
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The proposal aims at making Pony's type system aware that some data pointed by a variable doesn't leak the scope of that same variable, and thus will be able to prove that some local data moves are actually safe.
+
+# Motivation
+
+An issue with a iso or trn data, which have both write uniqueness, is that they don’t survive the call of a _pure_ function. A _pure_ function is a function which has no side effect, and thus shouldn't trouble the guarantees of a reference capability. But since Pony doesn’t know what happens to the data passed as arguments, to guarantee safety, Pony has to expects the worst, a data leak, even if it doesn’t actually happens.
+
+Here is an example of an actor which count spaces and append it to the string being processed.
+```
+actor CountSpaceAppender
+
+  be appendSpaceCount(s: String iso, callback: OtherActor) =>
+    let n: USize = count(s)
+    s.append(n.string())
+    callback.send(consume s)
+
+  fun count(s: String box): USize =>
+    var n: USize = 0
+    for c in s.values() do
+      if c == ' ' then
+        n = n + 1
+      end
+    end
+    n
+```
+
+`count` doesn’t requires more than reading locally the string `s`. All ref caps passed as argument except `tag` could fit that requirement. But since it is an argument of a function, it must be an alias. So it still works for `box`, `val` and `ref`, but not for `iso` or `trn`.
+
+Actually in this example the use of that alias doesn’t make its data leak. The parameter `s` is assigned to no other variable; it is not assigned to any field of the actor (its a box function); even if `s.values()` was exposing its data, it would only leak to another local variable `c`. And transitively the other variable `c` doesn’t leak either. The scope of the data represented by `s` is not goes outside its scope. So a `trn` or an `iso` variable could be passed: at end of the function the guarantee of the `iso` or `trn` are not transgressed.
+
+A work around is to make the function give back the ownership of the arguments by returning them. For instance the previous example can be rewritten as:
+```
+actor CountSpaceAppender
+
+  be appendSpaceCount(s: String iso, callback: OtherActor) =>
+    (let n: USize, s) = count(consume s)
+    s.append(n.string())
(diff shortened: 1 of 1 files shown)


SeanTAllen:
iso means "only alias", what you seem to be asking for is to change the definition of iso, which I find odd. that doesn't really make sense to me.

Reading between the lines, if this did what you want, you now have a public function `count` whose signature could remain the same but due to a code change, could break untold number of callers. Further, at the local site, I would have no way of knowing what I might be breaking. Changing the implementation of a method without changing the function signature shouldn't be able to cause code that previously compiled to fail.

hibnico:
I have not written about any solution yet, I am only exposing here some example of code that the type system is refusing to compile whereas the data safety guarantee is not actually broken.

I'll write about some solution I have in mind soon, unless somebody tells me it has been already worked on. And it is not about changing `iso`, more probably about introducing a new ref cap modifier, but we'll see because ideas often change as they start being written down.

hibnico:
RFC updated
There are still some work to do to be sure that the proposed solution is actually a good one, but here is the current state of the ideas I have.
https://github.com/nlalevee/rfcs/blob/safe-local-variable/text/0000-safe-local-variable.md

hibnico:
Updated again, with a more precise design, especially on the nature of the concept being introduced. I think it would work nicely.

sylvanc:
How does your approach handle this?

``` pony
actor Main
  new create(env: Env) =>
    let x: Foo iso = Foo
    let y: Foo iso = Foo
    scoped(x, y)
    // CRITICAL PROBLEM
    x.f is y.f // this is true, so isolation has been violated

  fun scoped(x: Foo ref~, y: Foo ref~) =>
    x.f = y.f

class Foo
  var f: Bar = Bar

class Bar
  var f: U32 = 0
```

Also, based on your motivation, why not use a recover expression? That would allow using the `count` function unmodified. For example:

``` pony
actor CountSpaceAppender
  be appendSpaceCount(s: String iso, callback: OtherActor) =>
    var n = USize(0)
    let s'' = recover
      let s': String ref = consume s
      n = count(s)
      s'.append(n.string())
      consume s'
    end
    callback.send(consume s'')

  fun count(s: String box): USize =>
    var n: USize = 0
    for c in s.values() do
      if c == ' ' then
        n = n + 1
      end
    end
    n
```

hibnico:
> How does your approach handle this?
> [...]

Hum, good catch. Just before submitting the last update I was starting to think about such case, like the one in the paragraph before the typing rules.
So something is definitively off in the suggested design. I am probably too optimistic in the subtyping rules. It definitively needs more work, it needs to be more precise.

> Also, based on your motivation, why not use a recover expression? That would allow using the count function unmodified. For example:
> [...]

The recover only works with sendable types, and when playing with fields it doesn't work nicely. The original troubling examples I had are described here:
https://pony.groups.io/g/user/thread/moving_trn_data/1896256?p=,,,20,0,0,0::RecentPostDate/Sticky,,,20,2,0,1896256
https://pony.groups.io/g/user/thread/calling_a_ref_function_on_a/1934548?p=,,,20,0,0,0::RecentPostDate/Sticky,,,20,2,0,1934548

I wanted to wrote an even simpler example for this RFC, probably should have keep examples closer to what I have encountered.

hibnico:
I have reworked the suggested design, I have introduced two new reference capabilities. It gets then more sane to reason about. This still need more work, but I think this is going somewhere nice, thus I am sharing it here for early comment and review.

SeanTAllen:
I think this will be very difficult to teach and as such I want to see thought going into how to teach this before I discuss the merits or lack thereof.

jemc:
My main concern is that the rest of the ref cap system currently has a formal proof, and I wouldn't want to add something as important as this without first subjecting it to the same rigor.

Note that the current reference caps that exist are the only useful ones that can exist along the currently defined dimensions of what reference caps can dictate.  There are some additional ones that _can_ exist along the same dimensions, but wouldn't be useful (for example, a write-only reference that cannot read).  I was trying to find the old ticket where @andymcn drew out this explanation and went through every possible reference capability along the currently defined dimensions and explained for each cap that didn't exist, why it wasn't useful.  At any rate, the point is that you're defining a new _dimension_ of reference capability, and that's something that we'd want to have a formal basis for.

hibnico:
I share your concern about the formality of the suggested new ref caps. Once this RFC complete, even if I think that adding these new ref caps would be a nice improvement, I wouldn't push them if there are as well validated as the existing ones. This RFC is a work in progress, doing it as well as I can, following my intuition.

At first I didn't want to add new reference capabilities, I was suggesting just a new modifier on the existing ones. But it was harder to reason about because for some of the ref caps, this modifier wasn't making sense, and I got lost, and Sylvan spotted some bad consequence. And actually, as I wrote in the proposal, these new ref caps are not different from the existing ones (`box` and `ref`) when only looking at their deny properties. What differs is how they can be aliased.

hibnico:
I have read papers and tried to figure out the formal representation of the new reference capabilities. And it is out of my league, especially considering the time I want to spend on it. So let's close this RFC.

PR #17, opened 2016-06-22
Numeric Parsing

ghost:
Proposed changes and additional package for parsing numeric values from strings.

### DIFF
--- a/text/0000-numeric-parsing.md
+++ b/text/0000-numeric-parsing.md
@@ -1,0 +1,325 @@
+- Feature Name: number_parsing
+- Start Date: 2016-06-21
+- RFC PR:
+- Pony Issue:
+
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I don't understand the motivation section as written.

What is the problem we are solving? Its unclear to me which makes it hard comment on the proposed solution (because I dont understand exactly what problem it is intended to solve).

In particular why do we want to:

`separate the functionality of numeric parsing and the relevant error checking away from the built-in String class.`

The rest of the motivation doesn't address this.

SeanTAllen:
This RFC will be closed at the end of June unless it sees movement before then.

rhagenson:
Closing as there was no movement on the topic or response from the author clarifying intent.

PR #18, opened 2016-07-03
http-status

Theodus:


jonas-l:
I would suggest getting rid of `Status` prefix. Consider the example:

``` pony
if response.status() isnt StatusOK then error end
```

I think it would look much nicer without repeating a _status_ twice:

``` pony
if response.status() isnt OK then error end
```

It can be achieved by having statuses in a separate package `net/http/status`. In worst case scenario a named import would give almost the same code as in initial proposal:

``` pony
use status = "net/http/status"

// ...

if response.status() isnt status.OK then error end
```

Theodus:
I personally prefer the Status prefix, but I am open to this change if more people would prefer that.

jemc:
I'd personally lean toward keeping the `Status` prefix, and defining a `trait Status` that each primitive provides.  I really like this pattern when working with a large group of simple "enum-style" types that all share a common trait.  See [here](https://github.com/jemc/pony-ast/blob/d5f97ff85153c19aab8289742f762b6832605e7a/ast/tk.pony) for an example of it in action.

SeanTAllen:
I prefer without leading Status.

jemc:
I would note that the prefixed-enum-style has precedence in the current official pony codebase as well.

In [tostring.pony](https://github.com/ponylang/ponyc/blob/afc74c61e500dbc5e5e04ce227f978cb52c777cc/packages/builtin/tostring.pony) there are several collections of types which demonstrate the pattern of prefixing the enum-style types with a prefix that corresponds to the trait, interface, or type union which contains them.

There may be other examples of this in the official pony codebase too (I would imagine so, since @andymcn seemed to like that pattern, and contributed much of the official codebase we have right now), but I just mentioned this one because it came to mind while discussing another RFC.

Theodus:
This should probably move to the final comment period.

SeanTAllen:
@Theodus can you announce this on pony+dev

SeanTAllen:
@sylvanc please have a look

SeanTAllen:
@ponylang/committer any input?

aturley:
@Theodus a few things:
- Is there any sort of consideration for how to handle custom status codes?
- Can you provide a simple, complete before and after examples of how this is currently done and how it would change? (I think I know, but I haven't spent any time with the http code, so I'm not 100% sure.)

Theodus:
@aturley 
- Custom status codes may be defined by the user, they just have to make a primitive implementing the Status trait.
- An example would be as follows:

``` pony
// before
Payload.response(201, "Created") 

// after
let stat = StatusCreated
Payload.response(stat(), stat.text())
```

jonas-l:
I think http API should be changed to use `Status` instead of _code, message_:

``` pony
Payload.response(StatusCreated)
```

SeanTAllen:
In general, I agree with @jonas-l, if this abstraction exists, it makes sense to use it as he suggests.

Perelandric:
Would it be worth adding a `string()` method that returns the same value as `text()`?

The name `string()` is more consistent with Pony's standard library, though the term "status text" seems pretty commonly used in HTTP implementations.

aturley:
@Theodus I'm looking at `PayloadBuilder._parse_response` (https://github.com/ponylang/ponyc/blob/master/packages/net/http/payloadbuilder.pony#L133), which parses the response header and sticks the raw values into the payload fields. It seems that something would need to be changed to make this work with user-defined status codes, but maybe I'm wrong. Is there something that I'm missing that makes this unnecessary, or is there a plan for addressing it?

malthe:
`string()` could also return the string `"<code> <text>"` – e.g. `"200 OK"`.

sylvanc:
I like `string()` giving the whole thing ( @malthe ), and I like passing the primitive, not the value and the text ( @jonas-l ).

@jemc is right that there are instances of `PrefixThing` in the `GeneralSpecific` pattern right now. On the other hand, using package scoping and module importing to provide those prefixes does seem pretty neat.

I'm ok with a `Status` prefix or without one. The former is more verbose but always unambiguous, the latter is more concise and can be made unambiguous when necessary.

The current string formatting stuff does it because it is in `builtin`, so the lack of a prefix would pollute all packages. Perhaps the right thing there is to figure out how to move string formatting into its own package? I'm not sure about that.

Implementing a `Status` trait ( @jemc ) is great. That makes status codes declared rather than implicit (which is good: not everything with a value and some text is an HTTP status code!), but it leaves them open-world so that custom status codes can be created ( @Theodus ).

jemc:
To me, the presence of the `Status` trait is what makes me want to use `StatusWhatever` as the format for the name.

If we used named imports, then the `Status` trait becomes weird.  For example, if we `import status = "net/http/status"`, then we have to refer to `status.Status`, which is pretty unfortunate.

> Perhaps the right thing there is to figure out how to move string formatting into its own package?

I think we've talked about wanting to do this anyway - make `Stringable.string` take no arguments and make a separate package for printing various objects (mostly/only numbers?) with specific formatting options.

SeanTAllen:
> If we used named imports, then the Status trait becomes weird. For example, if we import status = 
> "net/http/status", then we have to refer to status.Status, which is pretty unfortunate.

I'm ok with that given that it is probably a less frequent case. Python from my time with it in the 90s sort of beat me into being ok with that.

Theodus:
@aturley The only change that needs to be made is in the `Payload.response` constructor (https://github.com/ponylang/ponyc/blob/master/packages/net/http/payload.pony#L31) as follows:

``` pony
new iso response(status': Status = StatusOK,
    handler': (ResponseHandler | None) = None)
  =>
    """
    Create an HTTP response.
    """
    handler = handler'
    status = status'()
    method = status'.string()
    url = URL
    _response = true
```

Theodus:
I think that this is ready for a final vote.

PR #19, opened 2016-07-05
fmt

Theodus:


malthe:
Can you put in some examples in the motivation that highlights how the unnecessary complexity is dealt with?

hibnico:
I think it would be particularly useful to lazy evaluate a message to be sent to Debug.out.

SeanTAllen:
A couple requests and a couple questions:

When you say that this should be taught in the tutorial, where would that go in the tutorial? When I glance at the tutorial I don't see an obvious place for it to go. Further, I'm not sure where in the flow of the tutorial it would go.  I'm not saying it shouldn't go in the tutorial, rather that the current tutorial format doesn't really appear to do a good job of teaching features like this and I think that a rethink of tutorial might be in order for where such material goes (or perhaps this goes in a follow on to the tutorial).

The examples are pretty simplistic. I'd like to see more advanced usages with more "real world code". Can you show a couple before and after examples of how this removes complexity? 

Are there scenarios where this is more complicated than the existing?

How does this work with things like formatting as hexadecimal that exist in some current number formatting code?

Why is the primitive separate? Why not as part of string?

What is the unnecessary complexity of the existing solution?

Theodus:
With the current layout of the tutorial it would probably have to be a "string formatting" section of the Appendices. Instead of this, I think that another tutorial could be created that focuses on teaching Pony by example, similar to [Go By Example](https://gobyexample.com/) or [the Rust book](https://doc.rust-lang.org/book/).

SeanTAllen:
Re: complexity...

I don't really see from looking at the examples how this removes complexity. It looks at least as complex to me, if not more complex. How is this reducing complexity?

Re: tutorial

the Appendices are very tacked on. I like the "by example" idea a bit. I would imagine that the tutorial is "Pony for the impatient", would a "by example" build off of the tutorial or is it a complete fresh start where we don't assume that the person has read the tutorial? The latter might be a little hard to get off the ground.

Theodus:
Re: complexity

Even if you don't see a reduction in complexity, I feel that this approach would be much easier to teach since it resembles string formatting from every other language that I have used (except maybe Kotlin).

Re: tutorial

I think that it should build off of the tutorial so that it can focus on getting a sense of how to use the language without the need to explain subjects like reference capabilities and other type system guarantees. This could maybe allow the tutorial to focus more on the actor model and capabilities while the "by example" tutorial would include subjects like the C-FFI and use of common packages like time and collections.

SeanTAllen:
I don't think there is a great hurdle to teaching the existing means. If we are going to break a bunch of code (even pre 1.0), we need a good reason. I don't see any reduction in complexity. 

Would this deprecate the existing method of formatting strings? Remove it? Or leave it in place?

Theodus:
This does not necessarily have to replace the existing method of formatting strings, the example implementation works separate from the current method.

Theodus:
My position is that the string() method should take no arguments because its job is to produce a string representation of the object in its current state. Formatting should be handled externally since it contributes nothing to the string representation of the object.

jemc:
> My position is that the string() method should take no arguments because its job is to produce a string representation of the object in its current state. Formatting should be handled externally since it contributes nothing to the string representation of the object.

I'm aligned with this thinking, and this is also aligned with a direction we discussed in a recent sync call.

However, I personally was picturing an approach different from this one.  I was picturing a package for rendering numbers in various formats based on the given arguments, which would be similar to those currently given in `FormatSettings`.  That is, it would have a focus on converting numbers to strings in various formats, rather than one acting as a sort of `printf`-like string-composer from various disparate parts.

I worry a bit that the direction of this RFC as proposed is a bit too "dynamic" in nature to be a good pony pattern.  So far these are only vague feelings not rooted in any concrete implementation concerns, since I haven't spent much time or effort thinking more deeply about this.

Theodus:
@jemc How would you feel if the Fmt primitive was separated into two primitives, one for padding and alignment of stringables and the other for numbers? The one for numbers could also be split further into specific integer and float formatting.

jemc:
The main thing I was expressing doubt about was the string-interpolation approach, but again, it's only a vague doubt and not anything certain or decisive at the moment.

SeanTAllen:
Its an excellent point. I haven't used the string formatting that currently exists much. Is is possible to get runtime errors? Could we get runtime errors with this approach?

Theodus:
I will create a new RFC with a solution that fits in better with the current string formatting.

PR #20, opened 2016-07-06
Updates to RFC process re: sidestepping

SeanTAllen:
Per conversations at the last sync meeting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,18 @@ Some changes do not require an RFC:
    - Additions only likely to be _noticed by_ other developers-of-pony,
 invisible to users-of-pony.
 
+What is a bugfix?
+
+Bugfixes do not have to go through the RFC process. The definition of "bugfix" is still evolving. Some things we consider to be bug fixes:
+
+    - Fixing performance issues that are found in existing implementations. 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@sylvanc @aturley @jemc does this match your understanding from previous sync meeting?

aturley:
I think this statement should apply to all the bugfix items, not just performance bugs:

> Note, depending on level of API breakage, there might need to be some discussion around this.
> I say this with the assumption that some of those other things could cause breaking changes.

Otherwise it seems fine.

jemc:
My first thought is that the sidestepping cases we've discussed so far would be addressed by points 2 and 4 alone, and that points 1 and 3 may be more broad and prone to overreach.  My current inclination would be to leave points 1 and 3 out until we have a concrete real-world example of a case that we think _should_ sidestep, but doesn't meet the criteria in points 2 and 4.

That is, it seems to me like the "principle of least surprise" and "performance" criteria would cover most or all of the "API bug" or "omission" cases where we would want to sidestep, but I'm open to changing that thought if I see some evidence of a real case that is not covered.

We can certainly discuss this more in the sync call if we need to.

SeanTAllen:
API bug came from @sylvanc. Thoughts @sylvanc ?

I'm ok with dropping omission for now.

PR #21, opened 2016-07-09
io

Theodus:


SeanTAllen:
In terms of alternatives, an alternative would be to not do this.

That said, I think this is a reasonable idea. The standard needs to be using more interfaces that will make it easier to customize.

Re: moving Write and Read buffer, I was thinking yesterday that they weren't net specific. Moving them makes sense to me.

Important question, are there any places that we should change from using concrete classes to taking the interface if adopted?

Theodus:
I'm currently working on additions to the encode package and I was thinking that there could be an Encoder interface that writes to any Writer and a Decoder interface that reads from any Reader.

Theodus:
Currently that would affect the encode/Base64 package.

SeanTAllen:
I think it would be a very good idea as part of this RFC to document the changes that need/should take place as part of this change.

Adding those on to `Detailed Design` as you did with the Buffers makes sense.

text/0000-io.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Detailed design
+
+- The io package will contain the following interfaces:
+```pony
+interface Readable
+  """
+  Wraps the read method which returns up to len bytes from the input.
+  """
+  fun read(len: USize): Array[U8] iso^
```
jemc: There's no type in the pony standard library so far that conforms to this interface.

`File.read` is close, but the receiver cap is `ref`, not `box`.

The receiver cap is `ref`, because it has to move the cursor, which is state mutation.

text/0000-io.md
```diff
@@ -0,0 +1,64 @@
...
+```pony
+interface Decoder
+  """
+  Wraps the decode method which reads encoded data from the Reader and returns the decoded bytes.
+  """
+  fun decode(r: Reader): Array[U8] iso^
+
+interface Encoder
+  """
+  Wraps the encode method which writes the encoded data to any Writer
```
jemc: I don't think it's accurate to say that an interface "wraps" the methods.  These docstrings should just say what the method does, instead of saying that they "wraps the method that does X".

jemc:
The `Readable` interface doesn't seem right to me, as most "reading" in Pony is not synchronous - we have `File.read`, which is synchronous, but that is the exception rather than the rule.  Most Pony input is received with messages passed to notifiers.

I think making the `Readable` interface a synchronous one obscures the fact that asynchronous I/O is more relevant in Pony - this is likely to be confusing to newcomers, and make it more difficult to teach about asynchronous I/O.

jemc:
I think the `Writable` interface is missing a `writev` definition, which is very important to efficient zero-copy I/O.  If you're worried about making sure to support types that don't implement `writev`, you can provide a default implementation in the interface which just loops over the `ReadSeq[ByteSeq]` and calls `write` for each one (just as the current stdlib implementations of `writev` do).  However, we eventually want all major Pony writer types to use the `@writev` system call wherever possible (or windows equivalent).

Theodus:
Should a second set of interfaces be made for asynchronous I/O? What would that look like?

jemc:
Just realized that the current `Writeable` interface won't match any of the `write`/`writev`-implementing types that are actors (most of them, I think).  That's another confusing point where the `Writeable` interface won't match (like the notify-based types where `Readable` won't match).

SeanTAllen:
I like the idea of using interfaces in the standard library so users can swap in new implementations. I think however that we need to drive these from specific use cases where we can make the standard library more flexible and extensible by introducing a new interface. 

@Theodus I'd like to know where this would be used in the existing library. What places will be changed from using a concrete type to this interface?

Theodus:
I think that it would be best to close this PR for now and rethink its goal.

PR #22, opened 2016-07-21
Object copying idiom extension

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/copy-idiom/text/0000-copy-idiom.md).

### DIFF
--- a/text/0000-copy-idiom.md
+++ b/text/0000-copy-idiom.md
@@ -22,7 +22,17 @@ class Object
   fun clone(): Object iso^
 ```
 
-This interface should only be a basic template and per-class modifications should be allowed, such as changing the return capability of `clone`. It should also be allowed to not provide `clone` if every operation can be done with `from` (for example in numeric primitives).
+This interface should only be a basic template and per-class modifications should be allowed, such as changing the return capability of `clone`.
+
+The `from` constructor on a given type `A` should be allowed to take objects of another type `S` if it makes sense to construct objects of type `A` from objects of any subtype of `S`.
+
+On collections, `from` and `clone` should perform deep copies. We'll add a `Cloneable` interface to ensure that contained objects are indeed cloneables. Collections should provide a `shallow_from` constructor and a `shallow_clone` function for shallow copies.
+
+## Changes to existing functions in the standard library
+
+- `collections.List.from`, signature: `new from(seq: Array[A^])`. Would be modified to `new from(that: List[A] box)`. It is possible to construct `List`s from `Array`s by using `List.concat`.
+- `collections.persistent.Lists.apply`, signature: `fun apply(arr: Array[val->A]): List[A]`. Would be removed. This function is redundant with `from`, which is taking a more generic `Iterator`.
+- `collections.persistent.Maps.from`, signature (abridged): `fun from[K, V](pairs: Array[(K, V)]): Map[K, V]`. Would be modified to take an `Iterator`.
 
 # How We Teach This
 
@@ -30,7 +40,8 @@ Advocating for the use of this idiom in user libraries would be good for the con
 
 # Drawbacks
 
-This can introduce some code duplication between `from` and `clone`.
+- This can introduce some code duplication between `from` and `clone`.
+- The new shallow copy vs. deep copy distinction would break existing code.
 
 # Alternatives
 
--- a/text/0000-copy-idiom.md
+++ b/text/0000-copy-idiom.md
@@ -1,0 +1,41 @@
+- Feature Name: copy-idiom
+- Start Date: 2016-07-21
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Extend the current idiom for object copying in the standard library with a `from` constructor in addition to the `clone` function.
+
+# Motivation
+
+The current "standard" way of copying an object is with the `clone` function (usual signature: `fun clone(): Object iso^`). While this function is very useful in some situations, like copying an `iso` object without consuming said object, it has limitations. For example it isn't possible to construct an embedded field with a copy of another object. Adding copying constructors would solve that problem.
+
+# Detailed design
+
+The copying functions should be provided as:
+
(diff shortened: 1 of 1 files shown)


jemc:
@Praetonus - are there any cases of `new from` in the current standard library that would be incompatible with this meaning of `new from` (for example, if they take other arguments)?

Other than that, this seems reasonable.

Praetonus:
There are 4.
- `collections.List`, signature: `new from(seq: Array[A^])`. A container of ephemereal objects doesn't really make sense so I think we should replace it. Being able to construct collections from other kinds of collections is good, so changing the function to take only `List` isn't the best idea. It could be `Seq` but think `Iterator` is even better as it means "build a collection from that arbitrary succession of objects" (we would do the same thing for `Array.from`). To generalise that, `from` on a type `A` should be allowed to take some type `S` instead of `A` if it makes sense to construct `A`s from every subtype of `S`.
- `capsicum.CapRights0`, signature `new from(caps: FileCaps box)`. Given what was said above, this function is compatible with the RFC.
- `collections.persistent.Lists`, signature: `fun from(iter: Iterator[val->A]): List[A]`. This isn't exactly a Pony constructor as it is on a primitive but it is constructing new things (persistent lists). The strange thing is that there is an `apply` function on the primitive constructing a persistent list from an `Array`. For consistency with the above rule for collections, we should remove `apply` and keep `from` as is.
- `collections.persistent.Maps`, signature (abridged): `fun from[K, V](pairs: Array[(K, val->V)]): Map[K, V]`. Same thing as above, allowing only `Array` isn't the best. We should change that to `Iterator`.

I'll update the RFC with the possibilities for the parameter types for `from` and the planned modifications to existing functions in the standard library.

jemc:
> Being able to construct collections from other kinds of collections is good, so changing the function to take only `List` isn't the best idea. It could be `Seq` but think `Iterator` is even better as it means "build a collection from that arbitrary succession of objects" (we would do the same thing for `Array.from`).

Note that we already have an idiom for that, though - see `Array.concat` and `List.concat`, which take an `Iterator`.

jemc:
I think this approach (at least the original one outlined in this RFC, I know you're in the middle of making changes) may solve my oft-gripe about not being able to get an `iso` out of `Array.clone`.  I'll have to give it a try later and see if it solves my woes.

Praetonus:
I forgot about `concat`. I don't think a constructor from `Iterator` would be better than `create` followed by `concat`, so having `from` on collections take an object of the same type looks like the best solution as it would allow per-type optimisations such as `Array` copying chunks of memory instead of iterating the whole collection.

Regarding `iso` and `Array.clone`, I think it can be addressed in this RFC by fleshing out `clone` semantics a bit more. I'm thinking `clone` should do a deep copy (same for `from`). For collections, it would mean adding a `Cloneable` interface, with `clone` only callable on collections of `Cloneable`s. The current meaning of `clone` for collections (shallow copy) would be moved to a `shallow_clone` function. Maybe a `shallow_from` constructor would also be good, I'll see if it adds anything.

jemc:
Thing is, I usually don't really _want_ a deep clone when I want to get an `iso` out of the deal.  The capabilities for this work okay on `val` (and `tag` too, I guess), so theoretically I should be able to make an `iso` `clone` of any collection of `#share`ables, even without doing a deep clone.

For example, in my mind, I should be able to clone a `Array[String val] ref` to a new `Array[String val] iso` without also copying all of the underlying `String`s.  This is super important for zero-copy messaging and scatter-gather IO, where dealing with sendable collections of `ByteSeq`s is the order of the day.  I don't want to copy all of my `ByteSeq`s if its possible to be safely shared.  I want to save all of those extra allocations as much as I can.

Praetonus:
I see. I think what would solve your problem would be (per-capability) generic specialisation. I guess that would be a big chunck of work, I don't know if it is somewhere on the roadmap or if it would require an RFC.

I'll add the shallow/deep copy stuff to the RFC as I think it would be useful to have it in other situations.

Praetonus:
I think this should go to the final comment period.

SeanTAllen:
@Praetonus can you send an email re this RFC going into the final comment period to the pony dev mailing list?

SeanTAllen:
Where in the tutorial would we discuss object copying? I would like to have a reasonable idea as part of RFCs for what needs to be added to the tutorial and where. Do we need a new section of an existing chapter? I new chapter? Is there additional supporting material that needs to be added at the same time?

Can you flesh out the teaching story a little @Praetonus?

Praetonus:
@SeanTAllen I've added some details on that.

SeanTAllen:
@Praetonus i don't think the reasoning behind:

"For example it isn't possible to construct an embedded field with a copy of another object. Adding copying constructors would solve that problem."

is obvious. It certainly isn't for me. Could you expand upon that idea? I think having more detail around that will be important when it comes time to accept/reject this and if it is accepted, when implementing.

SeanTAllen:
In this sentence:

"This interface should only be a basic template and per-class modifications should be allowed, such as changing the return capability of `clone`."

I'm unsure what "such as changing the return capability of `clone`" means. Can you clarify?

SeanTAllen:
@Praetonus i like the idea of an "idioms" section for the tutorial (or perhaps standalone after the tutorial). perhaps discuss the idioms idea in more detail at an upcoming sync meeting?

Praetonus:
@SeanTAllen Yes, discussing it at a sync meeting could be good.

SeanTAllen:
@Praetonus once that conversation is done, do you feel this is ready to move out of final comment period and into ready to vote?

Praetonus:
@SeanTAllen I think it is ready.

SeanTAllen:
@Praetonus ok, we have a full slate for today so this will make it onto next week's agenda.

sylvanc:
@jemc and I are voting yes!

PR #23, opened 2016-07-28
String Formatting

Theodus:
[Rendered](https://github.com/Theodus/rfcs/blob/format/text/0000-format.md)

### DIFF
--- a/text/0000-format.md
+++ b/text/0000-format.md
@@ -1,0 +1,87 @@
+- Feature Name: format
+- Start Date: 2016-07-27
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the seperation of the `string()` method and formatting. This wil remove formatting from the builtin package of the standard library and seperate the disperate concerns of producing a string representation of an object and formatting strings.
+
+# Motivation
+
+The seperation of string formatting will allow packages to implement string formatting that may not be relevant to general formatting. This may enclude date and time formatting or JSON pretty printing. Pony's string formatting should also have improved documentation since it differs from that of most other languages.
+
+# Detailed design
+
+- Redefine the Stringable interface:
+```pony
+interface box Stringable
+  """
+  Things that can be turned into a String.
+  """
+  fun string(): String iso^
+    """
+    Generate a string representation of this object.
+    """
+```
+- Modify all classes that currently implement the Stringable interface to reflect the new interface
+- Allow numbers to keep implementing the interface by modifying the `_ToString` primitive in builtin to create strings in base 10 only.
+- Create a package `format` that includes a `Format` primitive with the following methods:
+```pony
+primitive Format
+  """
+  Provides functions for generating formatted strings.
+
+  * fmt. Format to use.
+  * prefix. Prefix to use.
+  * prec. Precision to use. The exact meaning of this depends on the type,
+  but is generally the number of characters used for all, or part, of the
+  string. A value of -1 indicates that the default for the type should be
+  used.
+  * width. The minimum number of characters that will be in the produced
+  string. If necessary the string will be padded with the fill character to
+  make it long enough.
+  *align. Specify whether fill characters should be added at the beginning or
+  end of the generated string, or both.
+  *fill: The character to pad a string with if is is shorter than width.
+  """
+  fun apply(str: String, fmt: FormatDefault = FormatDefault,
+    prefix: PrefixDefault = PrefixDefault, prec: USize = -1, width: USize = 0,
+    align: Align = AlignLeft, fill: U32 = ' '
+  ): String iso^ =>
+
+  fun int[A: (Int & Integer[A])](x: A, fmt: FormatInt = FormatDefault,
+    prefix: PrefixNumber = PrefixDefault, prec: USize = -1, width: USize = 0,
+    align: Align = AlignRight, fill: U32 = ' '
+  ): String iso^ =>
+
+  fun float[A: (Float & FloatingPoint[A])](x: A,
+    fmt: FormatFloat = FormatDefault,
+    prefix: PrefixNumber = PrefixDefault, prec: USize = 6, width: USize = 0,
+    align: Align = AlignRight, fill: U32 = ' '
+  ): String iso^ =>
+
+```
+- The following standard library packages need minor updates: crypto, json, net/http, serialize, and strings
+
+Example use of `Format`:
+```pony
+Format.int[U32](3, FormatDefault, PrefixDefault, 3, 5))
+// => "  003"
+```
+
+# How We Teach This
+
+Formatting will not change, but the way that it is applied to strings will. Explanations and examples will be given in both the package documentation and in the Pony tutorial.
+
+# Drawbacks
+
+- Exisiting code will be broken
+
+# Alternatives
+
+- Not to do this
+
+# Unresolved questions
+
+None


text/0000-format.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Detailed design
+
+- Redefine the Stringable interface:
+```pony
+interface box Stringable
+  """
+  Things that can be turned into a String.
+  """
+  fun string(): String
```
jemc: Note that the current standard is for `string` to return `String iso`, rather than `String val`.  I'm not sure which of these I prefer at the moment - I have mixed feelings because they have their advantages and disadvantages.

Using `iso` is nice when the string is being built from scratch on every call, because it gives more flexibility in the caller - it can become any other capability when consumed.

Using `val` is nice when the string doesn't need to be built from scratch on every call, because it can avoid unnecessary copies.  For example, a class could hold a `String val` internally that is returned whenever `string` is called.  If the return value of the interface were `String iso`, then a new string would have to be allocated and copied on every call, since you can't get an `iso` reference from a `val` one, and even if you could, it couldn't be shared across multiple calls.

jemc: Thinking more about it, it seems like `val` is more specialized - more often, `iso` will be appropriate since the string will be built from scratch on every call.  In this paradigm, the best way to avoid allocating multiple strings is simply to have them be held onto by the caller.

If a class had a strong use case for returning a `String val` representation, it could provide another method, which would not provide the `Stringable` interface.

So, I think `String iso` is indeed more appropriate here.

text/0000-format.md
```diff
@@ -0,0 +1,64 @@
...
+primitive Format
+  fun apply(str: String, fmt: FormatSettings = FormatSettingsDefault): String =>
+  
+  fun int[A: (Int & Integer[A])](x: A,
+    fmt: FormatSettings[FormatInt, PrefixNumber] = FormatDefaultNumber
+  ): String =>
+  
+  fun float[A: (Float & FloatingPoint[A])](x: A,
+    fmt: FormatSettings[FormatFloat, PrefixNumber] = FormatDefaultNumber
+  ): String =>
```
jemc: These should probably be `String iso` to give the caller more flexibility.  The strings are being built from scratch each time, so there is no loss on the receiver's end.

SeanTAllen:
Before I would sign off on this, I need to see more in the motivation:

"String formatting should not be involved in the string representation of an object. Pony's string formatting should also have improved documentation since it differs from that of most other languages."

I'm left asking "why?" after reading that. 
Why should string formatting not be involved in the string representation of an object.

Let's assume that we accept this RFC, this sounds like it has important design ramifications for the standard library in general but the reasoning isn't at all laid out.

SeanTAllen:
Under how to teach this,

can you lay out what needs to change in existing documentation and where formatting should be taught going forward? Is this updating some Ponydoc and calling it a day? Does it belong in the tutorial somewhere? If yes to either or both, where and what? If no, then how would it be taught?

SeanTAllen:
From

"fix packages in the standard library that use string formatting (crypto, json, net/http, serialize, and strings)"

is that the complete list of packages that need to change? I'm guessing it is, but it isn't clear. If it is the complete list, could you make it explicit? Something like:

"the following standard library packages need to be updated.. X, Y, Z".

as well as a general level of effort involved in changing them. 

Are there corresponding tests for this? Ideally, if not, then we should be adding tests around this functionality before changing. (or adding tests to verify the new functionality)

jemc:
In general, I'm in favor of a change like this, and it's something we've discussed favorably on a few sync calls.  However, I'm of the opinion that we shouldn't keep the "container" format for the parameters.

To give a bit of history on this, the `Stringable.string` interface used to contain a lot of parameters, oriented (I would argue) mainly toward formatting numbers.  The method signature was always several lines long, and had to be copy-pasted for every `.string` method, even when these parameters weren't used (as was the case for pretty much all non-number user-defined types).

I think this was seen to be cumbersome, so a change was made in https://github.com/ponylang/ponyc/commit/e0ecfc15048b00e19e5c7a179f44f40f7e3d1c1c that moved these many parameters into `FormatSettings` containers.  This change certainly reduced the verbosity of the method signatures, making things easier on the receiving type implementation, but making things more complicated for the caller - instead of simply passing the appropriate arguments, they now had to construct an object with the appropriate fields (or even more cumbersome, a custom primitive).

This new approach of printing numbers with the `Format` type removes the problem of having a verbose method signature of every `string` method, so it also removes the need for the `FormatSettings` family of objects (in my opinion).  I think we should remove the `FormatSettings` containers, and use the indiviudal parameter form used before https://github.com/ponylang/ponyc/commit/e0ecfc15048b00e19e5c7a179f44f40f7e3d1c1c.

SeanTAllen:
@jemc I think that is excellent background and motivation and should be added to the RFC.

Perelandric:
I hope I'm not deviating too far from the subject of this RFC, but I think it would be nice if interfaces in Pony could allow a method that has extra params not described in the interface, to be a valid fulfilment of the interface, as long as those params have default values.

So if the interface is:
``` pony
interface box Stringable
  fun string(): String iso^
```

...all of the following method signatures would be considered a fulfillment of the interface:
``` pony
fun string(fmt: FormatSettings = FormatSettingsDefault): String iso^

fun string(indent: String = "", pretty_print: Bool = false): String iso^

fun string(): String iso^
```

Because all of the methods are able to respond to a `string()` call, they would correctly implement that interface.

Whereas if any of the parameters did not have a default value, the interface requirements would not be fulfilled.

Theodus:
The entire RFC has been implemented (including the most recent changes) at [https://github.com/Theodus/ponyc/tree/str](https://github.com/Theodus/ponyc/tree/str)

Theodus:
Before pushing this forward, I would like to hear from either @SeanTAllen or @jemc about how this change to formatting will be affected by the future changes to Strings that have been discussed.

SeanTAllen:
@jemc has a better handle on that than I do. I think I'll leave that to him.

malthe:
This is an orthogonal concern, but it's often argued that getters and setters have no place in object-oriented design – i.e. `set_precision(...)` and `set_width(...)` in the examples.

malthe:
What's wrong with `String.format(...)` – ? It's convenient and ergonomic.

SeanTAllen:
@jemc can you weigh in on @Theodus's question?

jemc:
@Theodus it's hard to say exactly how this will be affected by future changes to how strings are represented in Pony, but I think the answer is probably "not much".  There's no RFC yet, but as I see it there are two general directions in which we could move strings in the future (or a third direction of not changing it at all):
1. `String` as a container for an `Array[U8]` plus an encoding setting.
2. `Array[U8]` as strings, passed to methods of primitives that know how to treat bytes like strings with a given encoding.  Perhaps `String` would be a type alias for `Array[U8]`.
3. Change nothing about `String`.

Under all three of these options, I don't think the signature of the methods you outlined would change.  At most, they might return an `Array[U8]` instead, if we do option 2 without the type alias.

Theodus:
In that case, I think that this should be moved into final comment period. Do you agree @SeanTAllen?

malthe:
@jemc with regards to what could change to Pony's string implementation, a [short string optimization](https://github.com/elliotgoodrich/SSO-23) may be relevant.

SeanTAllen:
@malthe that's off topic for this RFC.

malthe:
@Theodus, I definitely agree with removing formatting from `string()` – since it makes the `Stringable` interface apply to a lot of classes implementing this method  – but I don't understand why `Format` has to be a primitive.

I think the following is more obvious:

``` pony
let x: U32 = 3
x.format(FormatDefault, PrefixDefault, 3, 5)
```

(Being object-oriented.)

Theodus:
After some consideration I find @malthe's suggestion to be a better solution. I have modified the RFC to include a `Formattable` interface which will take the formatting concerns off of the current `Stringable` interface.

jemc:
One of the things I liked about the previous version of this RFC is that it reduced the amount of code and responsibility in `builtin`, and moved it to a new package that had only that responsibility.

[Reducing the number of type names in `builtin` is also a laudable goal, since it reduces the number of type names that are off limits for every developer](https://github.com/ponylang/ponyc/issues/286).

jemc:
Just to underscore, here's an important tidbit from the ticket I linked to.  These are the qualifications that @andymcn explained for whether a type should be in `builtin`:

> 1. Types the compiler needs to know exist, such as `None`.
> 2. Types with "magic" internal workings that must be supplied directly by the compiler, such as `U32`.
> 3. Any types needed by others in builtin.

I don't think things like `FormatDefault` and `PrefixDefault` fit these requirements, which was one of the original motivations discussed on the sync call where we talked about moving the formatting logic to a separate package.

Theodus:
@jemc Is there any way to move the logic of formatting into a separate package while also maintaining a more object-oriented API?

jemc:
No, I don't think there's a way to do that, though I don't think I understand the motivation for making these methods of the numeric types.
1. It's not any more succinct in terms of characters you have to type.
2. If anything qualifies as "data" in Pony, it's probably numbers - I don't see the problem with taking a functional approach to formatting numbers.

Theodus:
The recent changes have been reverted, I think that this RFC should move on to final comment period.

malthe:
I think that usability and coherency in how Pony presents its OOP model is more important than the length of the built-in class definitions.

I definitely agree that it's important to limit the number of type names exposed through `builtin`. However, I don't see how for example `String.split` would be placed better outside its class definition (outside builtin).

And so with that reasoning it seems arbitrary to want number formatting somewhere else. It fits naturally as a `Formattable` interface or trait.

An object that ends in "-er" such as a formatter can be seen as not an object at all and I think `Formatter` for example qualifies for this criticism. What is it? It's an object that formats. That makes no sense. What does make sense is that a number can be turned into a string.

jemc:
@malthe - I don't see any type named `Formatter` in the RFC, so I don't know what to make of your criticism about it.

> What does make sense is that a number can be turned into a string.

Note that numbers will still be `Stringable`, so you can still call `my_number.string()`.  You only need to use `Format` if you want to control the formatting of the number to be something other than the default (base 10, no padding).

> I don't see how for example `String.split` would be placed better outside its class definition (outside builtin).

No one's proposing removing `String.split`, and I don't see it as a good analogy.
1. `String.split` doesn't have any impact on the number of type names that `builtin` forces upon all pony program namespaces, as it doesn't have any accompanying primitives and interfaces.
2. `String.split` is part of the core use case for `String`s.  Strings are about manipulating byte sequences in terms of characters and codepoints.  Numbers are about computation and arithmetic.
3. `String.split` requires being able to access internal state of the `String` to be implemented in an efficient way.  Formatting numbers requires no knowledge of the numbers except for what is exposed in their public interface.

malthe:
Fair points. Perhaps a more object-oriented way then would be a `NumberFormat` class.
1. The operating system would probably provide one through the locale.
2. You could create your own, too:
   
   ``` pony
   class NumberFormat
     new create(fmt: FormatDefault = FormatDefault, prefix: PrefixDefault = PrefixDefault) =>
       ...   
   ```

That is, it should be a class, not a primitive. That at least feels more object-oriented to me and not like a primitive with a bag of functions on it – which I think is a code smell in OOP even if a number is admittedly, as close to pure data as you can get. The idea is that you're supposed to be able to talk about it as an object as opposed to a set of functions – and that there could be another object that satisfies the same interface.

In this case it could be the locale-based number format versus an application's "own" number formats.

Theodus:
This could be a good compromise, considering that it would also be easier to reuse an object to format with the same settings.

jemc:
I don't think I understand the advantage - if anything all I see is a disadvantage that you'd have to allocate a useless object just to call a function on data.

> a primitive with a bag of functions on it – which I think is a code smell in OOP

@malthe - The thing is, Pony is not an OOP-purist language, or an FP-purist language - it incorporates elements from both worlds where it makes sense, and I think this contributes significantly to its flexibility and power.  Like it or not, I think "primitive with a bag of functions on it" is an established and useful Pony idiom.

> it would also be easier to reuse an object to format with the same settings.

@Theodus - Actually, this would make it significantly _harder_ to reuse a format with the same settings.

If you want to reuse a particular format, all you have to do is create your own primitive function that wraps the `Format` primitive.  The pony compiler should be able to inline your function and you get a reusable format with no object allocation and no additional overhead.

``` pony
primitive _MyFormatForPrintingBytes
  fun apply(byte: U8): String iso^ =>
    Format(...)
```

I think you had in mind that if the reusable format were a class, you could create an instance of the class and keep it around in memory - this way you could just do one object allocation, and the cost of that allocation would be amortized over the course of all the subsequent uses.  However, this is not possible in cases where you do not have a viable place to store state.  For example, if I have a `primitive` that I want to use to convert a complex object to a string representation, and that includes formatting numbers in a particular way.  In a primitive, I have nowhere to store an object instance to amortize the allocation cost, so I end up allocating a useless object every time I need to call a function.

> You could create your own, too

@malthe - Exactly.  This is a case where you can build the OOP-friendly construct on top of the FP-friendly construct, but the opposite is not true.  So, we should not eliminate the FP-friendly construct.

SeanTAllen:
To reinforce one of @jemc's points. Primitives should be favored over classes where possible for performance reasons.

Theodus:
Considering the point of performance of a primitive over a class I believe this should go to final comment period.

malthe:
I would think that the compiler could optimize away the instantiation of the class in the case where it's "throw-away". That is, that no heap allocation would actually happen.

If that's not correct now then can we expect that to be correct some time in the near future?

I think that favoring primitives over classes at this stage is premature optimization. Because the FP argument isn't there without algebraic data types etc. – an FP language has tooling that supports this way of modelling logic. But this might not be the place to discuss that.

Praetonus:
> I would think that the compiler could optimize away the instantiation of the class in the case where it's "throw-away". That is, that no heap allocation would actually happen.

It can in most cases (it can turn heap allocations into stack allocations and then LLVM internal passes can remove unneeded stack allocations).

SeanTAllen:
@Theodus are you ready for this to go to a vote?

Theodus:
@SeanTAllen yes

SeanTAllen:
@Theodus excellent. can you squash the commits under the assumption that this might get merged next week and it will make it a little easier on everyone?

Theodus:
@SeanTAllen the commits have been squashed.

PR #25, opened 2016-08-03
Type case for match

Perelandric:
This RFC defines a syntax for `match` expressions that allows a developer to define a case as a "type case". A type case is one where a value comparison is not desired and a variable assignment is not needed. The logic used in determining a matching type is the same as that of the `as` operator.

[Rendered](https://github.com/Perelandric/rfcs/blob/type-case-for-match/text/0000-type-case-in-match.md)

### DIFF
--- a/text/0000-type-case-in-match.md
+++ b/text/0000-type-case-in-match.md
@@ -1,0 +1,134 @@
+- Feature Name: type_case_in_match_expression
+- Start Date: 2016-08-02
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This RFC defines a syntax for `match` expressions that allows a developer to define a case as a "type case". A type case is one where a value comparison is not desired and a variable assignment is not needed. The logic used in determining a matching type is the same as that of the `as` operator.
+
+# Motivation
+
+There are cases where the primary desire of a `match` case is only to disambiguate its type. In some cases, a developer may wish to work with the disambiguated value, while in other cases the value itself may be irrelevant.
+
+The current syntax that matches on value can easily be mistaken for a type comparison. With a `| MyType =>` case, it's easy to forget that if `MyType` is a `class`, then that class is expected to have a no-param constructor and an `eq()` method. A dedicated syntax for matching on type should help in this regard.
+
+So the type case syntax will allow a more clear and clean interface to what is ultimately the intent of the case. That improved syntax is the primary motivation of this RFC, and was raised as a concern in [issue #876](https://github.com/ponylang/ponyc/issues/876).
+
+It seems there may also be *potential* to simplify management of *reference capabilities* in some cases. *(See 'Unresolved questions' below.)*
+
+# Detailed design
+
+The proposed syntax is to simply allow a `match` case to be defined as the keyword `as` followed by the name of a type and an optional *reference capability* annotation. If no *rcap* annotation is provided, the default for that type is used. As expected, the *rcap* of the type must be a valid subtype of the original.
+
+If the type case matches the subject of the `match`, the branch for that match is then entered just like any other case, except that if the subject of the match was a field or variable, then that field or variable could be directly used as the matched type.
+
+As a simple example, if we have a field `n` in a `box` method that is a `(USize | String ref)`, we currently may do something like this:
+
+``` pony
+let m = match n
+| let us: USize => us
+| let str: String box =>
+  try str.usize()
+  else USize(0)
+  end
+end
+```
+
+This is simple enough, though it's less clean than it could be. Ultimately the variables are there only to facilitate working with a variable that we could potentially work with directly.
+
+Given this RFC, we could update the above code to this:
+
+``` pony
+let m = match n
+| as USize => n
+| as String box =>
+  try n.usize()
+  else USize(0)
+  end
+end
+```
+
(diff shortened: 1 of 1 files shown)


jemc:
I think one potential point of contention with this RFC is the behaviour of giving access to the type-refined reference under the same name within the branch of the `match` block.  Arguably, this could be considered a kind of shadowing, and Pony has taken decidedly anti-shadowing stances on other issues.

That said, I'm not personally opposed to this aspect, and I think it makes sense from a "proof" point of view - within the branch of the `match` block, you know that a reference is refined to the given type, so you should be able to execute code that acts on that knowledge.

However, a possibly sticky example comes to mind with `var` references where the value can be reassigned.  Take a moment to think through the following example, decide whether it should compile okay or not, and consider what the type of `x` should be considered to be by the compiler at any given point in time:

``` pony
var x: (U64 | String) = 0
match x | as U64 =>
  x = x + 1
  x = "foo"
end
```

Granted, this example is contrived, but I think it highlights some potential problems with considering the type of `x` to be refined within the block.  Or maybe it doesn't.  But it's definitely behaviour we should define in this RFC.

(Note that compiler currently doesn't let you do things like `match | var x': U64 => ...` - you can only bind on `let` within a `match` block branch).

Perelandric:
@jemc Thanks for your input. It is a departure from how Pony's type system works. I don't think it's a bad direction but I understand it wouldn't be without controversy.

WRT the `var` example, that's a good point and I'd say code should not compile because of the `"foo"` assignment since within that scope the `as` effectively makes it a `U64` in that scope. The increment of `x` as a `var` doesn't bother me so much since the compiler permits other outer scoped vars to be mutated, but I get your point.

I'll update the RFC later to reflect this in a little bit.

I'm curious, do you happen to know why `var` was dropped for match bindings? Only thing I could think of would be to perhaps accommodate case functions.

---

_Edit:_ While this may be a little bit of a stretch, it could be argued that there's _some_ precedence for refining the type of a variable by using the `as` operator.

``` pony
var x: (U64 | String) = "foo"

(x as String).lower()
```

Not exactly the same, but this is effectively refining `x` to be a String, albeit for only that moment. This could be seen as expanding on that concept.

jemc:
> I'm curious, do you happen to know why `var` was dropped for match bindings?

No, I don't know, unfortunately - I'd be interested to know that as well, and I think it might impact the discussion of this RFC.

Praetonus:
It looks like `var` for captures was dropped in [this commit](https://github.com/ponylang/ponyc/commit/323cb909c84f981b97363abd4974abbe96fe3ff8). Unfortunately, the reason isn't explained.

SeanTAllen:
@Perelandric this seems to have stalled. anything you need to get it moving again that I might be able to help facilitate?

Perelandric:
@SeanTAllen  I think it should remain stalled or be dropped. The main reasons are:
- It's non-essential
- It could probably be added later if desired
- It violates the principle of not having more than one way to do something

If someone else wishes to champion this, they certainly may, but I don't know that it's the best use of time at the moment.

malthe:
The motivational example in the proposal could be rewritten as:

``` pony
let m = match n
| let n': USize => n'
| let n': String box => try n'.usize() else USize(0) end
end
```

That is, without introducing new variable names but leveraging that Pony allows the use of prime in variable names.

I think this illustrates that the current syntax is probably adequate.

SeanTAllen:
n' is a new variable name

malthe:
Yes, `n'` is a new variable name, but it's trivial to come up with as opposed to the example's `us` – and it's easy to read as a sort of "local, subtyped capture".

SeanTAllen:
@Perelandric's point still applies. You are creating a variable you don't really want or need.

Perelandric:
@malthe Yes, this isn't solving something that can't currently be done; it's just providing a nicer syntax IMO but at expense of _"two ways to accomplish the same thing"_ as noted in the drawbacks.

The somewhat clumsiness of the current syntax is evidenced in the [issue](https://github.com/ponylang/ponyc/issues/876) that I linked to in the "motivation" section. Naturally a prime (or multiple if required) can be used for better<sup>?</sup> naming, but as @SeanTAllen mentioned, it's the superfluous name itself that drives this RFC.

So ultimately, you're right. There are other ways to do this, which is why I suggested above that I think this should be at the bottom of the list, if on the list at all.

Perelandric:
In the detailed design, I mentioned _"the rcap of the type must be a valid subtype of the original"_. Thinking about this now, that seems wrong unless the value would be implicitly consumed if, for example, we're going from `iso` to `ref` in that branch. I don't think that would be a good thing.

Unless someone has other thoughts on that, I'll remove that part.

jemc:
In general, I'm inclined to see this RFC as desirable, but the parts of it that relate to using the same name as the matched type inside the branch seem like they will be difficult to implement, due to how the compiler passes currently work.  However there may be some kind of neat trick that can be applied to make it more feasible.

I think I'd like to have someone (possibly me, if I eventually get the time) take a crack at implementing this and see how it would work (and how much of an ugly and complicated hack it would be in the compiler) before I would approve of this RFC.

SeanTAllen:
@jemc there's been no movement on this, what do we want to do?

jemc:
I'd still like to see this change happen, but I expect it won't be easy until we do a rewrite of the typechecker to use a kanren-style of constraints. Unless anybody has a clever idea about how to do this. I suspect it would be similar to how `iftype` works, but `iftype` is still a bit buggy.

SeanTAllen:
Closing as stale.

PR #26, opened 2016-08-11
Generic type inference

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/generic-inference/text/0000-generic-type-inference.md).

### DIFF
--- a/text/0000-generic-type-inference.md
+++ b/text/0000-generic-type-inference.md
@@ -1,0 +1,53 @@
+- Feature Name: generic-type-inference
+- Start Date: 2016-08-11
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add type parameter inference for generic functions based on the existing inference rules for local variables.
+
+# Motivation
+
+Programmers are lazy and typing type arguments is a lot of work. A smart compiler should infer type parameters when possible.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


jemc:
Should we also consider inferring a type's type parameter if it is included in the constructor?

For example:

``` pony
class Foo[A]
  new create(a: A) => // ...
```

``` pony
let x = Foo("hello")
// inferred as:
let x = Foo[String]("hello")
```

Praetonus:
Yes, I don't think it would cause any problem, even when the constructor itself is generic. I'll add it to the RFC.

Praetonus:
I think this should go into final comment period.

sylvanc:
I'm hugely in favour of this. It will also make parameterised operators (e.g. things like `add[A](a: A)`) work much more nicely.

jemc:
@Praetonus - can you send an email to the mailing list notifying of the final comment period?

cquinn:
This is very nice. 👍

Praetonus:
I think it's ready for vote.

cquinn:
👍

chalcolith:
:+1:

Theodus:
:+1:

sylvanc:
👍

PR #27, opened 2016-08-19
Make Map.upsert return the value for the given key

SeanTAllen:
A change to the API for Map.upsert

### DIFF
--- a/text/0000-upsert-returns-value.md
+++ b/text/0000-upsert-returns-value.md
@@ -1,0 +1,71 @@
+- Feature Name: upsert-returns-value
+- Start Date: YYYY-08-19
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+Currently, Map.upsert returns `this` to allow for chaining. This RFC would switch upsert to returning the new value stored for the given key.
+
+# Motivation
+
+Chaining is a nice convenience but in the end, serves no practical purpose beyond programmer ergonomics. Returning the new value from upsert saves another call to get that value (which in my experience using upsert, is a thing you commonly want to do). 
+
+# Detailed design
+
+Here is the new upsert method body:
+
+```pony
+    (let i, let found) = _search(key)
+
+    try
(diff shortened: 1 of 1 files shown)


sylvanc:
I like this, particularly if we adopt a new syntax for chaining.

jemc:
On a conceptual/consistency level, does the conflict with the current Pony idiom of returning the _old value_ when you do a replacement?

SeanTAllen:
@jemc yes but upsert is also a bit of a different case. in those other cases, you know what the new value is. in the upsert case, you don't. we could return a tuple of "old value, new value". So far in using this, I've only needed new value.

jemc:
Okay, sounds fine - I definitely have no problem with getting rid of the "return this for method chaining" when there is a useful value we can return instead.  As @sylvanc mentioned, we'll hopefully be adopting another syntax for chaining anyway.

This can probably go ahead and enter final comment period, right?

jemc:
Yeah, I suppose if you knew you needed the old value, there would be no reason to use the `upsert` method - you could just as easily get the current value and set the new one from it, and you already have access to the old value.

SeanTAllen:
@jemc so just a single return value or the tuple of old and new? its quite possible you wouldn't now either. so, i like the tuple idea.

jemc:
I wasn't necessarily arguing for the tuple idea.  What would you return for the old value if there was no value for that key prior to the `upsert`?

SeanTAllen:
Ugh. Good point. I'm inclined to leave as new value and it could be open to change later as it sees more usage.

jemc:
Sounds good.

SeanTAllen:
Alrighty. Welp, let's see if anyone else has comment over the next few days. I'll give this til the 31st when it can go up for a vote.

sylvanc:
👍

chalcolith:
:+1:

Theodus:
:+1:

PR #28, opened 2016-08-31
FFI taming

dckc:
Apply a uniform taming policy to FFI in the pony standard library and document it as a best practice for the community. In brief, the policy is to "dominate" all impure FFI calls with AmbientAuth.

### DIFF
--- a/text/0000-ffi-taming.md
+++ b/text/0000-ffi-taming.md
@@ -1,0 +1,36 @@
+- Feature Name: (fill me in with a unique ident, my_awesome_feature)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+One para explanation of the feature.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Detailed design
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with the language to understand, and for somebody familiar with the compiler to implement. This should get into specifics and corner-cases, and include examples of how the feature is used.
+
+# How We Teach This
+
+What names and terminology work best for these concepts and why? How is this idea best presented? As a continuation of existing Pony patterns, or as a wholly new one?
+
+Would the acceptance of this proposal mean the Pony guides must be re-organized or altered? Does it change how Pony is taught to new users at any level?
+
+How should this feature be introduced and taught to existing Pony users?
+
+# Drawbacks
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Can you walk me through the issues with something like Time.now() @dckc ?

dckc:
OK... I'll give some thought to how to unpack the argument.

For level-setting: it's clear that `Time.now()` is not a pure function, right?

dckc:
As I say, I intend to unpack the argument in a revision of the RFC, but meanwhile, in case you're in the mood, the most precise treatment I've seen for this stuff is:
- [Reasoning about Object Capabilities using Logical Relations and Effect Parametricity](http://www.eros-os.org/pipermail/cap-talk/2016-March/016762.html)

That's a March 24 cap-talk thread about a paper and presentation given by Dominique Devriese at the IEEE EuroS&P conference.

The paper explains:

> In the object capabilities model, effects can only be
> produced by sending messages to objects. These can be
> objects in the sense of object-oriented programming, but
> not necessarily so. Device objects are primitive objects that
> model resources in the outside world and produce effects
> in the outside world when receiving a message

A clock is such a device, and like I/O in haskell, reading is an effect just as much as writing.

SeanTAllen:
So you believe that anything that isn't a pure function needs to have a capability?

dckc:
That's a reasonable gloss of "effects can only be produced by sending messages to objects" so yes, not only do I believe it, but the authors of that paper and many others; as another example:

> All Authority Accessed Only by References. The authority an object has to affect or be affected the world outside of itself should be exactly represented by the references it holds. All primitives for interacting with the external world must be embodied by primitive objects, and one must obtain references to these primitive objects in order to exercise the associated primitive functions. Anything globally accessible must therefor be transitively immutable, otherwise it would constitute an unconditional source of authority not controlled by the reference-passing rules. -- [ode](http://www.erights.org/elib/capability/ode/ode-capabilities.html)

SeanTAllen:
I don't see any real world utility in requiring Time.now() to be an async message send. I do see serious performance implications. Also major usability issues.

dckc:
Async? Oh... I see how the "sending messages" terminology could be misleading. It's meant to be read in the smalltalk sense, where normal function calls are also message sends.

SeanTAllen:
@dckc ah, yes that's a big difference. i'm on the sync call now. i'll go back and re-read later with that new understanding.

SeanTAllen:
Closing as stale.

dckc:
What deadline did I miss? This is still relevant, as far as I can tell. As of Aug 2016, you were going to go back and read it again.

PR #30, opened 2016-09-01
Extend the itertools package

Theodus:
[Rendered](https://github.com/Theodus/rfcs/blob/extend-itertools/text/0000-extend-itertools.md)

### DIFF
--- a/text/0000-extend-itertools.md
+++ b/text/0000-extend-itertools.md
@@ -1,0 +1,49 @@
+- Feature Name: Extend tertools
+- Start Date: 2016-09-01
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+The itertools package of the standard library may be extended to provide useful classes and primitives for performing transformations on collections through the use of iterators. But before we implement the equivalent of [Rust's Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html) or [Elixir's Enum](http://elixir-lang.org/docs/stable/elixir/Enum.html), we should come to a consensus on whether to add these features to itertools, as an extension to the Seq interface, or as an extension to the Iterator interface.
+
+# Motivation
+
+The goal of this RFC is to implement various data transformations such as map, filter, reduce, etc., now and with more to come, without re-implementing for every collection.
+
+# Detailed design
+
+1. Add the following to the itertools package:
+  ```pony
+  class Enum[A, B: (Real[B] val & Number) = USize] is Iterator[(B, A)]
+    """
+    An iterator which yields the current iteration count as well as the next value
+    from the given iterator.
+    """
+  
+  primitive Fold[A, B]
+    """
+    Apply a function to every element, producing an accumulated value.
+    """
+    fun apply(iter: Iterator[A], f: {(B, A!): B^ ?} box, acc: B): B^ ?
+  ```
+
+2. Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).
+
+3. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Enum(Array.values())`)
+
+# How We Teach This
+
+The current documentation in the itertools package will serve as the template for documenting classes added to the package.
+
+# Drawbacks
+
+This will break existing code.
+
+# Alternatives
+
+Add similar functionality to the Seq or Iterator interface
+
+# Unresolved questions
+
+None
--- a/text/0000-extend-itertools.md
+++ b/text/0000-extend-itertools.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-The itertools package of the standard library may be extended to provide useful classes and primitives for performing transformations on collections through the use of iterators. But before we implement the equivalent of [Rust's Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html) or [Elixir's Enum](http://elixir-lang.org/docs/stable/elixir/Enum.html), we should come to a consensus on whether to add these features to itertools, as an extension to the Seq interface, or as an extension to the Iterator interface.
+The itertools package of the standard library may be extended to provide useful classes and primitives for performing transformations on collections through the use of iterators. The package may eventually be extended to the equivalent of [Rust's Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html) or [Elixir's Enum](http://elixir-lang.org/docs/stable/elixir/Enum.html).
 
 # Motivation
 
@@ -13,24 +13,28 @@ The goal of this RFC is to implement various data transformations such as map, f
 
 # Detailed design
 
-1. Add the following to the itertools package:
+1. Move the functionality of the primitives in itertools to a single wrapper class `Iter[A]`, the constructor of which will have the following signature:
+```pony
+  class Iter[A]
+    new create(iter: Iterator[A])
+```
+2. Add `enum` and `fold` methods to the class
   ```pony
-  class Enum[A, B: (Real[B] val & Number) = USize] is Iterator[(B, A)]
+  fun enum[B: (Real[B] val & Number) = USize]: Iter[A]^
     """
     An iterator which yields the current iteration count as well as the next value
-    from the given iterator.
+    from the iterator.
     """
   
-  primitive Fold[A, B]
+  fun fold[B](f: {(B, A!): B^ ?} box, acc: B): B^ ?
     """
     Apply a function to every element, producing an accumulated value.
     """
-    fun apply(iter: Iterator[A], f: {(B, A!): B^ ?} box, acc: B): B^ ?
   ```
 
 2. Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).
 
-3. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Enum(Array.values())`)
+3. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Iter[A](Array.values()).enum()`)
 
 # How We Teach This
 
@@ -42,7 +46,7 @@ This will break existing code.
 
 # Alternatives
 
-Add similar functionality to the Seq or Iterator interface
+Add similar functionality to the Seq or Iterator interface directly
 
 # Unresolved questions
 
--- a/text/0000-extend-itertools.md
+++ b/text/0000-extend-itertools.md
@@ -33,10 +33,6 @@ The goal of this RFC is to implement various data transformations such as map, f
     """
   ```
 
-3. Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).
-
-4. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Iter[A](Array.values()).enum()`)
-
 # How We Teach This
 
 The current documentation in the itertools package will serve as the template for documenting classes added to the package.
--- a/text/0000-extend-itertools.md
+++ b/text/0000-extend-itertools.md
@@ -13,11 +13,12 @@ The goal of this RFC is to implement various data transformations such as map, f
 
 # Detailed design
 
-1. Move the functionality of the primitives in itertools to a single wrapper class `Iter[A]`, the constructor of which will have the following signature:
-```pony
-  class Iter[A]
+1. Move the functionality of the primitives in itertools to a single wrapper class:
+  
+  ```pony
+  class Iter[A] is Iterator[A]
     new create(iter: Iterator[A])
-```
+  ```
 2. Add `enum` and `fold` methods to the class
   ```pony
   fun enum[B: (Real[B] val & Number) = USize]: Iter[A]^
@@ -32,9 +33,9 @@ The goal of this RFC is to implement various data transformations such as map, f
     """
   ```
 
-2. Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).
+3. Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).
 
-3. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Iter[A](Array.values()).enum()`)
+4. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Iter[A](Array.values()).enum()`)
 
 # How We Teach This
 


SeanTAllen:
@Theodus i'd suggest soliciting feedback as to whether this approach or adding to Seq is the better approach. I'm sure @aturley might have some thoughts.

I'd also be very interesting in hearing your thoughts on the pros and cons of each approach.
I think as you collect this feedback you should add it along with your own thoughts to the RFC as pros/cons for each approach.

aturley:
I think putting these in `itertools` makes more sense than adding them to `Seq`. I could see wanting to apply these functions to things that don't conform to the `Seq` interface, but that you could still get an iterator for (like generators).

malthe:
It would be useful to survey which functions would conform to the `Seq` interface and which wouldn't. I feel that there are better ergonomics in having them attached to the most minimal interface.

SeanTAllen:
@Theodus are you ready to move this to final comment?

Theodus:
@SeanTAllen yes

ryanai3:
I actually heavily disagree with implementing filter/map/fold style functions _only_ this way - I think it loses out in terms of ease of use and readability vs the .filter.map.fold style you see in ruby and scala. Without some nice haskell-esque syntax for composition, writing these and reading them gets nasty:
i.e.:
something easy to write and easy to read in "scala-style":

``` pony
   let words: Seq[String] = mySeq.map(this~gen_list(size))
  .flat_map(decoder~())
  .splitAt(encoder~is_delimiter())
  .map(this~section2string())
  .join(" ")
```

becomes something I really had to think hard about for a couple minutes to write in "python-style". It's also not immediately obvious that it does the same thing - you have to really stare at it for a bit (even with the formatting!)

``` pony
let words: Seq[String] = join(" ", 
  map(this~section2string(), 
    splitAt(encoder~is_delimiter(),
      flat_map(decoder~(), 
        map(this~gen_list(size), mySeq)
      )
    )
  )
)
```

Swapping the order of the function and the iterable arguments to the functions manages to make it more confusing: 

``` pony
let words: Seq[String] = join(
  map(
    splitAt(
      flat_map(
        map(mySeq, this~gen_list(size)
      ), decoder~()
    ), encoder~is_delimiter()
  ), this~section2string()),
 " ")
```

I actually really struggled to format the above in a way that made sense.
We could use variables for each step, but naming is hard, and many of the steps don't really make sense to split out into a new variable  - I just want to `return my_seq.map(fn).filter(pred)`, not 

``` pony
mapped_seq = map(my_seq, fn)
filtered_seq = filter(mapped_seq, fn)
return filtered_seq
```

When it comes to applying these sort of functional operators on collections, it's much more intuitive to say "I'm mapping, then filtering, then mapping, then flatmapping, then joining" as opposed to "I'm joining something which is the result of mapping over something which is the result of filtering over the result of mapping over my original collection".

In my mind, this is one of the biggest problems you see with a lot of pythonic apis (even numpy!). There's a reason everyone uses list comprehensions when they want to write this sort of code in python. For example, the biggest reason I'll use ruby over python for one-off scripts where it's easily expressed as a bunch of functional operations(map, filter, etc.) is the syntax and the format of ruby. It's all well and good to have these in the std lib, but the moment you _only_ have these versus the scala-style methods on collections, writing this kind of code becomes significantly more difficult and more time consuming and harder to parse.

With regards to generators - if we're adding them to the language as a language feature anyway, why can't we just internally make sure they obey some interface like scala's Traversable: `pony fun box foreach({(A)})`, and internally make them inherit our equivalent of a Traversable trait so I can do stuff like:

``` pony
let my_generator = for y in mySeq do
  if somecondition
    yield something
  else
    // doing a bunch of stuff here
    yield complex computation here
  end
end

return my_generator
  .map(decoder~())
  .collect(this~may_fail_mapping(key))
  .filter(this~is_defined())
```

I actually think the above is far superior to python's support for generators.

The only positive I see with this style over methods on collection objects is that you can call this on objects which the programmer is unable to make inherit from "Traversable". But, given that even inline objects can inherit from traits and interfaces, and that we can just make generators implement a foreach, and have them implicitly inherit from "Traversable", I don't see what justifies this over a method approach particularly given the loss in ease of use and readability. 
Is there a performance argument to be made?
Something about static typing?

If we want to have these so people can use them, sure, but if it comes at the expense of the method on collections approach, I don't think it's worth it.

aturley:
@ryanai3 I think the easy/hard distinction you're making is subjective. I'm used to using languages like Python and Clojure, so I have my own opinions on this matter. From a stylistic standpoint I can see going either way (either a "functional" approach like the current `itertools` package, or an "OO" approach like adding methods to an interface or trait).

If we add these as methods on a trait then the only consistent way to add new functionality is to modify the contract of the interface/trait. One could add their own extra functions (like in `itertools`), but then there's a difference in the calling conventions for methods that are part of the class versus these new functions. At that point it seems like you haven't avoided the concern you brought up, because they'll still come up for people adding these functions outside of the standard library.

With respect to generators, I wasn't suggesting that we introduce Python-style generators, I was trying to come up with an example of a situation where I would want to apply one of these functions to something that didn't implement the entire `Seq` contract.

SeanTAllen:
@aturley wouldn't the method chaining and pipelining RFC #4, address @ryanai3's concerns?

I'm very interested in that RFC in part because I think it adds a nice level of sugar to make things easier to read ala clojure's "threading" macro.

aturley:
It feels to me like it would solve that issue, but I'd like to hear from @ryanai3 about whether or not he agrees.

I thought about bringing it up but decided against it since the feature is speculative at this point. If the existence of chaining/pipelining is a key part of which way we go on this issue then I would be in favor of postponing a decision here until we resolve #4, or at least move it farther along.

jemc:
I agree with @SeanTAllen that some of the pipeline operators would be quite helpful for creating the UX that @ryanai3 wants.  But I also wanted to make a couple of other points:

If we wanted to "bake these in" as methods, I'd rather do it on a type like `Iterator` than one like `Seq`, as a `Seq.values` is only one kind of possible iterator.  That is, anywhere you have a `Seq`, you can turn it into an `Iterator`, but not vice versa.

Also, it seems like it would be trivial to add a "wrapper" object to the itertools package that wrapped a normal `Iterator`, and provided the various itertools operations as methods.  For example: `Itertools(my_seq.values()).map(...).zip(...).fold(...)`.  This should provide the desired convenience, while not limiting other uses of the `Map`, `Zip`, and `Fold` types, if this convenience is something we decide that we need.

jemc:
However, I'd also say that I kind of see this line of discussion as being outside the scope of this RFC, since the itertools package already exists and already uses types as operations.  If we wanted to pursue changing this paradigm, I feel like it should be discussed in another RFC.

ryanai3:
So I brought it up because the RFC itself explicitly mentions removal

> Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).
>  +
>  +3. Remove any iterators that include ordered indices, such as `ArrayKeys` (which may be replaced by `Range(0, Array.size())`) and `ArrayPairs` (which may be replaced by `Enum(Array.values())`)

I wouldn't be a favor of removing wholesale a way of doing the main thing you do with a standard library of collections based on a speculative feature that will take time to come.

There's many arguments that are not UI based - that's just the first one that jumped to mind:

@jemc, @aturley  - In a sense we already have both paradigms existing - Itertools is there, but so are all the methods on persistent/List and company. So I don't buy the argument of "Itertools is already here".

With regards to where we put the methods, we can have our cake and eat it too.
We have the magic of inheritance... 
 Scala doesn't have one "god-interface", rather, it's split out into a tree. All you have to do is implement `fun box foreach(fn: ref {(A)} | val {(A)})`, and you get all the rest of the behavior for free. Scala doesn't just have "Seq" and "Iterator", they have 



in ruby, enumerable does not promise "map_in_place", but "Array" does.

There's also a number of arguments that relate directly to pony's core philosophy:

lazy vs. strict (i.e. performance):

By doing this, you also remove from the programmer _any_ ability to decide if they want things to operate in a lazy or strict manner, unless they want to rewrite the whole collections library. In scala, and heck, even in ruby I can do so. I.e.:
`my_collection.view().map().filter().force()` has different performance trade-offs than:
`my_collection.map().filter()`, and there are cases when either is preferred. I can even tune it with:

``` pony
my_collection.chunk(buffer_size).view().map(
  lambda s: Seq[U64] => s.map(~).filter(~)
).flatten()
```

In practice, this can have _huge_ impacts on performance. By doing this through functions and only through functions that behave on an iterator, we remove the ability of programmers to make this distinction.
In scala, we can even blur the lazy/strict divide. There is a collection that looks like a buffered list i.e.:
[array, , , , ] -> [array, , , , ] -> [array, , , ]. map and filter and co. can be strict on each sub-array, but lazy on the whole list.

In fact, in the scala-approach, Iterator extends Traversable, so I can do my_iterator.whatever() if I want to...

Typing: 

There's another problem with the function approach, which is the type of the result. When I filter over a Set in scala, I get the values back in a Set(if they obey the constraint to be in a set), When I filter over a HashMap in Scala I get back a Hashmap. This behavior even extends to mapping, as long as the new types satisfy the necessary constraint. (If map on hashmap gives u back a hashable type, you'll get back a map, otherwise it'll be a list of tuples, and this is known at compile time)
What are we going to do, have a Filter Type for every subcollection? ArrayFilter, MapFilter, SetFilter?

Intermediate Computation:

With the itertools approach _everything is an iterator_. We're not allowing the programmer to make choices about what underlying collections are being used in the middle which can affect performance. In addition, if someone wants to reuse a part of a computation, they have to do the extra step of getting it from an iterator and building a collection. i.e.:

``` pony
   let intermediate = my_hashmap.map(~).filter(~).flatmap(~)
   ok_its_finally_a_hash = HashMap.from_iterator(k)
   what_i_need = map(keys, ok_its_finally_a_hash~())
   final = intermediate.map(~).filter(~).flatten(~).fold_r(~)
```

The above are 2 things that chaining/pipelining _does not fix_. 

Adding functions:

Unfortunately, pony doesn't act like haskell or clojure:
What does pony do if I define another `primitive Fold` with an apply method that applies only to Sets, since I want `my_set.map(lambda x: U64 => x * 2 end)` to return a set? 
I'm pretty sure the compiler yells at me. 
With the methods on objects approach, we can just do some clever messing around with the type system - scala allows this to happen with _very little_ added code.
What if the implementation of "map" for a certain collection can reap large performance gains by being done internally, because of the design of the datastructure? that is something we can do in the methods on objects approach but is something we _cannot_ do in the itertools approach. Isn't that something that we should do? Isn't that something that lines up with the pony philosophy?

@aturley I cannot recall a single time I've needed or wanted to add a method to scala's standard library (as in, it has never happened). I can tell you a number of times where I've gone "python why don't you have .each_cons like ruby? That would make what I'm about to do next 5 seconds of thinking and 9 characters vs 30 seconds of thinking and 4 lines", and ditto for all the other nice functions ruby provides on enumerable and array. To a certain extent I think that's a problem you rarely see if the standard library comes with batteries included. Even ruby, where you can monkey-patch the hell out of anything has all of these nice extra methods. 

I'd think it would be cool for pony to have some sort of compile-time adding of functions like ruby's monkey-patching, but I'm not holding my breath.

Pony != Python:

Pony is a statically typed language that values performance and correctness. Python is a dynamically typed language that ... is slow (but seriously, many decisions made in python really don't care about performance, and that's fine - that's their side of the tradeoff). I think we need to be mindful about why itertools is around in python, and why we see such a markedly different approach in scala.  I think there's a reason why the scala people, despite being functional fanatics in some ways went with the methods on objects route. In short, it gives the programmer massively more leeway when it comes to performance (because the itertools approach has none). I've actually almost never used any functional methods in python. I write it as a list/dict comprehension because it's easier and more clear.

I think just looking at the pony philosophy, doing this at the expense of allowing methods on objects loses out an almost all the fronts.

Philosophy?

> Correctness. Incorrectness is simply not allowed. It's pointless to try to get stuff done if you can't guarantee the result is correct.
> 
> Performance. Runtime speed is more important than everything except correctness. If performance must be sacrificed for correctness, try to come up with a new way to do things. The faster the program can get stuff done, the better. This is more important than anything except a correct result.
> 
> Simplicity. Simplicity can be sacrificed for performance. It is more important for the interface to be simple than the implementation. The faster the programmer can get stuff done, the better. It's ok to make things a bit harder on the programmer to improve performance, but it's more important to make things easier on the programmer than it is to make things easier on the language/runtime.
> 
> Consistency. Consistency can be sacrificed for simplicity or performance. Don't let excessive consistency get in the way of getting stuff done.
> 
> Completeness. It's nice to cover as many things as possible, but completeness can be sacrificed for anything else. It's better to get some stuff done now than wait until everything can get done later.
- Correctness: Given that we are a statically typed language, having more strictness than "everything is an iterator" is I think incredibly attractive. Traversable.filter() returns a traversable. Set.filter() overrides that to return a set - _why_ is doing `Set.from(filter(~~~, set.values()))` to get a filtered set more correct than `set.filter(~~~)` ?
- Performance: I wrote about this a bunch in the "lazy/strict" section. It's clear that the methods on objects approach wins on performance and flexibility, and the programmer's ability to make the performance tradeoffs that they want to make.
- Simplicity: With itertools, I'd argue we are pushed into the "god-object" box, because every itertools method has to work on an iterator. It totally makes sense that there are some things I can call on an IndexedSeq that it does not make sense to call on a Set (like .unique() for example). In addition, if an implementation of Map() on a collection that is implemented in a custom way for that collection can reap massive performance gains, the itertools approach does not allow us to do it.
- Consistency:  I think these are even
- Completeness: I think these are even

If someone wants to add itertools, fine by me. In fact, I'd be happy to see it, it would be pretty cool :). People who want to use it can use it. I would probably use it at times. But when the question is itertools and by extension _not_ methods on objects,  I think that just the performance loss on its own should make it a nonstarter.

ryanai3:
Sorry about the huge wall of text

SeanTAllen:
Touching on just a couple of points in this:

I think that the Scala collections library is not something I want to emulate in terms of implementation. Its a giant mass of inheritance that beyond a basic level is very difficult to work with. 

Ruby monkey patching is a nightmare. I don't want anything that looks like it especially if it looks like Scala's implicit.

malthe:
According to an article that admits that [scala ia hard](http://blog.goodstuff.im/yes-virginia-scala-is-hard), "dealing with collections is super easy".

What's in particular difficult to work with about the inheritance system?

SeanTAllen:
regarding performance concerns. I think that is very important. What I would like to see, if we are making performance claims would be actual numbers of possible implementations.

SeanTAllen:
@malthe i dont want to get into a conversation about issues with the scala collection system unless someone opens an RFC that says "lets make Pony's collections mirror scala collections". Ditto for Ruby monkey patching related stuff. At the point, I will go on record for saying, at present, I am not in favor of making Pony like either of those.

SeanTAllen:
@Theodus i can't at this time vote for this. I need to see a good discussion of the Pros and Cons of each of the approaches you lay out. This RFC tells me of 3 options and I have no basis for making a good considered decision.  At this time, the best I can say is... add functionality to itertools and leave the other functionality.

Do we really know the pros and cons of the 3 approaches you mention in the motivation?
If not, then removing any functionality seems unwise to me. Without those detailed pros and cons, I'd consider pursuing all 3 options with the understanding that ideally would like one way to do it (probably, again I just don't know).

Theodus:
After some thinking, I am strongly leaning toward adding these functions to the Iterator interface to maintain consistency with Pony's object orientated APIs such as 
`xs.values().map(...).fold(...)`. 
This may also be done with a wrapper object, as explained by @jemc such as 
`Iter[USize](xs.values()).map(...).fold(...)`.
I think that adding to the Iterator interface is more convenient for the user, but sadly I have not been able to look too far into possible implementations due to ponylang/ponyc#1216 and ponylang/ponyc#1193

SeanTAllen:
@Theodus do you to move this back out of final comment and do an update?

SeanTAllen:
@Theodus is this ready to get further review?

@ponylang/committer you might want to have a look again, I think this might be ready for another go round.

Theodus:
Yup, It's ready for further review.

jemc:
I think I'm starting to have a problem with the "removal" parts of this RFC.

---

> Remove the map, filter, and fold methods of the List and persistent/List collections with the intention of removing other methods in the future when they are implemented in itertools (such as flat_map).

I think @ryanai3 made a particularly good point when it comes to the possible performance benefits of some objects having access to the internal state of the object in their implementation:

> In addition, if an implementation of Map() on a collection that is implemented in a custom way for that collection can reap massive performance gains, the itertools approach does not allow us to do it.

---

The more I think about what the role of the `itertools` package is / should be, the more I think it is about bringing extra functionality for dealing with iterators, not necessarily _replacing_ that functionality for collections.  After all, note that an `Iterator` may not be associated with a collection - I can create iterators that operator over other kinds of data structures, or replay streamed data.

Theodus:
Agreed, there isn't much reason to remove existing functionality at this time.

Theodus:
Would anyone disagree with moving this RFC to the final comment period?

SeanTAllen:
Do it!

ryanai3:
Now that this no longer includes removal of the other methods, count me as a vote for!

PR #31, opened 2016-09-04
Allow calling tag functions without constructing an instance.

jemc:
[Rendered](https://github.com/jemc/pony-rfcs/blob/0a1a561f40ac996c005d7511259ba6731ed8d021/text/0000-fun-tag-no-construct.md).

### DIFF
--- a/text/0000-fun-tag-no-construct.md
+++ b/text/0000-fun-tag-no-construct.md
@@ -1,0 +1,48 @@
+- Feature Name: fun-tag-no-construct
+- Start Date: 2016-05-26
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This change will allow for calling `tag` functions on a type expression, without needing to call the constructor for that type.
+
+# Motivation
+
+Currently, defining functions on `primitive`s is the most common way of simulating the concept of a function with no receiver (even though the `primitive` is the receiver, as a singleton of sorts).
+
+The concept of a function with no receiver is closely related to the "static methods" found in some other object-oriented languages. These are commonly used for behaviour that is closely related to the instance methods, but that do not actually require being executed with a particular instance as the receiver.
+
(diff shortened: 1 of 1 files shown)


jemc:
[Rendered](https://github.com/jemc/pony-rfcs/blob/0a1a561f40ac996c005d7511259ba6731ed8d021/text/0000-fun-tag-no-construct.md).

hibnico:
I like the idea of adding support for explicit stateless function to classes. But as you well described, it has to avoid confusion and even conflicts with the implicit constructor call.
Often new Pony riders are confused by the `is` operator and don't see the implicit call. Since `A.foo()` syntaxly read as a call of the function `foo` on `A`, as an alternative I would be in favor of a more radical change to make the language a clear and lean as possible: remove the support implicit constructor call.

Praetonus:
I don't think trying to "simulate" static functions like this would be a good thing.

> Therefore, if the receiver capability of `foo` is `tag`, the instance created in the expression `A.foo()` will always be opaque, and any fields it may have will never be read by `foo` or any subsequent function.

This is not entirely true. If there is a finaliser for `A`, the fields of `foo` can be read there even if the only alias of the object was a `tag`. That means we must construct a valid object and we're back to the starting point. Even without a finaliser we'd still need at least a valid type descriptor since `tag`s can be used for pattern matching and virtual dispatch.

I'm not particularly opposed to the idea of static functions in Pony but if we were to implement them, I think it should be completely distinct from functions with receivers.

jemc:
We discussed this on the sync call - as @Praetonus indicated, there are lots of problems with using `tag` for this.  We also discussed some type-parameter alternatives for the particular recent use case that motivated this RFC.  It also became clear that even with static methods, I wouldn't be able to work the use case I wanted without being able to pass an uninstantiated class type as an expression value, which is not legal Pony and probably never will be.

@sylvanc also mentioned that he's sort of opposed to static methods in that it creates two types in the compiler (one for instance methods, one for static methods)

Maybe we can return to the idea of static methods in the future if we find use cases that can't be solved by other means, but for now, I'm going to close this RFC.

PR #32, opened 2016-09-07
Method must refer to trait or interface

malthe:
A public method on an actor or class must implement a declared interface or help complete a trait.

### DIFF
--- a/text/0000-method-must-refer-to-trait-or-interface.md
+++ b/text/0000-method-must-refer-to-trait-or-interface.md
@@ -5,15 +5,13 @@
 
 # Summary
 
-A public method on an actor or class must implement a declared trait or help complete an interface.
+Public methods on an actor or class must implement a declared trait or help complete an interface.
 
 # Motivation
 
 Since Pony is a statically-typed language, it makes sense to understands its OOP approach as contract-based. This allows the compiler to help the programmer write correct programs.
 
-In Pony, contracts are defined through traits and interfaces. The compiler validates objects against these, but it doesn't check if objects implements more than what's in the contract.
-
-This means that the contract can be changed with the implementation left out of sync.
+In Pony, contracts are defined through traits and interfaces. The compiler validates objects against these, but it doesn't check if objects implement more than what's in the contract (this applies for both traits and interfaces.)
 
 Java optionally provides the ``@Override`` annotation to require that a method overrides a method defined in a superclass and the same annotation can be used (rather confusingly) to require that a method implements a method in a declared trait (but again, the annotation is optional.)
 
@@ -23,7 +21,7 @@ This proposal is for a requirement that all methods defined on a class or actor 
 
 A method is allowed if it implements a method from a declared trait, or if it helps complete a trait – in any other case, it's a compile-time error.
 
-The requirement only affects public methods (those that do not begin with an underscore).
+Note that the requirement only affects public methods (those that do not begin with an underscore).
 
 # How We Teach This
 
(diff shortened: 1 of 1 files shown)


Perelandric:
Is the problem you're describing one where an interface/trait adds a new method signature to its requirements, and the object happened to already implement that exact method?

aturley:
I may be reading this incorrectly, but it sounds like you mean that _all_ public methods would need to implement a method from an interface or a trait. If this is the case, then wouldn't that break any Pony code that uses classes that do not subtype any traits or interfaces? That would break a whole lot of existing code.

malthe:
Yes, I mean your second interpretation, that all public methods must
implement an interface or (help complete) a trait.

It could be a flag "-Wpedantic", but one that Pony's standard library would
live up to and have enabled in its test suite.

Perelandric:
The motivation is unclear to me. I don't understand what the problem is with a type having more than is required by a trait or interface and how this improves that situation.

This seems like it will compel people to create unused interfaces simply to satisfy this requirement.

So if the trait is:

``` pony
trait Foo
  fun FooMethod()
```

And my class is:

``` pony
class MyClass is Foo
  fun FooMethod() => _do_something()
  fun OtherMethod() => _do_something_else()
```

I now need to add this unused interface, and keep it in sync with changes to `MyClass`:

``` pony
interface _MyClassHelper
  fun OtherMethod()
```

I'm unclear on how this enhances safety or strengthens the contract offered by the `Foo` trait, or how a situation is improved if this is a requirement of the standard library.

aturley:
@malthe based on what you've said, I think you should mention the changes that would need to be made to the standard library in your discussion of the implementation details. Also describe the `--Wpedantic` flag that would be required for backwards compatibility.

This also changes the flavor of the language a bit. One of the things interfaces are used for is structural subtyping (http://tutorial.ponylang.org/types/traits-and-interfaces.html), which means that a class can implement an interface without having to explicitly state that it does so, as long as it implements the methods described by the interface. What you're proposing would eliminate this. If you think structural subtyping needs to go away then you should discuss that in the proposal.

malthe:
I mistakenly swapped the meaning of traits and interfaces in my proposal. I'll update it shortly.

I don't think structural subtyping needs to go away, but I do think it's a much safer choice if this proposal is accepted because the compiler will complain if a class implements a method that's not part of a subtype.

malthe:
In the words of [Yegor Bugayenko](http://www.yegor256.com/2014/11/20/seven-virtues-of-good-object.html), "every public method in a good object should implement his counterpart from an interface. If your object has public methods that are not inherited from any interface, [it is] is badly designed."

"There are two practical reasons for this. First, an object working without a contract is impossible to mock in a unit test. Second, a contract-less object is impossible to extend via decoration."

I think those are good points, but perhaps @Perelandric has a valid concern that it may create a need for an unused and artificial interface which is obviously bad.

aturley:
I'm not of the opinion that the compiler should force certain behaviors for purely ideological reasons, so while I may agree with Bugayenko on some level I'm still going to reject that as a supporting argument for this RFC. Programmers should have some freedom to select different ways to do things as long as it does not violate the safety guarantees of the language. As a wise man once said:
 ♬How many ways to get what you want? ♬
 ♬I use the best ♬
 ♬I use the rest ♬

The problem failing to over-ride the intended method is real (though somewhat reduced from the Java world by the fact that Pony only provides a limited form of overloading by way of case methods). I'd be interested in finding a solution to that problem, but I think the impact of the solution should match the impact of the problem, which is relatively small.

SeanTAllen:
I'm quite against this change. I don't see any need for a actor or class to have to implement some set of interfaces. This seems like massive overkill to me.

malthe:
I'm going to close this one. Thanks for all the feedback. I think it's an interesting space to explore, but this idea doesn't seem like the right one.

PR #33, opened 2016-09-12
Export functions for calling class methods from C

aturley:
[Rendered](https://github.com/aturley/rfcs/blob/export/text/0000-export.md)


Praetonus:
I really like the general idea but there are a some incorrect things here.

> The compiler uses the the LLVM fast calling convention when it generates functions.

Fastcc is not used when building libraries, we use the C calling convention in those cases.

> Since reference capabilities are enforced a compile-time but not a run-time there would be no need to produce wrappers for multiple versions of `box` functions, nor would there be a need to include reference capability information as part of the function name as is done for the internal functions that the compiler generates.

While there are no capability checks at runtime, the generated code for different refcaps isn't always the same. We'll need the multiple versions of the functions.

Other than that, :+1:

aturley:
@Praetonus Are you saying that you think it's worth explicitly mentioning that in the case of generating libraries we don't use Fastcc? Since this RFC isn't really about generating libraries (in the `--library` sense) I didn't think it was worth mentioning that, but if you think it would help to avoid confusion then I can mention it.

The intention here is was to compile a Pony program that makes C FFI calls, and from those FFI calls the programmer can call methods on Pony objects that have been passed into it. That said, I suppose this feature would also be useful in the case of compiling libraries.

Can you help me understand what the differences are between the generated code for different refcaps? My understanding was that it was the same, but I probably misunderstood. I had hoped to avoid forcing exposing those details to the C programmer, but it sounds like you're saying that won't be possible.

Praetonus:
> @Praetonus Are you saying that you think it's worth explicitly mentioning that in the case of generating libraries we don't use Fastcc? Since this RFC isn't really about generating libraries (in the --library sense) I didn't think it was worth mentioning that, but if you think it would help to avoid confusion then I can mention it.
> 
> The intention here is was to compile a Pony program that makes C FFI calls, and from those FFI calls the programmer can call methods on Pony objects that have been passed into it. That said, I suppose this feature would also be useful in the case of compiling libraries.

After rereading the RFC, that point looks detailed enough. I don't know why I was confused the first time.

> Can you help me understand what the differences are between the generated code for different refcaps? My understanding was that it was the same, but I probably misunderstood. I had hoped to avoid forcing exposing those details to the C programmer, but it sounds like you're saying that won't be possible.

The problem it solves is explained by @sylvanc [here](https://github.com/ponylang/ponyc/issues/691). I don't think the additional cognitive load would be a problem, it's simply a matter of calling `Type_cap_function` on an object with a refcap of `cap`. It requires knowing the refcap of the Pony object passed to C, but I think any sane interface between Pony and another language should make that information available.

SeanTAllen:
@aturley are there any updates to this or is it ready to go to final comment?

aturley:
@SeanTAllen I'm ok with moving to final comment as long as we're ok with punting on the unresolved question of how to handle exporting concrete implementations of generic functions.

There's actually a workaround where you can where you can create a primitive with a non-generic function that calls the generic function for you. We can use that until we decide when and how to cover generic functions.

SeanTAllen:
@Praetonus what are your feelings on @aturley's suggestion?

Praetonus:
I like the proposed syntax for exporting generic functions. The workaround isn't really convenient.

text/0000-export.md
```diff
@@ -0,0 +1,111 @@
...
+
+When compiled with the `--export` flag, a .h file would be produced:
+
+```
+// example.h
+
+extern void *Foo_Alloc();
+extern void *Foo_create(long);
+extern long Foo_bar();
+extern long Foo_baz();
```
jemc: Aren't these function signatures missing their `this` argument?

aturley: Yes, they are. I'll fix that.

aturley:
@Praetonus I like the proposed generic function syntax too. But I think adding support for that syntax will be more complex than support for the non-generic function version of `export`. Doing this in two phases would give us a chance to get most of the functionality out there for people to use, and then we could add the remaining functionality (generic functions) later after giving it more thought.

Praetonus:
That makes sense. Let's go ahead then.

SeanTAllen:
@aturley ready for Final comment period once that fix is in?

aturley:
I'd like to hold off until we discuss the questions from https://github.com/ponylang/ponyc/issues/416#issuecomment-250013107 a little more and incorporate that feedback into the RFC.

text/0000-export.md
```diff
@@ -0,0 +1,111 @@
...
+    let foo = Foo(32)
+    @my_fun(foo)
+```
+
+When compiled with the `--export` flag, a .h file would be produced:
+
+```
+// example.h
+
+extern void *Foo_Alloc();
```
jemc: `Alloc` should take a `pony_ctx_t*` argument, right?

aturley: Yes, I believe so. Fixed.

aturley:
From Sylvan's comments in https://github.com/ponylang/ponyc/issues/416#issuecomment-250285276 :
1. Export should work with actors as well. That allows it to work with pony_become.
2. Export not only works with library generation, it becomes library generation. That is, if you have some packages, creating a library means creating another package that exports your desired API.

aturley:
I'd like to wait on adding actor support and getting rid of "library" until after "export" is implemented so that we can see if there are any potential issues. If "export" can converge on "library"'s functionality and overtake it that's fine, but I'd like to get the feature working with classes first.

@sylvanc do you have any objections to doing this in a piecemeal fashion? Or do you think it would be better to do the whole thing at once?

SeanTAllen:
@aturley do you want to move this to final comment period?

I think moving in a piecemeal fashion is fine personally.

aturley:
@SeanTAllen yeah, I'm ready to move this to final comment period now.

SeanTAllen:
@aturley please send an email to pony+dev announcing that this has gone into final comment period.

aturley:
Done.

SeanTAllen:
@aturley is this ready for a vote tomorrow?

aturley:
@SeanTAllen yes, it is.

PR #34, opened 2016-09-19
Buffered read_until method.

lisael:


jemc:
> The `greedy` parameter tells if the separator byte is included in
> the result (and incidently if it's consumed). This adds versatility for
> corner cases without impacting the main use case.

Can you explain this a bit more thoroughly, so that it's clear what happens in each case?

jemc:
I'm not totally sure, but I'm wondering if the `greedy=false` case is the most optimal UX.  That is, I'm inclined to say we'd want to have the separator byte be consumed as well.

I think it's pretty common to not want the separator included in your return value, but it seems to me that having it remain in the buffer just leaves it "in the way" before you can move on to the next expected entity in the protocol.

In all cases when calling this method, you know exactly what the separator byte is, since it can only have one value (the value you passed in), so I also don't see when it would be useful to have the separator byte still be in the buffer - you already know what that byte is at this point.

Also, the `line` method's documentation has the behaviour of not including the newline in the result, but also not leaving it in the buffer.  So I think that having `read_until` with `greedy=false` behave in a similar way to be consistent with `line` would be best for this reason as well.

However, if we're changing the behaviour of this parameter, maybe `greedy` is no longer the best name - not sure.

SeanTAllen:
I agree. I think the normal case is to ditch a separator.

SeanTAllen:
@lisael can you add an example or two of in the wild protocols that meet the description you give in the motivation? Ideally with links.

SeanTAllen:
meta comment but, no need to hard wrap the lines in the RFC @lisael. Its easier if they are soft wrapped.

SeanTAllen:
My overall impression:

I'm in favor of this. I think it adds needed functionality. To me the only questions are naming and a bit more info in the RFC.

In particular, examples of such real world protocols helps bolster the case and if we have a few, we can verify that this proposal works for them all.

Also, naming of "greedy" is perhaps not so good. I think that by default, we should emulate the behavior of the existing `line` method which should be reimplemented with this change (inclusion of that reimplementation should be mentioned in the RFC).

Looking at `line`, I am struck that perhaps just scanning by a single byte is suboptimal.
I might want to look for '\r\n'. Ideally I think as I said above, we should be able to implement `line` with `read_until` or whatever we call it.

jemc:
> Looking at `line`, I am struck that perhaps just scanning by a single byte is suboptimal.
> I might want to look for '\r\n'. Ideally I think as I said above, we should be able to implement `line` with `read_until` or whatever we call it.

I had the same thought at first, but I'd be interested to know the performance difference between scanning for a single byte and scanning for a string of arbitrary (but usually quite short) length.  That is, I'd like to confirm that scanning for a string containing a single byte is not significantly slower than scanning for a single byte as a U8.

SeanTAllen:
@jemc the performance is an interesting issue. good point.

one of the reasons I'd like to see a sampling of the real world usages for this it to understand if single byte is enough. If it doesn't meet the real world use cases (like Windows line separator) then its problematic to only support a subset. We still need to support the multibyte use case. How is an interesting question.

lisael:
I had the same UX question, in my first version, the parameter was `eager: Bool = false`. It seemed to me like a double-negation so I edited.

Concerning `line` and consistency you clearly have a point. Although, because of `line` behaviour that just swallows 1 **or** 2 bytes –you'll never know–, `read_until` can't be consistent if it doesn't just swallows the separator too, in which case we don't need the parameter at all.

What I understand is that `line` has to have this behaviour to permit a consistent code across platforms, abstracting the line separator.

OTOH, it seems odd to me to change the data that the user provided, but maybe you're both right, it's simpler to ditch the separator. (I have  YAGNI tendences :) ).

@SeanTAllen I needed this to parse PG protocol messages https://www.postgresql.org/docs/current/static/protocol-message-formats.html . For some of them like ParameterStatus, read_until(0) would help.

I update the proposal with this example and soft wrapping :). I'm not sure about the UX for now.

SeanTAllen:
@ponylang/committer anyone else know of protocols we can add to the RFC and should look at to make sure it will work for those? I don't want to hang this up for a long time looking for them as I think this can always be improved later but if folks have some they can contribute now, that would be good.

SeanTAllen:
@lisael @jemc is `line` a special case with being 1 or 2 bytes? or are there other platform specific oddities? If its only `line` then having line as a special case makes sense to me and then leave it as implemented rather than implementing on top of `read_until`

jemc:
@SeanTAllen - I don't think it's platform-specific, in that it doesn't work differently on different platforms.  On both unix systems and windows it will treat `\n` and `\r\n` as a valid line separator.  [The Redis protocol](http://redis.io/topics/protocol) is an example of a protocol that uses `\r\n` as the line separator, regardless of whether it is running on a unix system or a windows system. 

Since `line` accepts either kind of separator, I think it is kind of a special case where it probably doesn't make sense to make `read_until` flexible to emulate it.

SeanTAllen:
So in order to support Redis and similar protocols, we need to support multibyte separators in some fashion, yes?

lisael:
I was reading the redis protocol and was just thinking that it's easy to parse with `line` only, but it's only by luck. They could have chosen **any** separator string of any length.

SeanTAllen:
I love the "by luck"

jemc:
> but it's only by luck

Well, sort of - I would say it's "by convention" - I think it's pretty common for plaintext-oriented protocols to use `\n` or `\r\n` as the separator.  However, I don't have any other examples off the top of my head :stuck_out_tongue_winking_eye:.

Does anyone have a counter example of a protocol that uses a multi-byte separator other than `\r\n`?

jemc:
Actually, I do have another example off the top of my head - [telnet](http://www.freesoft.org/CIE/RFC/1123/31.htm)!

SeanTAllen:
I can't ever unknow what I just read.

Unrelated, but I am quite happy that the RFC process is sussing out all these important details.

lisael:
Telnet is the very reason why most protocol use CRLF and plain text for integers, BTW. I'm reading dozens of them and I can't find another separator. ~~We must look in non-text, telnet-unfriendly protocols like postgres is. Maybe http2.~~ EDIT: Bytes-oriented protocols probably use single-byte separators or null-terminated strings.

lisael:
So here's the plan: if we don't find a protocol that use two bytes separators within a couple of days, we go with `read_until(U8)` an unmodified `line()`. Right?

For the eager/greedy parameter I can't find a use case. I was very conservative with the data, but it makes the API and the implementation more complex, and we'd better just ditch the separator.

In the worst case, the user has to manually append the byte, but it should be very rare.

What's your opinion ?

jemc:
I think ditching the `greedy` parameter (and always ditching the separator) sounds like a good plan - it's consistent with how `line` works, and if someone eventually does find a use case for `greedy=false` they can do another RFC to motivate that change (or comment on this RFC before the final comment period is up).

SeanTAllen:
agreed.

SeanTAllen:
@lisael can you update the RFC with the info about various protocols, line etc that has gone on in the discussion? I think after that is in place, this is ready to move to the final comment period.

lisael:
I've updated the RFC.

I left the question about the name of the method, though.

I'm not sure about the English syntax. Please add reviews if something is not clear.

SeanTAllen:
@lisael are you ready for a vote on this at the sync on wednesday?

lisael:
Yes, ready. There's still the unresolved question of the method name. If it's voted as-is, we will go with `read_until()`.

SeanTAllen:
@lisael this was accepted as a RFC! thanks. I'll be adding an issue for this on the ponyc repo in the next day.

PR #35, opened 2016-09-22
C-string vs bytes (or C-pointer)

malthe:
With the introduction of zero-copy trimming methods, some Pony strings are not null-terminated while some are. This is a proposal to ensure that `cstring()` always returns a pointer to null-terminated data, while at the same time having a new method `bytes()` return a pointer to the already allocated data (zero-copy) with no guarantee of null-termination.

### DIFF
--- a/text/0000-strings-cstrings-bytes.md
+++ b/text/0000-strings-cstrings-bytes.md
@@ -17,6 +17,10 @@ The existing code to add null-termination to the allocated memory is kept since 
 
 However, the current `null_terminated()` will be removed (it's now obsolete).
 
+Adjustments to `from_cstring()` will be made such that it will no longer have a length argument. Instead, a new constructor `from_bytes()` is added.
+
+Also, an array will no longer provide a `cstring()` method, but just `bytes()`.
+
 # How We Teach This
 
 The description of `cstring()` and in particular how it's used to interface with C needs to be updated.
--- a/text/0000-strings-cstrings-bytes.md
+++ b/text/0000-strings-cstrings-bytes.md
@@ -1,0 +1,34 @@
+- Feature Name: C-string vs bytes
+- Start Date: 2016-09-22
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+With the introduction of zero-copy trimming methods, some Pony strings are not null-terminated while some are. This is a proposal to ensure that `cstring()` always returns a pointer to null-terminated data, while at the same time having a new method `bytes()` return a pointer to the already allocated data (zero-copy) with no guarantee of null-termination.
+
+# Motivation
+
+The definition of a C-string is a null-terminated string. Hence, it's a reasonable guarantee for the `cstring()` method to give. Meanwhile, sometimes we're just interested in the raw string data (as a pointer) and for this use-case we'll introduce the `bytes()` method.
+
+# Detailed design
+
+The existing code to add null-termination to the allocated memory is kept since it's a small cost compared to the advantage that most of the time, `cstring()` will be zero-copy.
+
+However, the current `null_terminated()` will be removed (it's now obsolete).
+
+# How We Teach This
+
+The description of `cstring()` and in particular how it's used to interface with C needs to be updated.
+
+# Drawbacks
+
+None.
+
+# Alternatives
+
+None.
+
+# Unresolved questions
+
+None.
--- a/text/0000-strings-cstrings-bytes.md
+++ b/text/0000-strings-cstrings-bytes.md
@@ -31,7 +31,7 @@ None.
 
 # Alternatives
 
-None.
+The name `cpointer()` is suggested instead of `bytes()`.
 
 # Unresolved questions
 
--- a/text/0000-strings-cstrings-bytes.md
+++ b/text/0000-strings-cstrings-bytes.md
@@ -23,11 +23,19 @@ Also, an array will no longer provide a `cstring()` method, but just `bytes()`.
 
 # How We Teach This
 
-The description of `cstring()` and in particular how it's used to interface with C needs to be updated.
+There is a general lack of C FFI documentation for Pony and the implementation of this RFC should contribute here.
+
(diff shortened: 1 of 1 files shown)


jemc:
I think this could be a good idea - after all, there is no purpose I can think of for caring that a string is `null_terminated` except for sending it to FFI, so it could make sense to "fold" it into the operation for getting the FFI pointer.

However, I'm not necessarily wild about the `bytes` name - I wonder if something like `cpointer` would be better?

Also, if we do this, I think we'll want to see some more extensive changes than those currently enumerated in the RFC.  For example:
- `Array.cstring` should probably be renamed to use the name for the new non-null-terminated `String` method (`Array.cpointer`?), since `Array`s have no concept of null termination.
- Things like `String.from_cstring` will need consideration - they should probably renamed to use the new non-null-terminated name.

---

One possible option for the "Alternatives" section could be "add an optional boolean parameter to `String.cstring`"

jemc:
Also, I'll add that I've always been a tad annoyed that `cstring` is the name of the method for getting the underlying `Pointer` - it makes some sense for `String` (when null terminated), but not much sense at all for `Array`.  So, introducing a different name for the general case (like `cpointer`) would be a big plus for me.

malthe:
It makes sense perhaps to have the following:

``` pony
new from_bytes(str: Pointer[U8], len: USize = 0, alloc: USize = 0) => ...
new from_cstring(str: Pointer[U8], alloc: USize = 0) => ...
```

(Or `cpointer()`. I actually considered that, too. I will add it to the RFC as an alternative.)

SeanTAllen:
This breaks existing code and should be detailed as such under Drawbacks. Any RFC that breaks existing code has drawbacks and the possible level of that breakage in the standard library as well as for 3rd parties needs to be part of the RFC.

SeanTAllen:
Under "How we teach this". I think there's a general lack of C FFI documentation and starting some C FFI code and including the differences between these methods is something that should be considered. 

The "How we teach this" at the moment is to me, very vague and I'm not sure what it is proposing.
At minimum, I think we should have something in the class docstring detailing bytes and cstring and the differences between them.

SeanTAllen:
If this was updated to change cstring to cpointer on both String and elsewhere that cstring exists o classes, as well as introducing bytes (and adding it to any classes that might be relevant) I would be in favor of this along with the additions above.

It seems @jemc suggested cpointer as both a replacement name for cstring and for bytes.

I like cstring => cpointer is a big win. I think bytes is a decent name but could be better I suppose ¯_(ツ)_/¯.

I'm in favor of the idea behind this RFC I just think it needs to be cleaned up some.

SeanTAllen:
Can you add the new signatures for from_bytes and from_cpointer to the RFC? The lack of specificity there could lead to an incorrect implementation.

jemc:
To be clear, I was suggesting to keep `cstring` for the null-terminated version (since "C strings" are null terminated), and use `cpointer` instead of `bytes` (as a "C pointer" conveys a similar idea but does not imply null termination).

SeanTAllen:
I'm not a fan of `cstring`. It makes sense for `String` but not for `Array`.

I prefer `cpointer` and `bytes`.

Shall we rock, paper, scissors for this?

malthe:
@SeanTAllen note that for `Array` we would have either `bytes()` or `cpointer()` and not `cstring()`.

That is, `cstring()` makes sense only on `String` and going forward, it'll be removed from `Array`.

SeanTAllen:
@malthe @jemc I think this is ready to go into final comment period which should have it ready for a vote the second week of October, agreed?

malthe:
Yup

SeanTAllen:
@malthe i'm away over the next week. I'm willing to accept whatever names you and @jemc agree upon.

jemc:
One reason I'm not a big fan of the `bytes` name is that I think it sounds more like "an iterator over the bytes in a string", as it seems to match the names of other stdlib iterators (`keys`, `values`, `pairs`).  I know that `String.values` currently iterates over the bytes in a `String`, but in a rethinking of the `String` API we might decide that it makes sense to have `codepoints` and `bytes` iterators, with one of them possibly being aliased as the default `values`.  It's also possible that `codepoints` and `bytes` could be methods that return different "views" of the string, although the notion of what a string "view" is in Pony would need to be fleshed out.

At any rate, since bytes are a native concept in Pony (in fact, `ByteSeq` used to be called `Bytes`), I don't feel like the name `bytes` makes it sufficiently clear that this is a method that is only useful in the context of the C FFI.  Something like `cpointer` accomplishes that, though I would be open to other names that are able to similarly emphasize the _C_ or _Pointer_ or _FFI_ concepts that we want to evoke with this method.

Theodus:
I prefer the name `cpointer` to `bytes` because it is more clear about the intended purpose of the returned `Pointer[U8]`, where `bytes` may imply an array of U8.

malthe:
Great points. I'll update the RFC asap.

SeanTAllen:
@malthe is this ready for a vote?

malthe:
Yes.

cquinn:
I know it is a little late, but how about the name `cbytes()` instead of `cpointer()`? I know `bytes` alone fails to convey the low-level pointer aspect, but I think `cpointer()` misses conveying the type.
I am 👍  on this either way.

jemc:
@cquinn - Note that we're talking about the name of the method both for `String` and `Array`.  I think that `cbytes` makes sense for an `Array[U8]`, but does it also make sense for something like an `Array[F64]` or `Array[MyStruct]`?  I'm not sure, but I'm inclined to say it's not quite intuitive or clear (though it's still better than the current status quo of `Array.cstring` for that same purpose!)

cquinn:
Oh, I see. So then `Array[F64].cpointer()` would return a pointer to F64. I was thinking that these two functions were always byte-oriented. If so, then I agree, `cpointer()` is more accurate.

PR #36, opened 2016-09-26
"tcp must be connected"

SeanTAllen:
# Summary

Verify TCP Connections are open before allowing data to be written.
# Motivation

Prevent subtle bugs and potential runaway memory usage that can be introduced by being unfamiliar with the internals of the `TCPConnection` class.

### DIFF
--- a/text/0000-tcp-must-be-connected.md
+++ b/text/0000-tcp-must-be-connected.md
@@ -1,0 +1,46 @@
+- Feature Name: tcp must be connected
+- Start Date: 2016-09-25
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+Verify TCP Connections are open before allowing data to be written.
+
+# Motivation
+
+Prevent subtle bugs and potential runaway memory usage that can be introduced by being unfamiliar with the internals of the `TCPConnection` class.
+
+# Detailed design
+
+Currently in `TCPConnection` 3 write methods check before that the connection isn't closed before trying to write. The 3 methods are `write`, `writev` and `write_final`. This prevents TCPConnection from buffering data if other actors send it write messages after a connnection is closed. 
+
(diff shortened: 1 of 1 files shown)


malthe:
Couldn't you argue that the `write*` methods should be allowed to throw? – since internally, the `os_send` call does. The current consequence of a failed send is that the connection object is closed via `_hard_close()`, but a caller would have to remember to check for that after every write.

I think instead of silently discarding data, it would be better to throw an error.

SeanTAllen:
Write calls are async. They are behaviors. So throwing an error isn't possible.

SeanTAllen:
As a user, you get informed via the not connected callback when a connection fails to establish. The issue comes if you ignore that.

malthe:
On the other hand, the write calls could be synchronous (since the actual I/O is non-blocking). If you need a specific interleaving of calls, then it's always possible to use behaviors at a higher layer.

jemc:
Does `TCPConnection` try to reconnect if the connection fails?

SeanTAllen:
No. That would be another connection. You are notified of the failure and then spin up a new TCPConnection.

SeanTAllen:
Any additional comments from folks? If not, I'll put it into final comment period as that tends to heat up the commenting.

SeanTAllen:
While I don't think it impacts on this RFC, I've been thinking about why the fire and forget to `write` on `TCPConnection` is as it is, how that could be addressed and why it highlights a problem that exists with how messaging is done in Pony and composition with classes and actors. @jemc I think that dovetails into conversations around embedding actors, asynclambda which are in my mind part of a larger conversation around pain points in points composition and messaging models. See my comment on #38 about having a higher level conversation on that subject.

As I said at the beginning of this comment, I don't think any of the above impacts on this RFC as this RFC provides value and fixes a problem in the existing implementation that will be with us for a while.

PR #37, opened 2016-09-30
TCP backpressure hooks

SeanTAllen:


SeanTAllen:
I'm leaving for vacation on October 1 and won't be back til October 9. Any feedback during that period of time will be responded to/incorporated after my return.

SeanTAllen:
I think that for the magic Windows values, it would make sense to have them configurable by the programmer. Via constructor seems like a reasonable option as long as it's wrapped in an ifdef as exposing the values to programmers on platforms where they have no impact doesn't strike me as a good API.

text/0000-tcp-backpressure.md
```diff
@@ -0,0 +1,135 @@
...
+
+    _notify.throttled(this, true)
+
+  fun ref _release_backpressure() =>
+    _notify.throttled(this, false)
+```
+
+For non Windows platforms:
+
+`_apply_backpressure` is called `write_final`
```
jemc: This confused me a bit - based on the context from the next few examples I think you mean "is called in `write_final`" rather than "is called `write_final`"

jemc:
In the ticket where we discussed this concept some months ago (https://github.com/ponylang/ponyc/issues/806), @sylvanc mentioned (https://github.com/ponylang/ponyc/issues/806#issuecomment-218570658) that we discussed on the sync call supporting also the "other end" of backpressure, with a behaviour that tells the socket to stop listening until we tell it otherwise.

For example, if an "outbound" data flow gets `throttled`, but it's fed directly from an "inbound" data flow, this would let you pause listening for a bit to until the "outbound" flow is no longer `throttled` - this would help you avoid queueing (or queue growth) in the middle.

When I was working on that ticket, before the RFC process, and had marked it as "in progress" I was calling this `mute`, though there may be a better name available.  The only thing that kept me from moving to the "finished" state was trying to write decent tests for these two ends of backpressure.

@SeanTAllen - do you think this additional feature belongs in this RFC, or should be added later in another?

SeanTAllen:
I have that as part of my credit flow work but that requires additional support. I can add mute to this. It makes sense to add here. I have it implemented already, I just considered it part of another RFC but I can easily see it as part of this.

text/0000-tcp-backpressure.md
SeanTAllen: i updated.

SeanTAllen:
@jemc i'm adding `mute`. its different than what I have in my credit flow stuff but would actually work better than what I have. its a performance improvement (slightly). and more flexible. so, I call that a win.

SeanTAllen:
@jemc extensive changes done that were inspired by the addition of "mute".

PR #38, opened 2016-09-30
Add RFC for embedded actors.

jemc:
[Rendered](https://github.com/jemc/pony-rfcs/blob/embed-actor/text/0000-embed-actor.md).

### DIFF
--- a/text/0000-embed-actor.md
+++ b/text/0000-embed-actor.md
@@ -52,9 +52,7 @@ None.
 
 # Alternatives
 
-Other designs considered:
-
-* Consider breaking backwards compatibility in favor of better consistency - apply the don't-call-`create` rule even in cases where `create` has no required parameters (and thus could easily be called). In short, this would break cases where the `create` method had side effects through FFI (or a library that uses FFI).
+None.
 
 # Unresolved questions
 
--- a/text/0000-embed-actor.md
+++ b/text/0000-embed-actor.md
@@ -48,7 +48,7 @@ That is to say, I propose we allow embedding actors in the same memory space as 
 
 # Drawbacks
 
-None.
+* This implementation would add some significant complication to the current compiler, and possibly also the runtime. This complication would be carried as additional maintenance burden. It could also be a source of bugs, just as we've seen and worked through some subtle bugs related to the introduction of embedded classes.
 
 # Alternatives
 
--- a/text/0000-embed-actor.md
+++ b/text/0000-embed-actor.md
@@ -1,0 +1,62 @@
+- Feature Name: embed-actor
+- Start Date: 2016-09-29
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This change will allow for embedding actors within other actors, as a method of extending, augmenting, or otherwise wrapping the API of another actor through composition.
+
+# Motivation
+
+To quote from the tutorial, Pony does not support inheritance, and promotes composition as an alternative paradigm:
+
+> In some object-oriented languages, a type can inherit from another type, like how in Java something can extend something else. Pony doesn't do that. Instead, Pony prefers composition to inheritance. In other words, instead of getting code reuse by saying something is something else, you get it by saying something has something else.
+
+For classes where we truly want to wrap another class, we can even use the `embed` keyword to remove extra overhead associated with dereferencing an internal object - the embedded object is laid out in the same memory space as the outer object, as if it were "part of" the object. The intuitive caveat is that the inner object must be constructed inside the constructor of the outer object.
+
+For actors, composition is desirable but it comes with even more overhead than for classes. That is, the outer actor can only communicate with the inner actor by sending asynchronous messages. This leads to several classes of problems and pain points:
+
+* Additional message passing overhead, including garbage collector tracing.
+* The outer actor cannot call synchronous methods of the inner actor, the behaviours will not be executed immediately, and any results must be communicated indirectly as messages instead of as return values.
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@jemc I think this could work well hand in hand with "asynclambda" to address a lot of pain points we've hit at Sendence. I'll get that written up sometime after I get back from vacation (aka post October 9).

SeanTAllen:
This probably results in a fairly decent chunk of new code, I think its reasonable to note the carrying cost of that code as a drawback.

Additionally, I would expect this change to introduce a lot of instability and bugs that we would be chasing for a while, I think its worth noting that as well.

SeanTAllen:
I find the overhead of cross actor messaging that you highlight very familiar. In the first versions of our application, user defined computations can in their own actors, however the overhead was considerably. We've since come up with rules that allow us to put multiple computations within a single actor. This has been a big performance boon for us. We were able to do this because those computations where defined as actors. I think that the goal of lowering that overhead for actors is quite laudable.

SeanTAllen:
I think people can rightly raise the following alternative:

You could accomplish this by putting the functionality you want to embed in a class which can then be used by various different actors.

Its a fairly valid alternative except, it falls apart in the case of ASIO events that have to be delivered to an event on an actor.

Another alternative would be to allow both classes and actors to receive messages. There are actor systems where this is how it operates. Sylvan and I have recently discussed the value of making Pony work in that fashion (there are many and "asynclambda" that I need to write up a RFC for is a step in that direction).

I think within how Pony currently works, none of the alternatives make sense however, I think before this goes to final comment that it makes sense to devote a special additional meeting of an hour or two for a conversation about the pain points you have experienced that this addresses and the pain points I've experienced that asynclambda addresses and see if there is a more holistic, deeper change to Pony that would be better for addressing both. I would say we have this meeting separate from the regular sync, allot 60 to 120 minutes and make sure it includes at least you, me and @sylvanc plus anyone else from the community who wants to be involved.

I think we've both experienced different problems with Pony's composition and messaging models and it behooves us to look at them at a global level and not just individually in two RFCs designed to address different aspects of what is perhaps a single larger problem.

SeanTAllen:
@jemc a question for clarification: if I call a behavior on an embedded actor, that would happen asynchronously, correct?

jemc:
@SeanTAllen I've added the drawbacks and alternatives you mentioned, thanks.

> it makes sense to devote a special additional meeting of an hour or two for a conversation

I would definitely welcome such a conversation - I think it would be a good idea.

> if I call a behavior on an embedded actor, that would happen asynchronously, correct?

Correct.  From a scheduler perspective, it would be just like calling an actor calling a behaviour on itself, which is also asynchronous.  The extra power would come in the fact that the inner actor could expose a synchronous API as well as an asynchronous one.

jemc:
We discussed on the sync call that there are some technical feasibility concerns related to how actor message queues and dispatch currently works, so this would either require a creative solution on that end, or it would be infeasible.  I'm going to leave the RFC open for a bit, because I want to try to investigate some creative solutions for that, but it may prove futile.

The main alternative discuessed was the first one in the RFC, do nothing different in the language, but encourage actors (like `TCPConnection`) to be written as classes that handle most of the functionality, but are wrapped in a thin layer of actor by the library, and potentially also by the user in a custom actor.  This is not necessarily ideal, given that it requires extra types for no reason other than extensibility, and does nothing to help with actors in libraries that you do not control.  I'm going to investigate in parallel what a synchronous `TCPConnectionCore` might look like.

Praetonus:
I've been thinking about the implementation side of the RFC and I might have found a solution with little overhead.

We could adapt the selector painting algorithm used for the vtable IDs to work for message IDs. It would allow us to make sure that the message IDs for a given actor and the message IDs for its embedded actors never overlap. Then the dispatch function for the enclosing actor would simply have to partition the possible IDs and dispatch to the right actor.

The sending side is a little more complicated. I think the most straightforward way is to add the enclosing actor's address in embedded actors and to use that address in the message sending functions, if the actor is indeed embedded. It would probably be possible to add that only on actor types that can be embedded so that there is no overhead on actor types that are never embedded.

jemc:
> It would probably be possible to add that only on actor types that can be embedded so that there is no overhead on actor types that are never embedded.

If I had an actor that embedded a `TCPConnection`, for example, does that mean that every instance of `TCPConnection` in the program would have to pay the cost of checking whether it is embedded or not?

Praetonus:
Yes. But I expect it would be possible to make that cost very low, for example by setting some of the low bits of the actor's address. That might interfer with proxy actors in distributed Pony so it would need to be implemented with that in mind.

I also think it would be possible to remove the check when a given actor type is always embedded and is never used in an interface with non-embedded actors. That might have a high impact on compilation times, though.

jemc:
If the design I proposed in https://github.com/ponylang/rfcs/pull/91#issuecomment-306005552 gets accepted as the design for #91, I'll close this one, as my motivating use case for this RFC will have been fulfilled.

Praetonus:
In case we still want to pursue this RFC, I think I've found a way to send messages to embedded actors with no overhead.

The basic idea would be to change the runtime layout of an actor to include pointers to the message queue and to the heap. For an embedded actor, these pointers would refer to the fields of the enclosing actor.

Then, when sending a message or allocating heap memory, we'd always dereference the pointers instead of accessing the fields directly.

In order to avoid additional cache misses, the pointers should be placed next to their respective fields. Also, this approach would mean that the pad for embedded actors is mostly unused memory.

SeanTAllen:
Closing as stale.

PR #42, opened 2016-10-16
Capability Based Access Modifiers

sirinath:
Please see: https://github.com/sirinath/rfcs/blob/master/text/0000-capabilities-based-access-modifers.md


SeanTAllen:
Thanks for taking the time to write this up. At the moment there are many issues that need to be addressed.

From a totally mechanical standpoint, there are issues with this PR. Please remove all the .idea files.

Additionally, the 0000-template.md file that contains your PR is supposed to be given a unique name like: 0000-capabilities-based-access-modifers.md

Please update those.

As to the content of the PR. There's no way I can support this at this time:

You are advocating for a huge change to the language without any supporting evidence of why this is advantageous. What's the motivation? Give concrete examples of problems this solves. What are you currently unable to do that this would provide. 

The how we teach this section is horribly empty. You are advocating for a huge change and for something that as far as I know exists in no other language. That requires a large amount of thought and great detail on how it should be taught. You have 1 sentence. There's a really large disconnect there.

Based on the scope of the change for this RFC, I would expect for there to be several pages of content for this.

sirinath:
I corrected the main 2 issues as they are the easiest. Also perhaps other can give ideas so I can incorporate them.

SeanTAllen:
the .idea files are still present in the PR. you need to repo them from your repo (git rm -rf .idea) and then commit and push again.

sirinath:
I am gett the following

```
C:\Users\...\IdeaProjects\rfcs>git rm -rf .idea
fatal: pathspec '.idea' did not match any files
```

jemc:
I don't understand how the proposed scheme is related to capabilities, even though you assert in the RFC that it is a extension of Pony's capabilities.  Honestly it seems more like an entirely new ACL-based system than a capabilities-based system that is an extension of the current one.

Also, I'm still not entirely clear what the Pony syntax would look like.  You gave some examples, but they are not in any clear context.  It would be good to include at least one example in the context of a complete Pony snippet.

sirinath:
Added an example

text/0000-capabilities-based-access-modifers.md
```diff
@@ -0,0 +1,72 @@
...
+
+A more concrete example
+
+class Foo
+  var x: U32 pvt[this]
+
+# How We Teach This
+
+This is using capabilities as means to express access privileges hence more in line 
+with with another language construct than something completely new.
```
aturley: While this may (or may not) be true, it doesn't address how the concept should be taught. Or are you suggesting that no additional teaching needs to take a place?

text/0000-capabilities-based-access-modifers.md
```diff
@@ -0,0 +1,72 @@
...
+
+# How We Test This
+
+You should be able to specify:
+
+- Type state / phantom type 
+- Friend functions (class / actor / package specific access)
+- Different protection levels
```
aturley: Can you give some details about what tests for each of these features would look like, either based on our current testing infrastructure, or any new infrastructure that we would need to add?

text/0000-capabilities-based-access-modifers.md
```diff
@@ -0,0 +1,72 @@
...
+- pvt[A, B, C] - only an instance of A, B, C; generic type parameters A, B, C or package A, B, C
+- pvt[A #f, B #g] - only an in f and g of type A, B
+- pvt[(A :>) #f, (B <:) #g] - only an in f and g in a subtype of A and a supertype of B
+- pvt[A | A =:= B] - private to A when A is B
+
+A more concrete example
+
+class Foo
+  var x: U32 pvt[this]
+
```
aturley: Please place the above code inside triple-backticks so that it is correctly formatted in the document. It would also be helpful to state exactly what is happening here. I think this means that `x` would now be private to instances of of class `Foo`, but that's not entirely clear. If that is the case, consider showing what the code would look like currently, and compare that to what the code would look like using your proposed change.

text/0000-capabilities-based-access-modifers.md
```diff
@@ -0,0 +1,72 @@
...
+While capabilities is a language feature, Pony does not use this for the benefit 
+of access modifiers.
+
+# Motivation
+
+Capabilities based access modifiers will support more fine grain protection including:
+
+ - fine grain access modifiers
+ - type state / phantom types
```
aturley, line 16: Since Pony doesn't provide phantom types, it might help to include a little more information about them, either here or in the "Detailed Design" section.

sirinath:
Anything else I should add / change?

jemc:
> I don't understand how the proposed scheme is related to capabilities, even though you assert in the RFC that it is a extension of Pony's capabilities. Honestly it seems more like an entirely new ACL-based system than a capabilities-based system that is an extension of the current one.

SeanTAllen:
None of the real issues I raised have been addressed:

```
You are advocating for a huge change to the language without any supporting evidence of why this is advantageous. What's the motivation? Give concrete examples of problems this solves. What are you currently unable to do that this would provide.

The how we teach this section is horribly empty. You are advocating for a huge change and for something that as far as I know exists in no other language. That requires a large amount of thought and great detail on how it should be taught. You have 1 sentence. There's a really large disconnect there.

Based on the scope of the change for this RFC, I would expect for there to be several pages of content for this.
```

SeanTAllen:
Closing as stale.

PR #43, opened 2016-10-16
Add RFC for more Iter methods

Theodus:
[Rendered](https://github.com/Theodus/rfcs/blob/iter/text/0000-more-iter-methods.md)

### DIFF
--- a/text/0000-more-iter-methods.md
+++ b/text/0000-more-iter-methods.md
@@ -1,0 +1,95 @@
+- Feature Name: more-iter-methods
+- Start Date: 2016-10-16
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the addition of various methods to the Iter class of the itertools package. The methods included are based on functions from [Rust's Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html) and [Elixir's Enum](http://elixir-lang.org/docs/stable/elixir/Enum.html).
+
+# Motivation
+
+The purpose of these additional methods is to increase the functionality of the Iter class.
+
+# Detailed design
+
+The following methods will be added to the Iter class:
+```pony
+fun ref all(f: {(A!): Bool ?} box): Bool =>
+    """
+    Return false if at least one value of the iterator fails to match the
+    predicate `f`, otherwise return true.
+    """
+
+fun ref any(f: {(A!): Bool ?} box): Bool =>
+    """
+    Return true if at least one value of the iterator matches the predicate `f`,
+    otherwise return false.
+    """
+
+fun ref collect[B: Seq[A!] = Array[A!]](coll: B): B^ =>
+    """
+    Push each value from the iterator into the collection `coll`.
+    """
+
+fun ref count(): USize =>
+    """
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Does the "how do we test this" translate to "unit tests"?

## 

For "how do we teach this", I think that the high level package documentation in itertools.pony needs to be updated for previous iter changes and this. that package level documentation looks pretty out of date. agreed?

Theodus:
@SeanTAllen Agreed. I'll make changes to update the documentation and clarify that unit tests are required.

text/0000-more-iter-methods.md
```diff
@@ -0,0 +1,95 @@
...
+The purpose of these additional methods is to increase the functionality of the Iter class.
+
+# Detailed design
+
+The following methods will be added to the Iter class:
+```pony
+fun ref all(f: {(A!): Bool ?} box): Bool =>
+    """
+    Return false if at least one value of the iterator fails to match the
+    predicate `f`, otherwise return true.
```
jemc: Is it short-circuiting?  (Do all of elements get the function executed, or does execution stop when a `false` is reached?)  This should be made clear in the doc string.

text/0000-more-iter-methods.md
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+fun ref any(f: {(A!): Bool ?} box): Bool =>
+    """
+    Return true if at least one value of the iterator matches the predicate `f`,
+    otherwise return false.
```
jemc: Is it short-circuiting?  (Do all of elements get the function executed, or does execution stop when a `true` is reached?)  This should be made clear in the doc string.

SeanTAllen:
@Theodus you ready for final comment on this?

Theodus:
@SeanTAllen Yes I am.

Theodus:
The method `run` has been added to iterate over the values of the iterator without a for loop.

SeanTAllen:
@Theodus `run` is missing a return type. or if it isn't, I don't understand how it is used.

Lots of examples in the package docstring would be a good addition.

Theodus:
@SeanTAllen would there be a reason to return a spent iterator?

SeanTAllen:
@theodus how would you use this run method? I don't understand how it's useful. I must be missing something.

Theodus:
@SeanTAllen With the `run` method you can do the following:

The following snippet prints the numbers 1-3.

``` pony
let xs = Iter[I64](Range[I64](1, 4))
  .map[None](lambda(x: I64)(env) => env.out.print(x.string()) end)
  .run()
```

Without the `run` method, you would have to do the following:

``` pony
let xs = Iter[I64](Range[I64](1, 4))
  .map[None](lambda(x: I64)(env) => env.out.print(x.string()) end)

for x in xs do None end
```

SeanTAllen:
so map is lazy?

Theodus:
@SeanTAllen Many of the methods return iterators rather than evaluating, so yes.

SeanTAllen:
Ah, that makes sense. Definitely something good for the documentation. Discuss which are "lazy" vs which are "eager".

PR #44, opened 2016-10-17
Branch prediction intrinsics

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/branch-prediction/text/0000-branch-prediction-intrinsics.md) (new link).

### DIFF
--- a/text/0000-branch-prediction-intrinsics.md
+++ b/text/0000-branch-prediction-intrinsics.md
@@ -5,36 +5,17 @@
 
 # Summary
 
-Add variations of control structures to enable the user to make the optimiser aware of the likelihood of a given condition.
+Add program annotations to control structures to enable the user to make the optimiser aware of the likelihood of a given condition.
 
 # Motivation
 
-Users often know more than the optimiser about the high level properties of their code. Having these constructs would allow users to give hints to the optimiser for better code generation.
+Users often know more than the optimiser about the high level properties of their code. Having these annotations would allow users to give hints to the optimiser for better code generation.
 
 # Detailed design
 
-Add `likely` and `unlikely` suffixes to the `if`, `elseif`, `while` and `until` control structures. These variations are also added to `match` statements, with the following syntax.
+Add `likely` and `unlikely` annotations able to affect the `if`, `elseif`, `while` and `until` control structures, as well as individual cases of a match expression (by annotating the `|` symbol).
 
-```pony
-match x
-| likely None => // ...
-| unlikely let x': A => // ...
-end
-```
-
-When a `likely` conditional branch is evaluated, the optimiser is allowed to assume that the associated value is probably true and to generate code with fast execution paths where the branch is taken. `unlikely` works the same way but assumes that the branch is probably not taken. There are two constructs to avoid weird reversed conditions in some cases.
-
-The base construct and the variations are considered equivalent from the grammar standpoint so it is possible to write this kind of construct:
-
-```pony
-iflikely a then
-  // ...
-elseif b then
- // ...
-elseifunlikely c then
- // ...
-end
-```
+When a `likely` conditional branch is evaluated, the optimiser is allowed to assume that the associated value is probably true and to generate code with fast execution paths where the branch is taken. `unlikely` works the same way but assumes that the branch is probably not taken. There are two annotations to avoid weird reversed conditions in some cases.
 
 Examples:
 
@@ -42,7 +23,7 @@ As data on high-level properties.
 
 ```pony
 fun process(id: USize) =>
-  iflikely id == _common_id then
+  if \likely\ id == _common_id then
     // process common IDs
   else
     // process other IDs
@@ -53,7 +34,7 @@ As an optimisation when function contracts are violated.
 
 ```pony
 fun sqrt(x: F64): F64 ? =>
-  ifunlikely x < 0 then
+  if \unlikely\ x < 0 then
     error
   end
   // compute square root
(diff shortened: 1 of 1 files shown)


text/0000-branch-prediction-intrinsics.md
```diff
@@ -0,0 +1,73 @@
...
+
+This is a basic concept, the documentation string should be sufficient.
+
+# How We Test This
+
+These intrinsics will directly map to LLVM's intrinsics. We would assume that their implementation works.
+
+# Drawbacks
+
+None.
```
jemc: I'd say a significant drawback is that since these have to go in `builtin` (being compiler intrinsics) it means that a user can never create a type called `Likely` or `Unlikely`.

jemc:
I was just thinking about this the other day when @SeanTAllen was talking about his application being performance-sensitive to `if` branches.

I see this as an important addition to Pony, and it sounds rather easy-to-implement as well, since LLVM provides the appropriate intrinsics.

SeanTAllen:
I 100% want this. Very much want this.

re: the how do we teach: I think in addition to the doc-string, a short Pony pattern should also be done for this.

text/0000-branch-prediction-intrinsics.md
Praetonus: I'll add that. Maybe we need to come up with a way to scope some of the names in `builtin`.

jemc: Or alternatively, a way to have `compiler_intrinsic` methods in non-`builtin` packages of the standard library.

Praetonus:
> re: the how do we teach: I think in addition to the doc-string, a short Pony pattern should also be done for this.

Yes, it could be useful. I'll add it.

jemc:
@Praetonus - seems like a simple RFC that is pretty uncontroversial so far - want to go to final comment period today so that this can see a vote next week?

Praetonus:
Yes, I think it would be good to see it move forward quickly.

malthe:
The docstrings could mention that `__builtin_expect()` is "primarily useful in code which is complex enough to overflow the branch history buffer" and that the cost of being wrong is significant.

Praetonus:
@malthe Added.

SeanTAllen:
@Praetonus could Likely/Unlikely be used with match statements?

Praetonus:
Not as is. I think it would require changes in the code generator itself as a `match` generates a lot of boilerplate code, and we'd want `Likely`/`Unlikely` to affect a part of that code.

Praetonus:
Actually, I realised it could be implemented much more efficiently directly within the code generator because of the way LLVM's intrinsics and profiling metadata work. It could work correctly with compiler intrinsics but it would require some overhaul in the compiler to couple the intrinsics and the code generator more tightly. That overhaul might be necessary at some point but right now I don't think it's the right thing to do.

I'm going to amend the RFC to replace the `*likely` primitives with something embedded in the control structures themselves.

I'll remove "ready for vote" from the RFC but since it's mostly a change of "shape", maybe we'll be able to get back to it quickly.

Praetonus:
Amended. I'm not happy with the syntax for `match` but I couldn't think of anything better. I strongly welcome ideas on that side.

Praetonus:
Going back to final comment period.

georgemarrows:
To me, the weight of the additional syntax doesn't feel in balance with the edge-case-ish functionality. (Though I understand that Sean may not think it is an edge-case!) 

Is there an annotation mechanism that could be extended to allow annotations in the necessary locations? If not is it worth adding one to avoid the special syntax here?

So eg, with unthought-through `[annotation]` syntax, especially from a LR(1) point of view...

```
match x
| [likely] None => // ...
| [unlikely] let x': A => // ...
end


if [likely] a then
  // ...
elseif b then
 // ...
elseif [unlikely] c then
 // ...
end
```

Praetonus:
This is an interesting idea. We don't have an annotation system now, but adding a way to arbitrarily annotate code (which would translate to annotations to ASTs in the compiler), with some parts in the compiler able to look for specific annotation strings. It could indeed be useful for things that don't really need their own keywords or syntax. It could even be good to allow custom annotations, for things like @jemc's idea of a library to process Pony ASTs in custom ways (e.g. for third-party macros). Would you like to write up a RFC for that @georgemarrows? If not, I'll think about it.

Also, note that the `[]` syntax in your example is ambiguous because of array literals.

georgemarrows:
I don't think I could do justice to an RFC on annotations: it feels like it touches on too many potential areas and use cases for my level of Pony knowledge.

The only annotation system I really know is Java's attributes. There is a very big difference in that the annotations attach to the runtime representation of classes and methods, not the compile-time AST.

Perhaps closer is the Rust annotation system, which is not currently user-extensible: https://doc.rust-lang.org/reference.html#attributes  It sounds like it's not user-accessible either - just directives for the compiler. Also perhaps `#[]` stands a better chance of being parseable?

Praetonus:
I'll write up something then, thanks for the initial idea.

Using `#` could work now but unfortunately the upcoming compile-time values will use `#` and the two would be ambiguous.

Praetonus:
Removing final comment period since the RFC will undergo another major change.

georgemarrows:
> Using # could work now but unfortunately the upcoming compile-time values will use # and the two would be ambiguous.

That kind of knowledge is exactly why I don't feel confident about having a go myself! Will be interested to see what you propose.

Praetonus:
I've amended the RFC to use annotations instead of keywords. I'll put the RFC into final comment period, hopefully for the last time.

SeanTAllen:
Annotations being so new gives me a bit of pause but in the end, I'm happy to see them start getting used so we can shake out issues with them. I'm in favor of this RFC.

sylvanc:
I vote yes :)

SeanTAllen:
I vote yes.

Theodus:
I vote yes.

SeanTAllen:
@jemc if you are good with this, then we can merge this and do all the usual housekeeping for setting up the issue etc.

jemc:
:+1:

sblessing:
I vote yes as well J

Praetonus:
I'm going to merge and set up the issue on the ponyc repo.

PR #45, opened 2016-10-18
Non-blocking DNS

malthe:
[Rendered](https://github.com/malthe/pony-rfcs/blob/non-blocking-dns/text/0000-non-blocking-dns.md)

### DIFF
--- a/text/0000-non-blocking-dns.md
+++ b/text/0000-non-blocking-dns.md
@@ -1,0 +1,45 @@
+- Feature Name: Non-blocking DNS
+- Start Date: 2016-10-18
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+DNS name resolution should be non-blocking (and asynchronous) from a user perspective.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


aturley:
@malthe can you change the title to something like "Non-Blocking DNS"? That will make it easier to identify from the list of pull requests.

SeanTAllen:
"This will require some minor changes to network test code" isn't really how we will test.

How exactly will we test this?

## 

How will we teach... 

What documentation exactly should be updated?

## 

Alternatives:

Leave as is.

## 

Re: detailed design,

I'd like to see actual implementation details. Some implementations would be no better than the existing (spawning a different actor that blocks doesn't do us any good), a thread pool for just this is problematic for thread pinning. The implementation details are critically important here.

malthe:
With regards to implementation details, should I go forward with suggesting an own implementation in Pony, or incorporate one of the existing C-libraries that implement their own DNS resolver.

Should we aim for full platform parity or might some platforms be left with a "no better than the existing" solution?

SeanTAllen:
Closing as stale.

PR #46, opened 2016-10-18
String split zero copy

malthe:
[Rendered](https://github.com/malthe/pony-rfcs/blob/string-split-zero-copy/text/0000-string-split-zero-copy.md)

Implementation is here: https://github.com/malthe/ponyc/compare/issue-1309...malthe:string-split-zero-copy

### DIFF
--- a/text/0000-string-split-zero-copy.md
+++ b/text/0000-string-split-zero-copy.md
@@ -1,0 +1,53 @@
+- Feature Name: String split zero-copy
+- Start Date: 2016-10-18
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The string splitting method should work on an immutable copy and provide a "zero-copy" functionality.
+
+# Motivation
+
+This behavior is preferable because you get to decide whether to make zero or one copy of the complete string. It's easy to obtain an immutable copy of a string so usability is preserved.
+
+# Detailed design
+
+The splitting method will be revised to operate on the immutable string only, returning immutable zero-copy reference strings to the same memory allocation.
+
+The previous method signature was:
+```pony
+fun split(delim: String = " \t\v\f\r\n", n: USize = 0): Array[String] iso^ =>
+```
+This would be changed to:
+```pony
+fun val split(delim: String = " \t\v\f\r\n", n: USize = 0): Array[String] val^ =>
+```
+
+Typical usage:
+```pony
+let words = "Hello world".split()
+let s = "1,2,3".clone()
+let numbers = s.split(",")
+```
+The latter example relies on auto-recover to make the cloned ``String iso`` immutable.
+
+# How We Teach This
+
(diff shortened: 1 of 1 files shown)


jemc:
Can you make explicit the method signature of the current `String.split` method as it compares to the new one?

SeanTAllen:
I need convincing. I can see having another split method that does this. We have/had that at Sendence. However changing split to only work on vals is not a good thing. Its perfectly reasonable for me to want to be able to split a ref. This change would disallow that. Rather than changing the existing method, I believe what you want is a new method.

malthe:
I suppose with subtyping it would be quite easy to provide two implementations of `String.split` – one for `val` and one for `ref`.

SeanTAllen:
What subtyping are you referring to @malthe?

malthe:
I was thinking that with https://github.com/ponylang/rfcs/pull/62 it might be possible to "overload" `String.split` with two implementations.

If that's something we want?

An alternative is of course to go with different names but that seems a bit wrong to me – like Hungarian notation almost.

I think ergonomics should be a concern here as well as obviously performance.

jemc:
I wonder how the performance of `(String box).clone -> (String val).split` compares to the current case of `(String box).split`.  It's possible that the `clone.split` would actually be _more_ performant, since it involves one large buffer allocation instead of many small buffer allocations.  If that ends up being true, I'd think that there is not much justification to keep `(String box).split` around.

It would be interesting to see some benchmarks.

SeanTAllen:
@jemc that would be an interesting benchmark.  did you mean:

`(String box).split` in the following?

> I'd think that there is not much justification to keep (String box).clone around.

jemc:
@SeanTAllen yes, thanks for the correction - I've corrected the comment.

malthe:
The benchmark results are:

```
split               20    82373935 ns/op
split-zc            30    44659338 ns/op
```

For what it's worth, https://github.com/ponylang/ponyc/pull/1335 actually runs about 1% faster than `"split"`.

Program code:

``` pony
use "ponybench"

actor Main
  let length: USize = 1024 * 1024

  new create(env: Env) =>
    let s = recover ref String.create(length) end

    while s.size() < length do
      s.append("abcde ")
    end

    let s': String val = s.clone()
    let bench = PonyBench(env)

    bench[None]("split", recover val lambda()(s') => s'.split() end end)
    bench[None]("split-zc", recover val lambda()(s') => s'.clone().split_zc() end end)
```

jemc:
@malthe thanks for doing this benchmark.

I cloned your fork and tried a few variations on the benchmark, and couldn't come up with a case where the `split()` wasn't slower than `clone().split_zc()`.

With that in mind, I think there is no reason to keep a `fun split` because `fun val split` is more optimal, even when you have to `clone` first to get a `val`.

SeanTAllen:
Closing as stale.

PR #47, opened 2016-10-19
Stateful primitives

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/stateful-primitives/text/0000-stateful-primitives.md).

### DIFF
--- a/text/0000-stateful-primitives.md
+++ b/text/0000-stateful-primitives.md
@@ -75,7 +75,7 @@ This will add a certain amount of code in the compiler, which will have to be ma
 
 - Use separate objects from primitive functions as described in Motivation. This can be really bad for performance.
 - Pass data in a class. This can break encapsulation if the data is supposed to be intrinsic to the algorithm.
-- Use the upcoming compile-time values. This can't work with FFI objects.
+- Use the upcoming compile-time values. A compile-time value can be used to create an arbitrarily complex immutable object. However, no FFI can be involved in it, so it isn't a usable alternative if the data wraps calls to a C library (for example, a huge number library).
 
 # Unresolved questions
 
--- a/text/0000-stateful-primitives.md
+++ b/text/0000-stateful-primitives.md
@@ -27,6 +27,8 @@ This pattern is far from optimal since a new object is created each time one of 
 
 The core idea is really simple, it boils down to allowing fields on primitives. Since a primitive is always `val`, viewpoint adaptation will ensure that fields can never be mutable and that they can be used without concern in multiple actors. The only difference with class fields is that primitive fields will have to be initialised in the `_init` function rather than in the constructor.
 
+We don't want to allow actors in primitive fields because it would create patterns with globally accessible actors not created by the main actor. Since a primitive `_init` can't send messages, that restriction is already ensured.
+
 ## Isn't this ambient authority?
 
 No, for two reasons.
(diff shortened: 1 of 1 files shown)


jemc:
> the function and the field in the following primitive are semantically equivalent (with the exception of object identity but it is irrelevant here)

I'm not so sure you can say that object identity is irrelevant.  It may not matter much in the case of immutable data structures but that is not all you are allowing with this change.  You say that viewpoint adaptation would take care of making sure that immutable data is all that we could allow to be global, but recall that `tag` (which is still `tag` when viewpoint-adapted by `val`) is all that is required to send messages to an actor.  And this is also where object identity is very much important, regarding the distinguishing between "the function" and "the field" - "the function" creates a new actor on each invocation, and "the field" acts like a global reference to a single actor

Thus I could use fields-on-a-primitive to create a global mutable actor that runs without ever being explicitly created (and given the authority to execute) by `Main`.  I think this does count as ambient authority in a real way, whether we see it as an otherwise positive thing or not.

It's also possible that the primitive-as-entrypoint-for-a-global-actor pattern would be problematic in distributed Pony.

If we decided we wanted to adopt this RFC, but somehow make it illegal for the fields to contain an `actor` (even indirectly), I think this would end up losing the attractive simplicity it currently holds of just using viewpoint adapation and letting the normal rules apply.

jemc:
In terms of alternatives, I think there is an important one that is not on this list.

I think Luke Cheeseman's fork with dependent types and compile-time expressions may actually fulfill a lot of the motivations for this change.  Compile-time expressions can be arbitrarily complex immutable data structures, but they may not contain actors.

Praetonus:
Actors were on my mind but I forgot to add them in the actual text... My bad. Of course, actors should be disallowed in primitives fields. Actually we can have that guarantee for free since a primitive `_init` can't send messages and therefore can't create actors.

Compile-time expressions can solve part of the problem, but not everything. Specifically, if the data contains FFI stuff (like my SRP example, which uses the GMP C library) then it won't work.

jemc:
> Actually we can have that guarantee for free since a primitive `_init` can't send messages and therefore can't create actors.

Right, I forgot about that constraint on `_init`.  I suppose this wouldn't be inconsistent by that metric, then.

Can you add some of this info to the text of the RFC?

jemc:
Heh, actually looks like you just did. :smile:

SeanTAllen:
I am generally in favor of this idea. Seems useful. I think more about Luke's work should be included in "Alternatives" including your statement:

> Compile-time expressions can solve part of the problem, but not everything. Specifically, if the data contains FFI stuff (like my SRP example, which uses the GMP C library) then it won't work.

SeanTAllen:
Given the participants on the RFC are 3 of the most active, I think it is reasonable to take this to final comment period soon. Thoughts @jemc and @Praetonus?

Praetonus:
I'm comfortable with moving it forward.

Praetonus:
> I think more about Luke's work should be included in "Alternatives"

I added that.

Praetonus:
It's been more than a week since the beginning of final comment period, marking ready for vote.

jemc:
We want to get some input from @sylvanc before voting, in case he has any reasons not to do this that we didn't think of justifications for here.

Praetonus:
We discussed this RFC during the sync call and it doesn't seem to be the best candidate for that use case. What we'd like to do is to use the upcoming value dependent types work to allow creating constant objects, but also allow these objects to use FFI as a way to "finish" their initialisation at runtime. I'll close this RFC and open a new one once value dependent types are available.

PR #49, opened 2016-10-20
Unsafe operators

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/unsafe-operators/text/0000-unsafe-operators.md).

### DIFF
--- a/text/0000-unsafe-operators.md
+++ b/text/0000-unsafe-operators.md
@@ -35,7 +35,7 @@ Bitwise `not`, `or` and `and` are always fully defined so there is no need for u
 
 # How We Teach This
 
-This shoud be mentioned in the future documentation for unsafe operations.
+This should be mentioned in the future documentation for unsafe operations. Unsafe operators could be described quickly in the basic arithmetic section of the tutorial as a mean for advanced users to get better performance. In addition, the new symbols will have to be added to the symbol appendix in the tutorial.
 
 # Drawbacks
 
--- a/text/0000-unsafe-operators.md
+++ b/text/0000-unsafe-operators.md
@@ -1,0 +1,50 @@
+- Feature Name:unsafe-operators 
+- Start Date: 2016-10-20
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add a set of operators mapping to the unsafe math operations.
+
+# Motivation
+
+As discussed in ponylang/ponyc#993, it has been decided to separate the various mathematical and logical operations in the Pony language into two separate sets, a set of functions with fully defined semantics and another set of functions that can have undefined results but can have better performance.
+
+Since `x + y` is more convenient than `x.add(y)`, we also need to come up with operators sugaring to the `*_unsafe` functions.
+
+# Detailed design
+
+The proposed syntax is `op~`. The full list of required operators is
+
+- `+~` (`add_unsafe`)
+- `-~` (`sub_unsafe`)
+- `*~` (`mul_unsafe`)
+- `/~` (`div_unsafe`)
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Can you make `ponylang/ponyc#993` a link to the issue?

SeanTAllen:
"This shoud be mentioned in the future documentation for unsafe operations."

What future documentation? The how do we teach this section is very much about where do we document this, how do people learn about this feature etc. Is there a section of the tutorial where this should be added? Should we add a new section? etc etc. That sort of thing.

"Drawbacks: None"

SeanTAllen:
I'm slight confused by this RFC. Under alternatives, it discusses different sugar operations, while leaving the *_unsafe, methods as defined. However, those methods don't currently exist. What are the alternatives to the *_unsafe methods?  

It seems like a drawback to this is additional cost in terms of code, maintenance, possible bugs, additional documentation, larger API. Additionally, we are allowing programmers to create programs that might not be correct. The pony philosophy states correctness before performance. I think we should point out that all of this is true but in this case, you are opting into unsafe (much like C FFI).

So an alternative is to not introduce these unsafe options in the name of correctness which means sacrificing some possible performance gains that could be achieved via unsafe operations.

Given all of this, I think its rather important that we come up with a cohesive, "how to teach" strategy.
For example, I can see *_unsafe becoming an accepted naming convention and we mostly teach what that naming pattern means (probably somewhere in the tutorial - I don't know where at the moment) and can then document in docstrings, what in particular about any _unsafe method makes it unsafe.

Praetonus:
In ponylang/ponyc#993 and at the sync meeting, we argued that the current discrepancy in the language with mathematical operations (some fully defined, some with possible undefined results) was a bug and that it didn't need a RFC by itself.

It is true that programmers are allowed to get incorrect results with unsafe operations, in the same way that they are allowed to get incorrect results if they implement an algorithm badly. In both cases you're never going to stomp random memory or some other nasty thing. This is very different from FFI, where "all bets are off".

I think the "correctness before performance" philosophy is respected since the main point of the safe/unsafe distinction is to guarantee correctness by default on every operation, while still allowing an escape hatch for performance. "Unsafe" here means that if a visible side-effect depends on undefined result, that side-effect won't necessarily be observed in the intended way (e.g. a message might not be sent). As long as uncontrolled FFI isn't involved into the mix the program will still be well-mannered and won't do awful things.

Regarding the documentation and teaching ways, I plan to add precise docstrings for unsafe functions. The best place to talk about the semantics of "undefined results" (as defined in ponylang/ponyc#993) would probably be the language specification but until we start that the tutorial looks like the best choice. Probably in the "Trust boundary" section.

Should I add that to "How we teach this" even if the RFC isn't about the unsafe operations themselves?

SeanTAllen:
Ah I understand much better now, I misunderstood the scope of the RFC. Thus my confusion. I think the precise doc strings is enough.

jemc:
Under how "we teach this" - I think we'll also need to add the new symbols to the appendix at the end of the tutorial.  We also might want to mention in the basic arithmetic section of the tutorial that there are "unsafe" operations available for performance critical applications that can deal with or avoid undefined behaviour in those operations.

SeanTAllen:
Excellent idea @jemc.

Praetonus:
@jemc Added.

SeanTAllen:
@Praetonus I feel like this is ready for final comment period, agreed?

Praetonus:
@SeanTAllen Yes.

Praetonus:
It's been more than a week since the beginning of final comment period, marking ready for vote.

PR #50, opened 2016-10-21
Remove default failure implementations from network notify interfaces

SeanTAllen:
Remove default implementations currently provided for failures on
`TCPConnectionNotify`, `UDPNotify` and `TCPListenNotify`.


SeanTAllen:
@ponylang/committer any comments? if not, i'm going to move to final comment period.

jemc:
I'm in general agreement.

jemc:
Added final comment period label, so that it can be voted on next week if there are no comments.

SeanTAllen:
Sounds good. I probably won't be around to advocate for it next Wednesday.

PR #51, opened 2016-10-21
Add RFC for concise lambda syntax.

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/0c6aebdf80ef61a2acc2d990506f9f4639f25b13/text/0000-concise-lambda-syntax.md).

### DIFF
--- a/text/0000-concise-lambda-syntax.md
+++ b/text/0000-concise-lambda-syntax.md
@@ -1,0 +1,95 @@
+- Feature Name: concise-lambda-syntax
+- Start Date: 2016-10-20
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This proposal shortens the lambda literal syntax to match the concise lambda type syntax.
+
+# Motivation
+
+In an actor language like Pony where all work crossing an actor boundary must take place asychronously, lambdas are a critical feature, whether chained to promises or as direct callbacks to asynchronous operations.
+
+However, in my experience with Pony codebases lambdas aren't as widely used as you might otherwise think, and I believe this is due to a set of usability issues that make dealing with lambdas more painful than it should be. One of these issues is that they are not very concise in syntax.
+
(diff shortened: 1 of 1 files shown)


jemc:
Note that this syntax change proposal was originally suggested by @sylvanc in passing, shortly after the lambda type syntax was added: https://github.com/ponylang/ponyc/issues/415#issuecomment-191354626

malthe:
Is it absolutely necessary to wrap in curly braces – in terms of grammar – ?

It would be clearer without them I think, but that perhaps goes against Pony's easy to parse grammar.

jemc:
@malthe - Parse-wise, I think it will be necessary both to have a token at the beginning that identifies what comes next, and a token at the end to mark the end of the lambda and the beginning of the code that is to follow.  So I think the answer is yes - if we don't want to make the Pony grammar more complex than LL(1) we need both the opening brace and the closing brace.

jemc:
Also, even if we decided it's okay to make pony harder to parse, note that having no ending terminator makes it ambiguous - there's no telling when the lambda code ends and the outside-of-lamba code begins (well, without using a whitespace-significant grammar like python or coffeescript).

``` pony
let f = (db: Database, doc: Document) =>
  doc.foo = "foo2"
  doc.bar = "bar2"
  db.update(doc)

let x = "foo" // is this inside the lambda or outside of it?
```

malthe:
True – but in your example, parentheses sort of "help" to delimit the lambda. Perhaps the `lambda ... end` could simply be dropped in that fairly general case:

``` pony
let c = a.map[U32]((a: U32): U32 => a * 2)
```

jemc:
@malthe - the lack of the starting marker in your example still causes LL(1) problems, as far as I can see.  Using the parentheses in this way probably also causes LL(1) problems, but I'm not as sure about that.

Also, I think it becomes pretty awkward to look at and parse when there are multiple parameters:

``` pony
foo.bar((a: U32): U32 => a * 2, 5)
```

``` pony
promise.next[U32]((a: U32): U32 => a * 2, (): U32 => 99)
```

Also, if we abandon the curly braces, then the benefits of matching the lambda type syntax are lost.

jemc:

PR #62, opened 2016-10-23
Subtype checking

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/subtype-checking/text/0000-subtype-checking.md).

### DIFF
--- a/text/0000-subtype-checking.md
+++ b/text/0000-subtype-checking.md
@@ -40,16 +40,16 @@ The single keyword `iftype` is proposed for consistency with the `ifdef` keyword
 
 ## Specialised generic functions
 
-These function signatures are syntactically similar to case functions
+These function signatures are syntactically similar to case functions. A function without an `iftype` guard is called a default function and a function with an `iftype` guard is called a specialisation.
 
 ```pony
-fun foo(x: A): B => // Default implementation. Optional.
+fun foo(x: A): B => // Default function. Optional.
 fun foo(x: A): B iftype A <: B => // A is constrained to B in this specialisation.
 ```
 
-If the subtyping relationship isn't verified, the default implementation is used. If there is no default implementation, the function is completely removed from the associated type. Since `A` must be a generic parameter, functions would only be conditionally removed from specific reifications of generic types. This wouldn't impair type checking.
+If the subtyping relationship isn't verified, the default function is used. If there is no default function, the function is completely removed from the associated type. Since `A` must be a generic parameter, functions would only be conditionally removed from specific reifications of generic types. This wouldn't impair type checking.
 
-If multiple specialisations of the same function are supplied with different subtyping constraints, the implementation with the matching relationship is selected. If multiple implementations match, causing ambiguity, a compilation error is issued.
+If multiple specialisations of the same function are supplied with different subtyping constraints, the specialisation with the matching relationship is selected. If multiple specialisations match, causing ambiguity, a compilation error is issued.
 
 The type of every specialisation of a function must be a subtype of the type of the default function. If there is no default function, there is no subtyping requirement. This is safe because for any given reification of the function, only one specialisation can match.
 
--- a/text/0000-subtype-checking.md
+++ b/text/0000-subtype-checking.md
@@ -1,0 +1,86 @@
+- Feature Name: subtype-checking
+- Start Date: 2016-10-23
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add a language construct allowing programmers to get information on subtyping relationships. This construct could be used in conditionals to conditionally compile code, or in method signatures to conditionally compile a whole method.
+
+# Motivation
+
+There are two main areas that can be improved by this feature.
+
+First, there are a lot of cases where it would be useful to have a method on some reified versions of a generic type, but not in the general case. For example, a `string()` method could be added to the `Array[A]` class in the standard library when `A` is a subtype of `String`.
+
+Second, this can greatly help performance. Currently, the only way to "specialise" an implementation for a specific subtype is to use pattern matching on an interface (see `String.append` for an example). This has a runtime cost, which wouldn't exist in a compile-time specialisation. Combined with type parameter inference, it could be a nice way to provide user-friendly static dispatch in the language.
+
+# Detailed design
+
+The core idea is to add a new "is subtype" binary operator, working on types at compile-time. The expression `A <: B`, with `A` a generic type and `B` an arbitrary type (including complex types) would return true if `A` is a subtype of `B` and `false` otherwise. If the result of the expression is true, `A` becomes constrained to `B` (i.e. as if declared as `[A: B]`) for the rest of the current block of code. Since that last property can dramatically change the code that must be generated, the `<:` operator is only allowed in two specific cases, detailed later in the RFC.
+
+`A` can also be a tuple of generic types to allow checking multiple types at the same time, e.g. `(A1, A2) <: (B1, B2)`. In that case, `A1` becomes constrained to `B1` and `A2` becomes constrained to `B2`.
+
+If it can be proven that `A <: B` is always true (`A` is already constrained to `B`) or always false (`B` isn't a subtype of the existing constraint of `A`), a compilation error could be issued.
+
+## The `iftype` conditional
+
+The syntax of this new conditional is
+
+```pony
+iftype A <: B then
+  // Generate this block if A is a subtype of B.
+  // A is constrained to B in this block.
+else
+  // Generate this block otherwise
+end
+```
+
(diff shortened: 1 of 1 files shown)


jemc:
@Praetonus - thanks for taking the time to write this up.  I like your choice of syntax, and I don't disagree with anything in the proposal.

---

Would it be possible to create an `iftype`-guarded method that has no default implementation?  That is, the method would only be a part of the type when the guard clause held true.  This seems desirable to me, but it's not immediately clear if it would cause other problems in the type system.

Praetonus:
> Would it be possible to create an iftype-guarded method that has no default implementation? That is, the method would only be a part of the type when the guard clause held true. This seems desirable to me, but it's not immediately clear if it would cause other problems in the type system.

Yes, I talk about it in the specialised functions section. I think I'll reword that section a little bit because the vocabulary isn't consistent currently.

jemc:
Thanks, that change does make it clearer. :+1:

malthe:
It seems to me that unless you're matching on a specific type then the compiler won't be able to work around dynamic dispatch.

In the case of `String.append`, wouldn't it be good enough if you could tell the compiler to specialise the method for the two types `String box` and `Array[U8] box` – ?

Alas, this is something you get for free with a JIT-compiler.

malthe:
This might be an alternative syntax. Example taken from https://github.com/ponylang/ponyc/pull/1360.

``` pony
fun clone(): Array[this->A!] iso^ where A <: Any #send
```

It's a reuse of `where` but I think it's possibly easier to read and definitely say.

jemc:
@malthe - the motivation for using `iftype` is again the pesky LL(1) grammar requirement - it is a marker that lets the parser know what's coming next.  There is some discussion about this in the original ticket that formed the basis of this RFC: https://github.com/ponylang/ponyc/issues/683

However, after playing with the parser a bit last night I think it might be possible to parse `<:` in the grammar just like any other infix operator (without compromising LL(1)), then impose limits in the `syntax` pass for what is possible to put on the left hand side and right hand side.  I'll investigate a bit more and report back here.  If we can make it work, then we can indeed avoid adding the `iftype` keyword, and use the `<:` operator in "normal" (value-oriented) `if` statements and function guards.

``` pony
if A <: B then
  // Generate this block if A is a subtype of B.
  // A is constrained to B in this block.
else
  // Generate this block otherwise
end
```

Praetonus:
> It seems to me that unless you're matching on a specific type then the compiler won't be able to work around dynamic dispatch.

Yes, this will only be applicable when the program has some static informations about subtyping. But I think this is the common case, for example you rarely call `String.append` with a `ReadSeq`, you generally have a concrete object like a `String` or an `Array[U8]` to pass to the function.

> In the case of String.append, wouldn't it be good enough if you could tell the compiler to specialise the method for the two types String box and Array[U8] box – ?

An hypothetic implementation of `String.append` with that feature could be

``` pony
fun ref append[A: ReadSeq](seq: A, ...): String ref^ =>
  iftype A <: (String box | Array[U8] box) then
    // Fast append for byte sequences.
  else
    // Generic append for other stuff.
  end
```

That kind of function would be much more convenient to use once generic type inference is in.

> _`where` syntax_

IIRC, the syntax for guards was changed from `where` to `if` a while ago. I used `iftype` here mainly for consistency with other guards.

> However, after playing with the parser a bit last night I think it might be possible to parse <: in the grammar just like any other infix operator (without compromising LL(1)), then impose limits in the syntax pass for what is possible to put on the left hand side and right hand side. I'll investigate a bit more and report back here. If we can make it work, then we can indeed avoid adding the iftype keyword, and use the <: operator in "normal" (value-oriented) if statements and function guards.

I thought about that too but I don't think it would be good. Since `<:` can change the generated code, we can't treat the result as a plain boolean. The `iftype` block is a straightforward way to make sure that no runtime value will interfere. With a standard `if` block, we'd have to add lots of verifications and that would be error prone.

jemc:
> I thought about that too but I don't think it would be good. Since `<:` can change the generated code, we can't treat the result as a plain boolean. The `iftype` block is a straightforward way to make sure that no runtime value will interfere. With a standard `if` block, we'd have to add lots of verifications and that would be error prone.

You're right, those are good reasons to stay away from that line of approach.

Praetonus:
I've amended the RFC to allow things like this

``` pony
// C2 <: C1

fun foo() iftype A <: C1
fun foo() iftype A <: C2
```

The most specialised function (i.e. the one with the most specific subtype in the constraint) will be used, but there must be an appropriate subtyping relationship between the function types.

This is a generalisation of the default case, as `fun foo()` is the same thing as `fun foo() iftype A <: Any`.

jemc:
@Praetonus - good call.  This will allow for useful things like:

``` pony
fun foo() iftype A <: Integer[A]
fun foo() iftype A <: Any #share
fun foo()
```

malthe:
I don't think the `issubtype` operator is really necessary for what the proposal motivates. Is there something in the way for just having:

``` pony
fun foo(): C iftype A: Integer[A] =>
```

That is, to use the same semicolon as we already use in generics and use `iftype` only to impose a type constraint per-method.

I don't see how the following `iftype` avoids dynamic dispatch:

``` pony
iftype A <: (String box | Array[U8] box) then
  // Fast append for byte sequences.
else
  // Generic append for other stuff.
end
```

How can that be "fast" if the compiler anyway has to operate on A using dynamic dispatch due to there being a union of two types?

jemc:
@malthe - the principle motivation of this RFC is not to avoid all dynamic dispatch - it is to provide specialized implementations of generic methods that are resolved when the generic type is reified.

To use your same example, this RFC will allow us to do this (note that I've expanded the comment to be more explanatory of what will be going on):

``` pony
iftype A <: (String box | Array[U8] box) then
  // Fast append for byte sequences, using memcpy with the underlying pointer.
else
  // Generic append for other stuff, pushing bytes one-by-one in a loop.
end
```

Instead of this:

``` pony
match seq | let seq': (String box | Array[U8] box) =>
  // Fast append for byte sequences, using memcpy with the underlying pointer.
else
  // Generic append for other stuff, pushing bytes one-by-one in a loop.
end
```

Both examples allow us to specialize and use `memcpy` when we have a type that exposes a pointer to a contiguous chunk of memory.  The difference is that one case will specialize at compile time (when the method is reified) and the other will specialize at runtime (forcing a `match` cost every time the function is executed).  This is a winning scenario for performance, but the only catch is that we have to have the sufficiently specific type passed in as a type parameter (`A`) in order for it to work.

> I don't think the `issubtype` operator is really necessary

There is no `issubtype` keyword in the RFC, so not sure what you're talking about here.

SeanTAllen:
In case I haven't mentioned, I very very much want this feature. This would be huge for us with our application where we are paying that match cost a ton.

Praetonus:
@malthe Like @jemc said, since `A` is a generic type, the function is fully reified at compile-time with the concrete type, as if you wrote the function with e.g. `String` everywhere `A` appears. That means if we call the function with `String`, we will call `String.cstring()` directly, and not `(String box | Array[U8]).cstring()`.

The constraint of a generic is purely a compile-time concept for type checking, it doesn't mean we'll use the vtable of the constraint at runtime.

Praetonus:
> There is no issubtype keyword in the RFC, so not sure what you're talking about here.

I think @malthe is talking about `<:`.

Yes, it would be conceivable to use `:` for consistency. We could also use `<:` everywhere, including in the existing constraints, since `<:` is used in several languages (and in the Pony tutorial) to denote subtyping. That would break a lot of existing code, though.

jemc:
I'm in favor of keeping `<:` as this RFC currently describes, rather than using `:` for the new `iftype` conditions.  I think `<:` reads quite well when read as "is a", and I also think it would be better not to continue overloading new meanings on the `:` symbol.

I would also be in favor of using `<:` to specify type parameter constraints.  As you mention this would break code, but I think it would be more consistent and more readable overall.  However, I think it would be better to scope this effort in a different RFC.

malthe:
@jemc, thanks – now I get it. I think what I had somehow missed is that of course the `iftype` operator only restricts on the reified generic type or types.

jemc:
@Praetonus want to go to final comment period on this?

Praetonus:
@jemc I'd like to resolve the question about `A :< B` always true/false resulting in a compilation error before going into final comment period. Being able to identify redundancy would be good but it might prove frustrating to users when their otherwise valid code is deemed suboptimal by the compiler.

Praetonus:
Actually, since type checking is fully done before reification, the programmer doesn't know less than the compiler about subtypes and can easily figure out what's wrong. So maybe that's not a problem after all.

Praetonus:
Since nobody raised an argument against this, I'm going to amend the RFC and put it into final comment period.

Praetonus:
After giving it some additional thoughts, an error when `A :< B` is always true doesn't seem right. It will basically forbid the use of the function with every subtype of `(B & [constraint of A])`. I'll remove that.

The always false part is fine since it depends only on `B`, which can't be generic.

jemc:
This looks to be ready for a vote tomorrow, so I'm switching the label.

Please speak up if it's not actually ready for some reason.

Praetonus:
Looks like I forgot to update the RFC with what I said above. I've updated it. It's probably still ready for vote since it is a minor change.

SeanTAllen:
@Theodus i'm really busy right now, can you create the issue for this?

PR #63, opened 2016-10-26
initial RFC for bool <-> numeric conversions

nitbix:


Praetonus:
It seems to me that the only use case would be serialisation, to put the boolean in a "bag of bytes" like buffered `Reader`/`Writer`. Also, as I said on your original PR, I don't like having a universal conversion function for booleans because I think these boolean conversions should be made by context with the various comparison operators.

I'd say that the underlying problem is on the `Reader`/`Writer` interface, which should indeed have functions to read/write booleans and hide the conditional for the conversion from the user. If there are other serialisation interfaces in the standard library that don't allow boolean serialisation, I think they should be included in the RFC as well.

What do you think about shifting the scope of the RFC to only deal with serialisation interfaces?

jemc:
@Praetonus - note the other context when you may want to convert booleans to integers is when dealing with FFI - so it's not just a case of serialization.  However, I still contend that an explicit `if` statement is better for these cases.

PR #67, opened 2016-11-09
Amend method chaining

Praetonus:
The original reference capability semantics were slightly wrong. This is a modification @sylvanc and I talked about.


SeanTAllen:
@theodus can you create the issue for this as well?

Theodus:
@SeanTAllen wouldn't this fall under the current issue for the 'Sugar for Method Chaining' RFC?

SeanTAllen:
Didn't realize there was an existing issue

PR #68, opened 2016-11-15
Nuke "return this"

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/return-this/text/0000-nuke-return-this.md).

### DIFF
--- a/text/0000-nuke-return-this.md
+++ b/text/0000-nuke-return-this.md
@@ -127,8 +127,6 @@ Only public methods are included here. Private methods will be updated based on 
   - `seek`: will return `None`
   - `flush`: will return `None`
   - `sync`: will return `None`
-- `options.Options`
-  - `add`: will return `None`
 - `time.Date`
   - `normal`: will return `None`
 - `net.http.Payload`
--- a/text/0000-nuke-return-this.md
+++ b/text/0000-nuke-return-this.md
@@ -1,0 +1,158 @@
+- Feature Name: nuke-return-this
+- Start Date: 2016-11-15
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Remove the "return this" idiom used for chaining in some APIs of the standard library in favor of the new caller-side chaining feature.
+
+# Motivation
+
+With the recent implementation of caller-side method chaining, the "return this" idiom has become useless. It is also harmful to some compiler optimisations because it impedes precise alias analysis.
+
+# Detailed design
+
+Every method that currently uses "return this" for chaining will now either return `None` or return some useful result.
+
+Only public methods are included here. Private methods will be updated based on their usage in the internal API.
+
+- `builtin.Seq`
+  - `reserve`: will return `None`
+  - `clear`: will return `None`
+  - `push`: will return `None`
+  - `unshift`: will return `None`
+  - `append`: will return `None`
+  - `concat`: will return `None`
+  - `truncate`: will return `None`
+- `builtin.Array`
+  - `reserve`: will return `None`
+  - `compact`: will return `None`
+  - `undefined`: will return `None`
+  - `insert`: will return `None`
+  - `truncate`: will return `None`
+  - `trim_in_place`: will return `None`
+  - `copy_to`: will return `None`
+  - `remove`: will return `None`
+  - `clear`: will return `None`
+  - `push`: will return `None`
+  - `unshift`: will return `None`
+  - `append`: will return `None`
+  - `concat`: will return `None`
+  - `reverse_in_place`: will return `None`
+- `builtin.String`
+  - `reserve`: will return `None`
+  - `compact`: will return `None`
+  - `recalc`: will return `None`
+  - `truncate`: will return `None`
+  - `trim_in_place`: will return `None`
+  - `delete`: will return `None` (this method and `Array.delete` are inconsistent, `String.delete` does the same thing as `Array.remove`, and `String.remove` does something specific to `String`. A separate RFC to clarify that would be good)
+  - `lower_in_place`: will return `None`
+  - `upper_in_place`: will return `None`
+  - `reverse_in_place`: will return `None`
+  - `push`: will return `None`
+  - `unshift`: will return `None`
+  - `append`: will return `None`
+  - `concat`: will return `None`
+  - `clear`: will return `None`
+  - `insert_in_place`: will return `None` (`String.insert` and `Array.insert` are inconsistent. `String.insert_in_place` does the same thing as `Array.insert`)
+  - `insert_byte`: will return `None`
+  - `cut_in_place`: will return `None`
+  - `replace`: will return the number of occurrences replaced
+  - `strip`: will return `None`
+  - `lstrip`: will return `None`
+  - `rstrip`: will return `None`
+- `buffered.Reader`
+  - `clear`: will return `None`
+  - `append`: will return `None`
+  - `skip`: will return `None`
+- `buffered.Writer`
+  - `reserve`: will return `None`
+  - `reserve_chunks`: will return `None`
+  - number write functions (e.g. `u16_le`): will all return `None`
+  - `write`: will return `None`
+  - `writev`: will return `None`
+- `capsicum.CapRights0`
+  - `set`: will return `None`
+  - `unset`: will return `None`
+- `collections.Flag`
+  - `all`: will return `None`
+  - `clear`: will return `None`
+  - `set`: will return `None`
+  - `unset`: will return `None`
+  - `flip`: will return `None`
+  - `union`: will return `None`
+  - `intersect`: will return `None`
+  - `difference`: will return `None`
+  - `remove`: will return `None`
+- `collections.ListNode`
+  - `prepend`: will return whether the node was already in a `List`
+  - `append`: will return whether the node was already in a `List`
+  - `remove`: will return `None`
+- `collections.List`
+  - `reserve`: will return `None`
+  - `remove`: will return the removed element (the method is inconsistent with `Array.remove`)
+  - `clear`: will return `None`
+  - `prepend_node`: will return `None`
+  - `append_node`: will return `None`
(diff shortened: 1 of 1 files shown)


Praetonus:
Since this seems uncontroversial, going into final comment period.

jemc:
It feels like more of these methods may have useful alternative return values, but I'm okay waiting for those alternate uses to arrive organically in future RFCs.

That is, at some point someone will say "it sure would be great if `foo()` returned `bar`", and we can introduce the return value then.

Praetonus:
Here, I've added non-`None` return values mainly for information that can't be obtained otherwise (e.g. `String.replace`).

sylvanc:
Looks good.

ryanai3:
To be honest, I think this goes too far and breaks correctness.

For many of these methods, returning this is not just for chaining, but pertains to the interface.
By making the mutable array return nothing on .push(), we make it so that it cannot ever share a meaningful interface with an immutable collection of the same kind.
One of the great things about scala is that they share interfaces - that it's easy to swap in or out different immutable or mutable collections to test performance in runtime, memory, etc.
Yes, this may be more performant, but if you're at the position where you really care about the performance difference between chaining on return this and a .> call, then you're in the position where it makes sense to replace every .push() with .>push().

For the methods that would now return none, where tbh I don't really see anything else useful that could be returned (what could be returned on map.concat that the user can't get trivially with map.size() ?) what do we gain by removing them other than ideological purity on method chaining?

IMO in cases like this, correctness > performance, particularly when the performance will not matter for the vast majority of cases, and when it does, it's trivial for the person in that case to change.

Many of these certainly make sense, but not the ones that would be shared with immutable collections.

SeanTAllen:
@ryanai3 how do you think this breaks correctness?

ryanai3:
@SeanTAllen For a lot of these cases, nuking 'return this' means that mutable and immutable versions of the same collection cannot share a meaningful interface.
With return this, for example, both immutable and non-immutable map implement ```.concat(): Map```. IMO having a meaningful interface beyond just the constructor and .size() shared between the immutable and mutable versions of a class is "correctness" for a collections library.
This is because ```map.>add()``` has different meanings on a mutable and an immutable object, whereas ```map.add(): Map``` has the same meaning. Leaving return this in for those cases means you can swap in or out the immutable vs mutable versions of different collections, or even write your own specialized one.

sylvanc:
@ryanai3 , your point about mutable and immutable collection types is excellent. A common `interface`, in the sense of structural type, for dealing with both  `Map` and a `persistent.Map` seems like an important goal. And, by extension, for collection types in general.

Off the top of my head, the methods on an immutable type (such as the hash-array mapped trie in `persistent.Map`) that would have to return a new instance of the receiver type:

```
// currently exist on persistent.Map
update
remove

// doesn't currently exist on persistent.Map
insert
upsert
add
sub
```

text/0000-nuke-return-this.md
```diff
@@ -0,0 +1,158 @@
...
+  - `shared`: will return `None`
+  - `group`: will return `None`
+  - `private`: will return `None`
+- `files.File`
+  - `seek_end`: will return `None`
+  - `seek`: will return `None`
+  - `flush`: will return `None`
+  - `sync`: will return `None`
+- `options.Options`
+  - `add`: will return `None`
```
jemc: While writing a response to @ryanai3, I noticed this problem.

I think we should leave the `add` method as returning `this`, since the binary operator sugar `x + y` only makes sense when the result of the expression is the mutated `this`, which represents the combination of the two.

jemc:
@ryanai3 -

Because rcaps are part of the type system, a mutable collection already cannot share a `interface` with an immutable collection if that `interface` includes the mutation methods.  That is, the mutation methods of a mutable collection will be `fun ref` and return a `MyCollection ref`, and their counterparts on immutable collections will have to be `fun` or `fun val` have to return a new collection, as a ` MyImmutableCollection val`.  In the Pony type system, these methods will _never_ be part of a common `interface` - it's not structurally possible.  The only places where we can share an `interface` is on the read-only methods (for example, as in `ReadSeq`), and that commonality is not undermined by this change.

Can you provide more detailed reasoning for why you think this breaks "correctness"?  So far I haven't seen any evidence toward this claim.  I'm wondering if you only chose this word because that's the only concern that ranks above "performance" in Pony's hierarchy of concerns.

Also, the examples you've given so far aren't even accurate:

> both immutable and non-immutable map implement ```.concat(): Map```. 

This isn't true.  `collections.persistent.Map` does not implement a `concat` method at all.  Even if it did, the two methods could not possibly have the same interface in Pony, for the reasons I explained above.

> ```map.>add()``` has different meanings on a mutable and an immutable object, whereas ```map.add(): Map``` has the same meaning.

First of all, `collections.persistent.Map` does not implement an `add` method (though it probably should), but no one is proposing changing `collections.Map.add` in this RFC, since it is *not a mutation method* - it already returns a new `Map` that is a clone of the original with the given key/value pair added.  That is, this actually *could be* in a shared `interface` between the two because there is no mutation involved, and it's already working by copying.  The difference would be that `collection.Map.add` performance would be rather crappy by comparison, since it's basically a persistent-style operation on a collection that is optimized for mutation.

sylvanc:
I wonder if `persistent.Map.update` should become `persistent.Map.add` and `persistent.Map.remove` should become `persistent.Map.sub` ? That would result in the consistent implied `interface` @ryanai3 wants, wouldn't it? And without breaking the `fun ref` vs `fun box/val` distinction that @jemc quite rightly points out.

sylvanc:
I should have been more clear: then `persistent.Map` would _not_ have any of the logically mutating methods, such as `update`.

text/0000-nuke-return-this.md
Praetonus: The current signature of this function is `fun ref add(long: String, short: (None | String) = None, arg: ArgumentType = None, mode: (Required | Optional) = Required): Options`. I don't think `add` is meant as the binary operator sugar here. I'm not sure if another name would make more sense.

jemc:
@sylvanc - sounds right to me.

text/0000-nuke-return-this.md
jemc: I did find this use of the binary operator sugar in `examples/yield/main.pony`:

```pony
    let options = Options(env.args) +
      ("punk", "p", None) +
      ("lonely", "l", None) +
      ("bench", "b", I64Argument) +
      ("debug", "d", None)
```

I think this style is reasonable for sugar-oriented factory-style methods, though I personally would have done the formatting a bit differently:


```pony
    let options = Options(env.args)
    + ("punk", "p", None)
    + ("lonely", "l", None)
    + ("bench", "b", I64Argument)
    + ("debug", "d", None)
```

jemc:
@sylvanc - that is, it sounds like the right, consistent thing to do, though I would note that the two would still not be able to share an `interface` because they would have a different rcap on the return value (unless you want to hamstring `add` and `sub` on the mutable `Map` to not return mutable copies).

text/0000-nuke-return-this.md
Praetonus: I wasn't aware of that use. It does seem reasonable, I'll amend the RFC.

Praetonus:
@jemc It seems to me that a common interface would be possible, with a function returning `box` on the `interface`.

ryanai3:
@jemc, with calling on unions, couldn't you have a common interface with box as @Praetonus mentions? That's what I was thinking of. so you could have something like the following in your collections library:
```pony
interface ImmutableMap[K, V]

interface MutableMap[K, V]

type Map[K, V] is (MutableMap[K, V] ref | ImmutableMap[K, V] val)
```

when I said that both immutable and non immutable map implement .concat(), what I meant to say was that they should both implement it, and that it makes sense for both to implement it - typo.
While it may be true that Map.add() is not a mutation method, it is certainly true that there are many mutation methods in the nuke list that also function on the immutable versions (seq.append(), etc.). I just chose Map.add() off the top of my head.

jemc:
>While it may be true that Map.add() is not a mutation method, it is certainly true that there are many mutation methods in the nuke list that also function on the immutable versions (seq.append(), etc.). I just chose Map.add() off the top of my head.

Which immutable collection type implements `Seq.append`? Not `collections.persistent.List` - so I don't think that is an accurate example either...

This is why I'm asking to talk about specific, concrete examples here.  I think I agree with your point in a general sense, but when we look at the specific examples, the point breaks down and so far I haven't seen a specific example where it's still a concern.

ryanai3:
@jemc, sorry I meant that the immutable seq *should* implement an immutable .append(), not that it does so now. I would say that the lack of these methods currently simply reflects that people haven't had a chance to add those methods yet, not that they aren't necessary or shouldn't be there. 

Do you disagree that immutable seqs should implement a .append()?

jemc:
I do disagree, on the grounds that we should avoid using the same method name for two very different operations - it's a recipe for confusion.  `persistent.List.append` would imply mutation, which isn't possible to provide on an immutable collection.

As @sylvanc was bringing up, we can emulate immutable/persistent operations on the mutable collections (in a less performant way), but we *can't* emulate mutation operations on an immutable collection (unless we wrap it in a mutable wrapper object).

Probably `persistent.List.add` would more closely match the semantics of taking two immutable lists and calculating a new concatenated list from them.  We already have `persistent.List.concat` which does this, though arguably that needs to be cleaned up to avoid confusion with the `Array.concat`, which takes an iterator and is a mutation method - we should avoid this for the same reason I cited above - not using the same method name to refer to different operations.

jemc:
Also, regarding your comment about using a type union - Pony won't let you call methods on a type union unless all of the types in the union include that method with the same signature, so this doesn't buy you any slack in terms of the problem at hand of needing to have the same method signature to be used in a common `interface`.

Praetonus:
Since the conversation has settled, I'll mark the RFC ready for vote.

PR #69, opened 2016-11-16
Behaviour return None

Praetonus:
This was originally discussed in the [method chaining RFC](https://github.com/ponylang/rfcs/pull/4).

[Rendered](https://github.com/Praetonus/rfcs/blob/behaviour-return-none/text/0000-behaviour-return-none.md).

### DIFF
--- a/text/0000-behaviour-return-none.md
+++ b/text/0000-behaviour-return-none.md
@@ -1,0 +1,28 @@
+- Feature Name: behaviour-return-none
+- Start Date: 2016-11-16
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Change the return value of behaviour calls from the receiver to `None`.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


Praetonus:

sylvanc:
Looks good. Also improves the way a `be` is a subtype of a `fun tag`.

jtfmumm:
Nice.  So if I'm interpreting this correctly, you can now create interfaces that can be implemented by both classes via methods and actors via behaviors?

Praetonus:
@jtfmumm It is already possible. Currently, this kind of interface subtyping is valid

```pony
interface tag I
  fun tag foo(): I
  
class C
  fun tag foo(): C tag =>
    this
    
actor A
  be foo() => None
    
actor Main
  new create(env: Env) =>
    let i1: I = C
    let i2: I = A
    i1.foo()
    i2.foo()
```

With this RFC, it will be like this

```pony
interface tag I
  fun tag foo()
  
class C
  fun tag foo() => None
    
actor A
  be foo() => None
    
actor Main
  new create(env: Env) =>
    let i1: I = C
    let i2: I = A
    i1.foo()
    i2.foo()
```

jtfmumm:
Great, thanks.

PR #71, opened 2016-12-13
Support streaming in the http package

pdtwonotes:
This proposal makes changes to the net/http package to efficiently handle large payloads, including the streaming of large files both on GET and POST operations.  In essence, the existing `Payload` object changes from being just the *container* of the body to being the *transmission facilitator*.

### DIFF
--- a/text/0000-http-streaming.md
+++ b/text/0000-http-streaming.md
@@ -5,19 +5,19 @@
 
 # Summary
 
-Rewrite package/net/http server parts to improve memory efficiency and support streaming in various forms.
+Rewrite `package/net/http` to improve memory efficiency and support streaming in various forms.
 
 # Motivation
 
 The current design of the http client and server packages assumes that all responses are relatively small and that all the responses in progress, across all sessions, can be contained in memory at once.  This fails when the responses are large media files.  For example:
 
-* A stereo MP3 file occupies about 1 megabyte for every minute of audio.  A 3-minute Pop song therefore requires 3 MB.  A 20 minute symphony movement requires 20 MB.  Just a 3 minute music video can require 35 MB.
+* A stereo MP3 file occupies about 1 megabyte for every minute of audio.  A 3-minute Pop song therefore requires 3 MB.  A 20 minute symphony movement requires 20 MB.  And video can be 10 times larger per minute.
 
 * Buffering all of this data before sending even the first byte back to the client delays transmission needlessly, in addition to consuming a lot of memory.
 
-The HTTP protocol specification provides several ways around this problem.  One is simply to send the body data in smaller pieces, allowing TCP buffering semantics to deliver the long bytestream to the client.  No changes to the headers are required.  The other, for use when the RequestHandler really does not know the total length of the data in advance, is Chunked Transfer Encoding, where the response body is sent in smaller pieces, each with its own length header.  No Content-Length header is used at all in this case.
+The HTTP protocol specification provides several ways around this problem.  One is simply to send the body data in smaller pieces, allowing TCP buffering semantics to deliver the long bytestream to the client.  No changes to the headers are required.  The other, for use when the `RequestHandler` really does not know the total length of the data in advance, is Chunked Transfer Encoding, where the response body is sent in smaller pieces, each with its own length header.  No `Content-Lengt`h header is used at all in this case.
 
-This RFC proposes implementing both of these mechanisms in the Pony stdlib http package.
+This RFC proposes implementing both of these mechanisms in the Pony stdlib `net/http` package.
 
 # Detailed design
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@pdtwonotes can you elaborate on required documentation and testing.

pdtwonotes:
Most of the stdlib packages have no external documentation beyond the automatically generated web pages that get extracted from the doc-strings.  But in the source code there is sometimes quite extensive information in comments.  For example, net/TCPConnection has a long discussion of how backpressure works.   Some files even have little examples in them.  That would be easy enough.

Updating the examples/httpserver and client code could go a long way.

HTTP servers can get pretty complicated.  I don't know that we want to get into all the ways that a whole Server can be written, with dispatching of URL fragments, etc.    yet.  I have used the `cowboy` http package in Erlang, and it had an extensive pattern matching/dispatch function.

pdtwonotes:
For testing, an arbitrary amount of synthetic data could be generated, simulating reading from a file.  This way the test would not require the presence of any external files, and the amount of data transfered could be changed to test out different buffering thresholds and the backpressure mechanisms.  Backpressure can be tested by having the receiving end stall for a few seconds with a timer so that all the TCP buffers fill up and the backpressure notifications happen.  When the timer expires and the receiver starts reading again, the reverse should happen.

I just realized there needs to be a way to tell the *push* dleivery mechanism to pause, so that backpressure can propagate from a sink back to the TCP link, and thence back to the source.

SeanTAllen:
@pdtwonotes this sort of stuff should be in the RFC. its very good. it will give everyone implementing it a document to refer back to for what we agreed should be done to call the RFC done.

So, lets say its all documented in the docstrings for a moment.

Should it be the standard, simple example. More in depth? 

Why are docstrings enough to make it usable? 

Does a package of this scope require a new type of documentation we don't have?

SeanTAllen:
Re: backpressure @pdtwonotes. I've been working on it extensively for our product. I've ended up rewriting large chunks of the existing network code to support. What "backpressure" means is very vague. If backpressure is going to be part of the RFC, I think a lot more clarity in that area is needed. 

What for example is the goal of backpressure? Its a simple question but the answers could have really large ramifications.

pdtwonotes:
Backpressure is used in media streaming, for example.  Mp3 data is 'consumed' in the client at a rate of about 17KB/second.  Video data goes considerably faster.   Back on the server, a file is being read as fast as the file system can go.  The network can transfer the data at somewhere in between those speeds.   To prevent RAM being consumed all along the path to buffer up stuff which has been read from the file but not yet 'played' on the client, there has to be a way for the client to signal "stop sending until I catch up".   So the HTTP client code has to have a way to do that, which causes the TCPConnection actor to stop reading from the socket.  The underlying network code then stops sending ACK packets and eventually the server end of TCP stops sending.  The "throttled" event then happens on the server side, which has to be communicated to the HTTP session actor so it will stop reading from the disk file.

Then when the client catches up, it has to reverse this whole process, eventually causing the server to start reading from the file again.

Without this mechanism, RAM consumption in both client and server will balloon out of control.  This is particularly bad on the server which might be handling hundreds of these connections simultaneously.  The approach is "leave the data on disk until you know you can get rid of it".

SeanTAllen:
@pdtwonotes all of this should be in the rfc.

pdtwonotes:
I find *just* docstrings on classes and funs to be insufficient for anything more complicated than math functions.  When there are several interacting objects and methods, that have to be used in a certain order, you need a page of text or more, possibly with diagrams and examples, showing how all those things relate to each other.  The http package is among the most complicated in the stdlib.

SeanTAllen:
@pdtwonotes for that you can have package level documentation as well. is that sufficient?

See https://github.com/ponylang/ponyc/blob/master/packages/ponytest/pony_test.pony for example. There's a nice chunk of package level documentation for pony_test.

pdtwonotes:
The main problem I am dealing with is that the `Payload` object, which carries all the HTTP headers, method, URL, status, etc has to be delivered across actors first, as either an 'iso' or a 'val', but the body data has to be delivered separately due to its arbitrary size.   Currently the fields and methods relating to the body are inside the `Payload` and once it has been handed off there is no place to hang the body-management operations.

jemc:
@pdtwonotes I haven't looked over the details of the proposal yet, but I would suggest separating the concerns of delivering the request object to the handler (currently in `Payload`) and the concerns of sending a response (currently also in `Payload`).

pdtwonotes:
The trouble is that requests and responses share nearly all the same issues.  Both have headers and both have the body-delivery problem.   A 'POST' request can have a body just as large as any 'GET'.  That is how WebDav works and how you upload files to internet sites.   The only differences are that responses have a `status` while a request has a `method` and an `url` field.  Everything else is the same.

jemc:
@pdtwonotes - I was more referring to the idea of getting the `respond` method out of the `Payload` class.  I understand that this isn't what you were *really* asking about, but I think it is related in part.  That is, I think that separation can help free you up to use some other patterns for delivering more than one messages for a given response.

pdtwonotes:
The HTTP parser in `PayloadBuilder` handles both message types - I would not want to duplicate that.

Pushing additional body data requires interaction with the ClientConnection or ServerConnection actors which basically represent the "session".  I have been playing with the idea of putting a reference to the relevant Connection actor inside `Payload` so that `Payload.add_chunk` can just pass it through to the connection thus avoiding modifying the `Payload` after all the headers have been filled in.  That way an `iso` `Payload` can be converted to `val` and handed off, while keeping a copy to act as a pathway to the session Connection.  Body data is then delivered separately from the rest of the Payload.

On the other hand, taking the `body` out of the `Payload` object entirely and always managing it separately might solve the problem.  It is, after all, just a bytestream as far as the `net/http` package is concerned.  Having the Connection keep track of them separately would be easier if it was not for the current pipelined request support, where multiple requests and their responses can be floating about within the same session.  My own preference is to remove pipelining entirely, but so far I am keeping it in with a way to disable it at the `Server` level.  Pipelining only makes sense if you can *guarantee* that all requests and responses will be small.

I have not even started to worry about Trailer fields, which are like header fields but they occur *after* the body.  They are possible in `chunked` transfers.

SeanTAllen:
@pdtwonotes i think it would make sense for this RFC for you to implement as a library outside of the standard library and then present that as a replacement for existing http.

pdtwonotes:
@SeanTAllen I agree with you.  I have had to write a lot of code to see how all the parts need to fit together anyway, with several false starts.

pdtwonotes:
I have written the replacement library as suggested, with examples for GET, PUT and server.   It is available for viewing [here](https://github.com/pdtwonotes/httpng).  The examples do file fetching and upload, and the server supports read and write file access with very simple authentication.  I have done succesful streaming to an MP3 player and the backpressure works as expected.

sylvanc:
This looks good. Should we put it to a vote?

pdtwonotes:
I should mention that the absolute latest version of my http library is [here](https://github.com/pdtwonotes/tokara/tree/master/packages/http).  I doubt it differ much from the previous ones.   And the full web server named [tokara](https://github.com/pdtwonotes/tokara) is considerably more featureful than the example `httpserver` I had before.  I don't know if you want an "example" to be that complicated -it has a CGI interface and user authentication.  But there are som bug-fixes in the basic file serving part I need to backport.

jemc:
If it's still in progress, seems like we should hold back on the vote then.

SeanTAllen:
I think it would make sense to let it get some use for a while and fix any problems outside the standard library as it gets used and when its "ready" bring it in. That said, the existing one is highly problematic so, replacing it sooner also has merits.

sylvanc:
@pdtwonotes This is looking really good. How are you feeling about stability of the http package API? If you think it's stable, shall we merge the RFC? Or if you think it's still evolving, we can leave this unmerged until you're ready.

pdtwonotes:
I think the _interface_ is stable.  There may be bugs inside of course.

pdtwonotes:
I've made some fixes since the last update to this PR.  Is it merged into the RFC repo or the ponyc 
repo?

jemc:
@pdtwonotes only the RFC itself is merged.  It would be up to you (or someone else, I suppose) to file a PR to `ponyc` that includes the code changes you've been working on.

PR #72, opened 2016-12-25
Packed structures

Praetonus:


SeanTAllen:
I'm hesitant on the "how we test this".

I think it makes sense to have a couple tests that exercise the packed structures code to make sure its working. Without such a test someone in the future might accidentally break packed structures and we wouldn't know.

Other question, where in the tutorial should packed structures go? What section? "Calling C from Pony"?

Praetonus:
> I think it makes sense to have a couple tests that exercise the packed structures code to make sure its working. Without such a test someone in the future might accidentally break packed structures and we wouldn't know.

Yeah, I agree.

> Other question, where in the tutorial should packed structures go? What section? "Calling C from Pony"?

It looks like plain structures aren't currently covered in the tutorial. We could add the explanation for structures in "Calling C from Pony" and cover packed structures in the same pass.

I'll amend the RFC.

Praetonus:
Updated.

Praetonus:
@SeanTAllen I'm going to put this into ready for vote, if you're okay with the updated RFC.

SeanTAllen:
I'd like to see more detail about what type of tests we should have. This RFC is fine is you implement it but if someone else was to try and work on it, the lack of specific around what should be tested isn't great. 

I for example am not sure what/how we would test this.

Praetonus:
I've added that.

sylvanc:
Passed the vote, merging :)

PR #73, opened 2016-12-26
Delegate removal

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/remove-delegate/text/0000-remove-delegate.md).

This RFC stems from a shared feeling about delegates among core team members. If you'd like to keep delegates in the language, please speak up.

### DIFF
--- a/text/0000-remove-delegate.md
+++ b/text/0000-remove-delegate.md
@@ -1,0 +1,28 @@
+- Feature Name: remove-delegate
+- Start Date: 2016-12-26
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This RFC proposes the removal of delegates from the language features.
+
+# Motivation
+
+Delegates are rarely used in real-world code (for example, this feature isn't used at all in the current standard library) and are often confusing for new users.
+
+# Detailed design
+
+We would remove the implementation and tests for delegates from the compiler and the explanation and examples from the tutorial.
+
+# How We Teach This
+
+An email on the mailing list could be sent to inform everybody of the feature removal.
+
+# Drawbacks
+
+This is a breaking change (although it probably won't have a huge impact on existing code).
+
+# Unresolved questions
+
+None.


SeanTAllen:
Looking forward to voting on this!

ryanai3:
Sorry if I'm late to comment on this but I wanted to say I really think we should keep delegates in the language in some form or another, for a couple of reasons.

Mainly, because of the lack of other language constructs to help with code-sharing when code needs to handle concrete members of an object or actor (i.e. not functions). We could accomplish this through some automatic getter-setter creation on traits as in [here](https://github.com/ponylang/ponyc/issues/282). 

Second, it allows for much nicer frameworks/libraries in concert with other language features. With embed actors, you can imagine writing a custom actor in some pub-sub framework is as easy as: 
```pony
actor CustomActor[A] is Publisher[A]
  embed _subscriber_handler: SubscriberStore[A] delegate Publisher[A] = HashSubscriberStore[A]
```
Want to use a different subscriber store? it's as easy as switching it out:
```pony
actor CustomActor[A] is Publisher[A]
  embed _subscriber_handler: SubscriberStore[A] delegate Publisher[A] = TreeSubscriberStore[A]
```
without delegates, doing something like this requires writing a forwarding method for every method in the trait you're using the delegate to implement.
I understand that delegates may add some complexity to the compiler, but given that they're basically equivalent to adding a bunch of forwarding methods for the relevant trait, I don't know if that should be the case.

If we're worried about new users/ the level of confusion around delegates, I would propose simply changing the name to something that's more clear - why don't we make the keyword "forward", since it's basically just forwarding methods?

I'm worried that if we remove delegates, we'll find ourselves in a position where it becomes very difficult to share code, and we'll miss out on a host of convenient and concise constructs that libraries and frameworks can use.

SeanTAllen:
@ryanai3 you can do all that without delegates, with traits without having to forward anything.

ryanai3:
@SeanTAllen Not if the methods on the object/actor you're delegating to touch some internal fields. 
I.e. code sharing w/ traits only works insofar as the implementation doesn't touch fields.

jemc:
@ryanai3 Have you ever used delegates in a real project before now?  Every time I've thought it was the right tool and tried to do so, some detail of the circumstances made them broken and impossible to use in that situation.  I approve of delegates in theory, but they've never successfully solved a problem for me in practice.

SeanTAllen:
@ryanai3 You can make private accessor fields available via Traits and access internal fields. I plan on writing a Pony Pattern about it. 

```
trait Foo
  fun touches_field() =>
    _get_field()

  fun _get_field(): Field

class Bar is Foo
  let _x: Field = Field

  fun _get_field(): Field =>
    _x
```

PR #74, opened 2017-01-02
Lazy Iterators

Theodus:
[Rendered](https://github.com/Theodus/rfcs/blob/lazy-iterators/text/0000-lazy-iterators.md)

### DIFF
--- a/text/0000-lazy-iterators.md
+++ b/text/0000-lazy-iterators.md
@@ -1,0 +1,312 @@
+- Feature Name: lazy-iterators
+- Start Date: 2017-01-02
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+This RFC proposes the extension of the current Iterator interface so that it provides the functions of the itertools/Iter class. This will allow for a single point of implementation for data transformations that make use of lazy evaluation for improved performance of multiple transformations.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


jemc:
I'm in general pretty opposed to this change.

In my view (and currently accepted Pony convention), `builtin` should be kept as lean as possible, implementing the bare minimum, as defined by these rules from https://github.com/ponylang/ponyc/issues/286:
> 1. Types the compiler needs to know exist, such as None.
> 2. Types with "magic" internal workings that must be supplied directly by the compiler, such as U32.
> 3. Any types needed by others in builtin.

This change doesn't meet any of those requirements, and I'm not convinced that moving these operations from the `itertools` package to `builtin` provides any significant value, since they can all be implemented in terms of `next` and `has_next`.  Any small convenience value it does provide is not worth the increased maintenance complexity/entanglement of `builtin` and the longer compile times that will result, in my view.

jemc:
Also, I think the title of this RFC is a bit misleading - the `next` and `has_next` methods exposed by the current `Iterator` interface are neither intrinsicly eager nor lazy - they are flexible enough to be used in either way.  Moving these functions from the `itertools` package to `builtin` does not make new lazy uses of iterators possible.

Theodus:
@jemc I agree that this change makes little sense considering those rules for builtin, especially when considering the impact on compile times. Though I feel that those rules should be clearly expressed in the documentation of the builtin package rather than a closed issue if they are going to define what the builtin package contains. Should I open a PR on ponyc to do so?

jemc:
@Theodus - I think that's a good idea.

Theodus:
PR opened: ponylang/ponyc#1495

PR #78, opened 2017-02-15
Added RFC for network errors handling

zevlg:
What do you think guys?

### DIFF
--- a/text/0000-network-errors-handling.md
+++ b/text/0000-network-errors-handling.md
@@ -42,9 +42,9 @@ essential for production software.
 * `SocketError` - other socket error not listed above
 
 2. Add private `_errno` field of `SocketErrNo` type into
-   `TCPConnection`, `UDPSocket` and `TCPListener`.  On failure, set
-   `_errno` using corresponding OS errno code fetched using `SO_ERROR`
-   socket option
+   `TCPConnection`, `UDPSocket` and `TCPListener`.  On connection
+   outcome, either success or failure, `_errno` must be set using
+   corresponding OS errno code fetched using `SO_ERROR` socket option
 
 3. Create publically available method `errno` which returns `_errno`,
    for example:
--- a/text/0000-network-errors-handling.md
+++ b/text/0000-network-errors-handling.md
@@ -39,6 +39,7 @@ essential for production software.
 * `SocketConnRefused` - no-one listening on the remote address
 * `SocketNetUnreach` - no routing to remote host
 * `SocketTimeout` - timeout while attempting connection
+* `SocketError` - other socket error not listed above
 
 2. Add private `_errno` field of `SocketErrNo` type into
    `TCPConnection`, `UDPSocket` and `TCPListener`.  On failure, set
--- a/text/0000-network-errors-handling.md
+++ b/text/0000-network-errors-handling.md
@@ -34,7 +34,7 @@ essential for production software.
 
    Where:
 
-* `SocketOK' - success
+* `SocketOK` - success
 * `SocketInUse` - local address is already in use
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I'm not in favor of new methods. I'd rather use existing failure methods that already exist. We are pre-1.0, this is a small change, backwards compatibility isn't a concern.

SeanTAllen:
The "How we teach this" depends on functionality that doesn't exist... deprecation annotation. So, I don't consider that to be a good answer.

I think it would be reasonable to update the doc strings for the given methods to account for the methods as well as add example code (or expand as the case maybe) for TCPConnection, UDPSocket and TCPListener to show usage of handling errors.

SeanTAllen:
Can you detail all the different SocketErrors we need to account for in the union type and what they would map to? There isn't enough here to implement. A complete mapping that is agreed upon would allow anyone to pick up and implement this.

zevlg:
I've added few modifications, favoring Design to not change existing API

text/0000-network-errors-handling.md
```diff
@@ -12,91 +12,73 @@ Provide API for handling network errors.
...
-    """
-    Called if it wasn't possible to bind the socket to an address.
-    Default implementation calls
-    """
-    not_listening(listen)
-  ...
+2. Add private `_errno` field of `SocketErrNo` type into
+   `TCPConnection`, `UDPSocket` and `TCPListener`.  On failure, set
+   `_errno` using corresponding OS errno code fetched using `SO_ERROR`
+   socket option
```
jemc: On success, we'd also want to set it back to `SocketOK`, right?

zevlg: Yeah, on any outcome (success or failure) we want to set `_errno`. I'll fix

Praetonus:
Why not have the error reason be a parameter of `connect_failed` and `not_listening`, as mentioned in Alternatives? I don't really see how that would be "less smooth" and there are several advantages. I think having the error code as a parameter is less error prone because the method signature reminds you to check the reason, and it avoids setting a field on every operation in the implementation.

SeanTAllen:
the markdown formatting on this is off and currently kind of hard to read as half of it is in italics

SeanTAllen:
I'm not in favor of this RFC and the errno value. I'd want to see error values passed to notify methods as appropriate.

zevlg:
Changing notifiers API is of course less smooth because it breaks existing code which uses "net" package. If keeping existing API is not important, then success and failure callbacks can be joined to single callback with error code provided

Praetonus:
As @SeanTAllen said, ensuring backwards compatibility isn't very important until we've hit 1.0. That said, I think we'd want to keep the success and failure callbacks separate since they are expected to take very different actions.

jemc:
I'd tend to agree with @Praetonus about keeping the callbacks separate - they are really very separate in terms how you want to handle them, and arguably the success case doesn't really need an error code.

zevlg:
Having single callback for connection result will also solve [RFC 23](https://github.com/ponylang/rfcs/blob/master/text/0023-network-dont-provide-default-implementation-for-failures.md), since it will force "net" package users to implement dispatch on error code

zevlg:
I actually like different callbacks as well for success&failure and also error provided in callback, that was the initial proposal in given RFC.  So having API like

```pony
  fun ref not_listening(sock: UDPSocket ref, error: SocketError)

  fun ref not_listening(listen: TCPListener ref, error: SocketError)

  fun ref connect_failed(conn: TCPConnection ref, error: SocketError)
```

is most affordable, correct ?

SeanTAllen:
@zevlg yes. i think providing the error to existing methods is a the best approach

PR #80, opened 2017-02-25
Initial CLI format RFC draft.

cquinn:
My proposal for a standard CLI format and corresponding library package. This addresses https://github.com/ponylang/ponyc/issues/471

I have an initial implementation which I will submit a PR for shortly.

### DIFF
--- a/text/0000-cli-format.md
+++ b/text/0000-cli-format.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-Provide a standard command line argument syntax for Pony tools and other programs, and a corresponding package to be provided in the standard library.
+Provide a standard command line argument syntax for Pony tools and other programs, and a corresponding package named `cli` to be provided in the standard library.
 
 
 # Motivation
@@ -19,7 +19,7 @@ Having a standard syntax for command line arguments and options will make Pony c
 
 Command lines are broken up into a series of tokens before they are delivered to a program as an array of strings. There are a number of ways to treat those tokens, but most Unix-like environments have settled on a syntax that combines commands, arguments and options. The GNU C library doc is a good reference for how the options should be handled:
 
-- http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
+ - http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
 
 Options are usually enough for simple programs, but more often it is desirable to have a richer grammar with library support for parsing sub-commands and positional arguments in addition to the options. It's nice not to have to rewrite that logic every time, and not to force users to learn a different command line syntax for every program they use.
 
@@ -94,7 +94,7 @@ Option may be specified as being required. Such options not being present is an 
 
 ## Library
 
-The types in the CLI package are broken down into three groups:
+The types in the `cli` package are broken down into three groups:
 
   - The spec classes and primitives which are used to declare a CLI grammar.
   - The parser class which is used to parse a command line against a spec.
@@ -138,7 +138,7 @@ A complete unit test suite should be written to validate the implementation of t
 
 # Drawbacks
 
-This complete CLI library may be overkill for small tools.
+None known.
 
 
 # Alternatives
--- a/text/0000-cli-format.md
+++ b/text/0000-cli-format.md
@@ -1,0 +1,91 @@
+- Feature Name: cli-syntax
+- Start Date: 2016-10-24
+- RFC PR:
+- Pony Issue: https://github.com/ponylang/ponyc/issues/471
+
+# Summary
+
+Provide a standard command line argument syntax for Pony tools and other programs, and a corresponding package to be provided in the standard library.
+
+
+# Motivation
+
+Having a standard syntax for command line arguments and options will make  Pony command line programs more consistent and thus easier to learn and use by users. Having a standard library package that provides this implementation will make writing conformant programs easier for Pony programmers.
+
+
+# Detailed design
+
+## Grammar 
+
+Command lines are broken up into a series of tokens before they are delivered to a program as an array of strings. There are a number of ways to treat those tokens, but most Unix-like environments have settled on a syntax that combines commands, arguments and flags. The GNU C library doc is a good reference for how the flags or options should be handled:
+
+- http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
+
+Flags are usually enough for simple programs, but more often it is desirable to have consistent grammar and library support for parsing sub-commands and positional arguments in addition to the flags. It's nice not to have to rewrite that logic in every program, and not to force users to learn different command line syntax for every program they use.
+
+In the Golang world, the Flags in the standard library are fairly weak and not posix compliant, so a couple of packages have become very popular extensions. The most interesting one is Kingpin, which presents a nice interface to the programmer and a consistent grammar to the program user:
+
+ - https://github.com/alecthomas/kingpin
+
+The command line is treated as a command with optional flags and positional arguments. The command is composed of the root command which represents the program, plus zero or more subcommand tokens. Each leaf command may specify required and optional arguments. And, each command may specify required or optional flags which add to those inherited from its parent.
+
+The general EBNF of a command line looks like:
+```
+  command_line ::= root_command (flag* command*)* (flag | arg)*
+  command ::= alphanum_word
+  alphanum_word ::= alphachar(alphachar | numchar | '_' | '-')*
+  flag ::= longflag | shortflagset
+  longflag ::= '--'alphanum_word['='arg | ' 'arg]
+  shortflagset := '-'alphachar[alphachar]...['='arg | ' 'arg]
+  arg := boolarg | intarg | floatarg | stringarg
+  boolarg := 'true' | 'false'
+  intarg := ['-'] numchar...
+  floatarg ::= ['-'] numchar... ['.' numchar...]
+  stringarg ::= anychar
+```
+
+Some Examples:
+```
+  tool action --switch1 --option2=42 sub_action -o 9 arg1 arg2
+  chat --channel=main say hello
+  chat handle fred --email=fred@com.com
+```
+
+## Library 
+
+The types in the CLI package are broken down into three groups:
+
+  - The spec classes and primitives which are used to declare a CLI grammar.
+  - The parser class which is used to parse a command line against a spec.
+  - The resulting parsed command, flag and arg classes.
+
+Pony programs use a builder pattern to create the spec for their command line syntax. This spec is checked for correctness at compile time, and represents everything the library needs to know when parsing a command line or displaying syntax help messages.
+
+Programs then use this spec at runtime to parse any given command line. This is often `env.args()`, but could also be commands from files or other input sources. The result of a parse is either a parsed command, or a syntax error.
+
+Programs then query the parsed command to determine the (sub) command specified, and the effective values for all flags and arguments.
+
+
+# How We Teach This
+
+The CLI package would have reference doc for all of the types like any other library package. We would also update existing tools to use this grammar and library when possible to serve as examples. Other types of doc should be written such as patterns and examples.
+
+
+# How We Test This
+
+A complete unit test suite should be written to validate the implementation of this RFC.
+
+
+# Drawbacks
+
+This complete CLI library may be overkill for small tools.
+
+
+# Alternatives
+
+A simpler flag parser.
+
+
+# Unresolved questions
+
+Can parsing of environment variables be included?
--- a/text/0000-cli-format.md
+++ b/text/0000-cli-format.md
@@ -15,7 +15,7 @@ Having a standard syntax for command line arguments and options will make  Pony 
 
 # Detailed design
 
-## Grammar 
+## Grammar
 
 Command lines are broken up into a series of tokens before they are delivered to a program as an array of strings. There are a number of ways to treat those tokens, but most Unix-like environments have settled on a syntax that combines commands, arguments and flags. The GNU C library doc is a good reference for how the flags or options should be handled:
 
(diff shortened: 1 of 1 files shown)


cquinn:
[Rendered](https://github.com/cquinn/rfcs/blob/master/text/0000-cli-format.md)

cquinn:
A first cut at an implementation is here: https://github.com/cquinn/ponycli

sylvanc:
I like this, particularly the nested commands. What would you think of having the methods on `Flag` instead be passed as optional args to `flag`? So instead of:

```pony
  fun bools_cli_spec(): CommandSpec box =>
    """
    Builds and returns the spec for a CLI with four bool flags.
    """
    let cs = CommandSpec("bools", "a sample CLI with four bool flags")
    cs.flag("a", BoolType).short('a')
    cs.flag("b", BoolType).short('b')
    cs.flag("c", BoolType).short('c')
    cs.flag("d", BoolType).short('d')
    cs
```

Have:

```pony
  fun bools_cli_spec(): CommandSpec box =>
    """
    Builds and returns the spec for a CLI with four bool flags.
    """
    CommandSpec("bools", "a sample CLI with four bool flags")
      .flag("a", BoolType where short = 'a')
      .flag("b", BoolType where short = 'b')
      .flag("c", BoolType where short = 'c')
      .flag("d", BoolType where short = 'd')
```

The idea being that, once a `Flag` is constructed, it should logically be immutable.

Also, sorry for bikeshedding - it's also fine with the current API :)

sylvanc:
Or, if the `CommandSpec` is though of in the same way:

```pony
  fun bools_cli_spec(): CommandSpec box =>
    """
    Builds and returns the spec for a CLI with four bool flags.
    """
    CommandSpec("bools", "a sample CLI with four bool flags",
      [ Flag("a", BoolType where short = 'a')
        Flag("b", BoolType where short = 'b')
        Flag("c", BoolType where short = 'c')
        Flag("d", BoolType where short = 'd')
      ])
```

Also, I will now stop with the bikeshedding :)

SeanTAllen:
This generally seems good. Before this is accepted, I would want to do a review of anything in the `ponyc` repo like programs in `examples` and explicitly list in the RFC "how we teach this" section, what needs to be updated.

Further on the "how we teach this", I think that this should either be included in the tutorial, in pony patterns or we should have a high level "guide to the standard library". the existing docs at http://www.ponylang.org/ponyc/ are moderately useful but lack any guidance in finding what you need.

cquinn:
Please feel free to bikeshed: it's clear there are Pony idioms that I missed in my API. 

In this case, while I always try to make objects as immutable as possible, I think I had given up too soon with the Specs. You can see that the resulting Command objects and friends are all immutable, but I forgot about named arguments in pony. I'll make a pass at updating them and see how they look.

In your example Sylvan, are those Flag's being passed as an array? No comma separators needed?

cquinn:
I will review the `packages` and `examples` to see what I would recommend updating. Do you think this package would replace or supersede the current Options class?

I agree with adding tutorial and patterns entries for this.

SeanTAllen:
I think we would want to deprecate Options. No one is a fan of it.

sylvanc:
I think we could even pull the `options` package out of the stdlib.

@cquinn yes, it's an array, but I'm afraid it should have commas. I've been fiddling with an alternate, expression based, array syntax, and I used it there without thinking! Basically, it turns arrays from:

```pony
[a, b, c]
[ a,
  b,
  c
]
```

To:

```pony
[a; b; c]

[ a
  b
  c
]
```

cquinn:
I played around a bit with using named arguments and call-site chaining, and came up with this:
```  
CommandSpec("bools", "a sample CLI with four bool flags")
      .>flag("a", BoolType where short = 'a')
      .>flag("b", BoolType where short = 'b')
      .>flag("c", BoolType where short = 'c')
      .>flag("d", BoolType where short = 'd')
```
Which allows full immutable FlagSpecs and ArgSpecs, and is still nice and clear. I also tried a quick approach where constructed FlagSpecs where then added to the command one by one, but that was verbose and awkward. I'll try the arrays approach: I should be able to add a mix of FlagSpecs, ArgSpecs and sub-CommandSpecs.

cquinn:
I do like the Arrays approach. The code to define specs ends up looking quite nice, allowing all-at-once construction of a tree as a single expression.
```
    CommandSpec.parent("chat", "sample chat program", [
      FlagSpec("admin", BoolType, "chat as admin" where default'=false),
      FlagSpec("name", StringType, "your name" where short'='n'),
      FlagSpec("volume", F64Type, "chat volume" where short'='v')
    ],[
      CommandSpec.leaf("say", "say something", Array[FlagSpec](), [
        ArgSpec("words", StringType)
      ]),
      CommandSpec.leaf("emote", "", [
        FlagSpec("speed", F64Type, "how fast to play emotion" where default'=F64(1.0))
      ],[
        ArgSpec("emotion", StringType, "emotion to send")
      ])
    ])
```
I will still allow subcommands to be added to a parent CommandSpec, to allow for the cases where subcommands are declared and implemented independently and added to a root/wrapper command host.

Another idea I had was to use named constructors to specify the type of FlagSpecs and ArgSpecs. So instead of:
```FlagSpec("name", StringType, "your name" where short'='n'),```
You'd have:
```FlagSpec.string("name", "your name" where short'='n'),```
Not sure about that, though.

**Comment by jemc** [comment]
Using named constructors would allow the parameters to be more specific, when it comes to accepting a default value of the same type.

That is, if you currently allow this:

```pony
FlagSpec("speed", F64Type where default'=F64(1.0))
FlagSpec("emotion", StringType where default'="happy")
```

Then it's not possible to disallow the following at compile time (you could only do it with an error at runtime, which isn't ideal):

```pony
FlagSpec("speed", F64Type where default'="happy")
FlagSpec("emotion", StringType where default'=F64(1.0))
```

lisael:
I like this 👍 

Maybe we could use a generic type, this would solve @jemc's objections and it would be much cleaner than F64Type, BoolType (in my opinion... I find type markers ugly, and I try to avoid them wherever I can)

something like:

```pony
FlagSpec[F64]("speed" where default'=1.0)  // I'm not sure if F64(1.0) is inferred yet
FlagSpec[String]("emotion" where default'="happy")
```

It looks a bit more like `options`, though I don't know if it's a good or a bad thing :)

lisael:
The array-based tree structure is nice @cquinn

And +1 for the mutable CommandSpec, even if static nature of pony makes it hard to add modules. This feature has limited use-cases. I see:
  * external modules as servers run in separate processes as sometimes done in Go
  * implement and release a re-usable library that exposes CLI arguments to add to Main (??)

BTW, it's a nice-to-have

jemc:
@lisael - I agree that using type parameters as in your example would be the most ideal solution, though I don't think it's possible to implement it this way at the moment, due to RFC #62 not having been implemented yet.

However, I'd love for someone to prove me wrong!

cquinn:
@lisael thanks. I have pushed an update to my branch that uses named constructors for the arg & flag specs. I think it reads better, but I will also try generics. I like immutable for most things, but I  have found that it's nice having some flexibility to have the subcommands implemented independently and then merge them all together at the top level.

SeanTAllen:
I'm really liking this. @cquinn do you feel like you are reaching a final comment period?

If I'm not around when this comes up for a vote, I think an interesting issue arises from this.
What do we do with Options?

I think deprecating it makes sense.

As we are pre-1.0, I think "deprecation" should mean that we remove it from ponyc without leaving it hanging around but, we create a new repo for just Options as a library so that if people want to continue using it, they still have access to it and if they think its a better approach, can always fork and take over maintenance.

As part of this, we'd need to make clear that the Pony org hosted package is in maintenance mode and as such isn't guaranteed to work going forward, won't receive bug fixes etc (if you want any changes, fork it, take over as maintainer).

@ponylang/committer thoughts? ^^^

jemc:
@SeanTAllen I think we should hold off on final comment period for a bit, since @cquinn is still in the process of trying alternate approaches for various details of the API.  I want to make sure we don't rush to the finish line when there are still a lot of ideas at play.

cquinn:
I think I have the approach and base features defined in the RFC now. The library is almost there—I think I just have left the generation of the user syntax message.

How do we want this process to work? If we all like the RFC but still want to iterate on the code a little, should we ratify the RFC and move the code discussion over to PRs in the main repo?

I'm sure there will also be features we decide we want at a later time: how will those additions be ratified later on?

Finally, we should have a name for this package. Probably not options2 :). How about flags, or cli?

jemc:
@cquinn - unless you wanted to do a separate RFC for the library API details, I'd actually like to get those details copied into this RFC before we move forward it.

That is, the RFC mainly talks about the command line formats accepted, but doesn't discuss the API, and if merging this PR means accepting both those formats and the library API, I'd like to have the API details added to the RFC before continuing.  Internal implementation details could be saved for the PR discussion, but I think the API design need to be part of the RFC.

> I think I just have left the generation of the user syntax message.

If you like, you could do a followup RFC on that part of the API if you wanted to finish this RFC first and add that feature later.

> Finally, we should have a name for this package. Probably not options2 :). How about flags, or cli?

If we want to pick a new name, I think `cli` would be most appropriate - `flags` would be too confusing with the `Flags` type in the `collections` package.

However, personally, I think I'd prefer to get rid of the current `options` package entirely and make this the new `options` package.  Since this seems to be a replacement for it in terms of features, I don't see a reason for them both to exist in the same standard library.

jemc:
@SeanTAllen as I mentioned in my comment just now, I personally think we should just get rid of the current `options` package entirely.

If somebody decided for some reason they wanted to take over maintenance for the old `options` package because they like the old version so much better, it would still exist in the `git` history, so they could still get to it.

I think it would be too confusing for the `ponylang` org on github to host deprecated packages we don't actually want anyone to use and aren't going to maintain.  Especially while we are in pre-1.0 mode here.

SeanTAllen:
@jemc if that is the case then i think as part of release notes when it is deprecated, we need examples on how to upgrade code from options to its replacement. we might be pre-1.0 but its still potentially a rather large change that could impact a decent chunk of pony code that is out there.

cquinn:
I think the implementation is functionally complete now. I'm sure it will need some polish, but I think it does what it needs to do for a first cut, and we can iterate after we start using it more.

The questions I see remaining for this RFC PR:

- Should this RFC just focus on the command line format as per #471, and draft a new RFC for the library? Or should we bundle both into this RFC?
- Name for the package: 'cli' or 'options'?
- If we go with 'options' maybe we should use the term 'option' instead of 'flag'.
- I was also considering merging Args and Flags, since what it really comes down to is that Flags are just named arguments, and Args are positional arguments. Not sure if that would be simpler or confusing to users.

jemc:
> Should this RFC just focus on the command line format as per #471, and draft a new RFC for the library? Or should we bundle both into this RFC?

I would tend towards bundling both here, but either would be okay, I think.

> If we go with 'options' maybe we should use the term 'option' instead of 'flag'.

Sounds reasonable to me.

> I was also considering merging Args and Flags, since what it really comes down to is that Flags are just named arguments, and Args are positional arguments. Not sure if that would be simpler or confusing to users.

Honestly, it seems more confusing to merge them - I think it's better to have separate names for those concepts.

killerswan:
I like this RFC, @cquinn.

(1)
I strongly disagree about the listed drawback about the complete library being overkill for small tools. ( Small tools become big tools, and doing a CLI wrong to start has a cost later.)  You should take that off the RFC, it's not a drawback at all!  🔥 

(2)
And I love the unresolved question.  What if we extended that EBNF grammar (and our thinking) to cover envvar parsing, too?
```
command_line ::= root_command (option | command)* (option | arg)* (expected_envvar)*
command ::= alphanum_word
alphanum_word ::= alphachar(alphachar | numchar | '_' | '-')*
option ::= longoption | shortoptionset
longoption ::= '--'alphanum_word['='arg | ' 'arg]
shortoptionset := '-'alphachar[alphachar]...['='arg | ' 'arg]
arg := boolarg | intarg | floatarg | stringarg
boolarg := 'true' | 'false'
intarg := ['-'] numchar...
floatarg ::= ['-'] numchar... ['.' numchar...]
stringarg ::= anychar
expected_envvar ::= alphanum_word'='arg
```

Perhaps this introduces ambiguity between envvars specified in the environment, and envvars specified in a command line call.  I'm not sure that bothers me.  But also, not too much harder to make the automated generation of `--help` distinguish between such things and list expected envvars.

Speaking of which, that default `-v`, `--version`, `-h`, `--help` are missing from the RFC.

killerswan:
Also, are deprecation warnings implemented (or in some other RFC) yet?

jemc:
> Also, are deprecation warnings implemented (or in some other RFC) yet?

No, we don't have a general system for issuing deprecation warnings at this time.

SeanTAllen:
Should this go into final comment period?

SeanTAllen:
My general statement. This seems like a definite improvement over the current `Options` package. I'm in favor of anything that's a definite improvement. I do think a drawback should be listed for breaking code when `Options` is removed. And that removing `Options` should be part of the RFC.

jemc:
@sylvanc agreed that removing the `options` package is fine, but suggests calling this package `cli` might be better than calling it `options`.

cquinn:
I agree that `cli` is a better package name, and will make it less confusing as we migrate from `options`.

I would like to wrap up this RFC and have it approved with the CLI spec and the package requirements as it is, and then write a small follow-up RFC that is the package concrete design.

jemc:
@cquinn - sounds great - we should be able to vote this in as soon as the 7-day final comment period is finished.

renatoathaydes:
I wrote something a little similar for a Java CLI I wrote, if you want to look at the [tests](https://github.com/renatoathaydes/osgiaas/blob/master/modules/cli/osgiaas-cli-api/src/test/groovy/com/athaydes/osgiaas/cli/args/ArgsSpecSpecification.groovy) for it, you could get some ideas, maybe even something helpfu ;).

I noticed you don't seem to have the concept of mandatory VS optional arguments? How about options that can be specified multiple times?

v217:
Since most cli interfaces are not overly complex, I want to mention a simple solution.
What is your opinion on http://docopt.org/ ?

jemc:
@v217 - good link, thanks for sharing!

Such a system definitely seems quite ergonomic to use, though I wouldn't necessarily call it simpler in terms of implementation, since it involves parsing a documentation string and converting that into a machine CLI spec from there.

It seems like such a tool would be built *on top of* the kind of traditional CLI specification API that we're talking about here.  Though I think it would be quite a useful layer on top, I do see it as being somewhat out of the scope of defining the basic in-language specification API.

So yes, I think it would make quite a nifty package on top.

PR #81, opened 2017-02-25
Add RFC for more-promise-methods

Theodus:
[Rendered](https://github.com/Theodus/rfcs/blob/more-promise-methods/text/0000-more-promise-methods.md)

### DIFF
--- a/text/0000-more-promise-methods.md
+++ b/text/0000-more-promise-methods.md
@@ -1,0 +1,63 @@
+- Feature Name: more-promise-methods
+- Start Date: 2017-02-25
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+This RFC proposes adding additional methods to the Promise actor.
+
+# Motivation
+
+The Promise actor is a useful tool for representing a value before it is available, though the public API is a bit sparse. Only 3 basic behaviors and functions are currently available to the user (apply, reject, and next). The following additional methods will improve composability and reflect common operations found in other promises/futures libraries.
+
+# Detailed design
+
+```pony
+
+new many(ps: Array[Promise[A]] val): Promise[Array[A] val]
+	"""
+	Create a promise that is fulfilled when all promises in the given array are
+	fulfilled. If any promise in the array is rejected then the new promise is
+	also rejected.
+	"""
+
+fun tag join[B: Any #share = A](p: Promise[B]): Promise[(A, B)]
+	"""
+	Join two promises into one promise that returns the result of both when they
+	are fulfilled. If either of the promises is rejected then the new promise is
+	also rejected.
+	"""
+
+fun tag select(p: Promise[A]): Promise[(A, Promise[A])]
+	"""
+	Return a promise that is fulfilled when either promise is fulfilled,
+	resulting in a tuple of its value and the other promise.
+	"""
+
+fun tag timeout(expiration: U64)
+	"""
+	Reject the promise after the given expiration in nanoseconds.
+	"""
+
+```
+
+# How We Teach This
+
+Each method added will contain documentation as shown above.
+
+# How We Test This
+
+Each method will have a unit test in the promises package to ensure that the implementations work as intended.
+
+# Drawbacks
+
+These new methods will increase the maintenance cost of the promises package.
+
+# Alternatives
+
+Other methods may be added in addition or as replacements to the ones described above.
+
+# Unresolved questions
+
+None.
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@Theodus I think documentation is a little lacking. I believe as part of this, package level documentation should be updated to discuss the different ways you can use a Promise. Previously with the "sparse" selection of methods, there were few options for how to use a Promise. This change would create many more options and as such I think greatly increases the need for more expansive and thorough documentation.

I think the minimum is expanded package level docs as I said before. Ideally this would detail basic use cases for different ways of using the expanded Promises API.

As part of thinking about and detailing that documentation for this RFC, I think it might become clear that some Pony Patterns should be added for the various ways of using Promises.

Additionally, we have zero documentation on Promises in the tutorial and that the addition of some content there would be a nice to have to go along with this RFC.

In general, I am in favor of this RFC, however, I think we would be remiss to not but a good amount of work into teaching how to use the Promises.

Theodus:
@SeanTAllen the Pony Tutorial doesn't seem to include content for the standard library packages and I don't think it should. Instead, I think there should be another resource that contains a section about each package of the standard library.

SeanTAllen:
@Theodus I just started a discussion that is something like that on Carl's CLI RFC.

So, each standard library package has documentation... it's the documentation for that package. However, if you don't know that it exists or how it fits in with Pony, its not very useful. The tutorial or something else can help play a role here. 

Perhaps, we just need a couple Pony patterns for using promises, perhaps something else. I'm not sure but we keep running into the same "missing documentation source" issue on RFC after RFC.

sylvanc:
I like these. More asynchronous combinators (and actually, more synchronous ones in other packages) is a good thing.

text/0000-more-promise-methods.md
```diff
@@ -0,0 +1,63 @@
...
+```pony
+
+fun tag join[B: Any #share = A](p: Promise[B]): Promise[(A, B)]
+	"""
+	Join two promises into one promise that returns the result of both when they
+	are fulfilled. If either of the promises is rejected then the new promise is
+	also rejected.
+	"""
+
+fun tag many(ps: Array[Promise[A]] val): Promise[Array[A] val]
```
jemc: I think this should be `new` rather than `fun tag` - using `fun tag` implies that the receiver promise will be part of the result, but I think that makes for an awkward API.  In my opinion, the passed array of promises should be the entire array expected to get the results of.

text/0000-more-promise-methods.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Motivation
+
+The Promise actor is a useful tool for representing a value before it is available, though the public API is a bit sparse. Only 3 basic behaviors and functions are currently available to the user (apply, reject, and next). The following additional methods will improve composability and reflect common operations found in other promises/futures libraries.
+
+# Detailed design
+
+```pony
+
+fun tag join[B: Any #share = A](p: Promise[B]): Promise[(A, B)]
```
jemc: If this method were named `add`, it could take advantage of the `+` operator sugar, which seems nice to me.

Maybe this method could be named `add`, and the `many` method could become `join`?

text/0000-more-promise-methods.md
Theodus: @jemc I agree, but we would need a separate primitive to do so since a constructor cannot specify a return type.

Theodus:
To address @jemc's concerns over the join (previously many) method implying that the receiver is included in the resulting promise, would it be a better idea to have a `Promises` primitive with a join method that takes an Array of Promise[A] and returns a Promise of Array[A]?

jemc:
Based on our conversation at the sync call today, I think @Theodus will agree that this RFC is stabilized enough to go into final comment period.  I'm marking it as such tonight, so it can be voted on at next week's sync call.

SeanTAllen:
I'm in favor of this. I think the Pony Patterns addition in how we teach this is a good start.

lisael:
I like this too.

I was thinking about Futures, recently and I feel like they should be a first class citizen. I mean the never-blocking paradigm of pony inevitably lead to callback hell if you don't use Futures and (or the listener pattern to some point).

What I like about Pony is precisely that I don't have to explicitly implement the actors ( by composing locking types as in Rust or writing the message passing code as in Elixir/Erlang servers). I'd love such a sugar for Future, but I have nothing precise in mind.

Maybe some kind of annotated behaviours (`Foo`) that calls other behaviours (`Bar` and `Baz`) before being executed. `Bar` and `Baz` would garantee that some future members of the actor are done before `Foo`...

Well, nothing precise enough for a RFC.

SeanTAllen:
I vote yes on this.

PR #83, opened 2017-03-09
Add RFC for improved-persistent-map-api

Theodus:
[Rendered](https://github.com/ponylang/rfcs/blob/improved-persistent-map-api/text/0000-improved-persistent-map-api.md)


sylvanc:
Looks good. I particularly like the non-partial update.

jemc (approved):

jemc:
Given that this is a pretty simple change, and already has expressed approval from two core members, should we go ahead and move it to final comment period?

SeanTAllen:
I vote yes on this.

PR #85, opened 2017-03-19
Add a TCPConnectNotify.received times parameter

SeanTAllen:


jemc:
I'm generally favorable toward this, but I do wish we had a way to test it.

That's probably for an RFC of the future though, in which we move a lot of the internal machinery of `TCPConnection` into its own synchronous class that manages reading/writing to/from file descriptors. Such a class could be tested more easily in isolation, and could also be reused anywhere we need to deal with file descriptors in a consistent way.

SeanTAllen:
I tried to do that at one point @jemc and it turned out to be a miserable failure. I want to try again now that I have some more experience but I think that there's some code reuse issues there that we probably need to address. With such a rewrite though, if pony test was expanded to be able to do "spying" it could help some.

Pony test has some of what you'd need from a spy framework (expect to know something was called) but you also want to be able to easily validate a method wasn't called (you can do this with fail()). Its kind of possible if the code is structured well right now, but still messy and its twisting pony test a little. 

I think what we could definitely use for testing is a way to replace FFI/system calls for testing.

PR #86, opened 2017-03-24
Add RFC for exhaustive match.

### LINKED ISSUE #56: Exhaustive match and unreachable cases
This was originally opened as ponyc [issue 60](https://github.com/ponylang/ponyc/issues/60).

> If we can prove a match is exhaustive, we can give an error if an else branch is present, and we can avoid unioning None with the expression type (from an else branch that will never be taken).
> If we can prove that a match case is unreachable, we can give an error.

Specifically:

> As previously discussed, since values are matched by calling eq(), we absolutely cannot determine if a set of matches are exhaustive if values are matched and any types involved provide their own eq().
> However, we can do checks when matching is based only on type. I think this would be well worth doing to aid cases such as switching on an enum.

See original issue for full details.

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/exhaustive-match/text/0000-exhaustive-match.md).

Resolves #56.

### DIFF
--- a/text/0000-exhaustive-match.md
+++ b/text/0000-exhaustive-match.md
@@ -1,0 +1,135 @@
+- Feature Name: exhaustive-match
+- Start Date: 2017-03-23
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The compiler will no longer insert an implicit `else None` clause in `match` blocks where all possible types of the match expression are covered by pattern types of the cases.
+
+# Motivation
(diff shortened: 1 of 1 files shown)


sylvanc:
Beautiful! We may at some point extend this approach with things like this:

https://www.microsoft.com/en-us/research/wp-content/uploads/2016/08/gadtpm-acm.pdf

PR #87, opened 2017-03-26
Add RFC for Serialise.signature method.

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/939d9a558f846c25f954958960c85a37a3aa1bd8/text/0000-serialise-signature.md).

### DIFF
--- a/text/0000-serialise-signature.md
+++ b/text/0000-serialise-signature.md
@@ -1,0 +1,72 @@
+- Feature Name: serialise-signature
+- Start Date: 2017-03-26
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+A `Serialise.signature` method, which returns a unique byte array for the current compiled Pony program.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
totally not content oriented: I notice that the newlines seem to be a little off in terms of wrapping in here.

Praetonus:
I have an idea for deterministic signatures, based on [AST serialisation](https://github.com/ponylang/ponyc/pull/1966). The basic idea would be to hash the data of the serialised AST to build the signature. I suspect there could be some problems with that scheme since the order in which packages and modules are loaded can differ between systems, so a better scheme could be:

- Serialise each module of the program separately
- Hash the serialised modules to produce module signatures
- For each package, sort the module signatures, concatenate them and hash that to produce package signatures
- Repeat the last step with package signatures to produce the program signature

During serialisation, we'd have to handle package IDs in `TK_NOMINAL` nodes carefully since the assigned IDs depend on the loading order.

SeanTAllen:
I like this. @jemc are you ready to move to final comment period?

sylvanc:
I think those are good ideas. It would be nice to, theoretically, be able to have some package `Foo`, and have two separate programs, both of which use `Foo`, be able to exchange serialised types from the `Foo` package. That's a bit tricky, as I think it involves either synchronising type IDs, or being able to build a type ID translation.

jemc:
If @Praetonus is confident in the implementation plan, I'm ready to move forward.

PR #88, opened 2017-03-30
Add RFC: more-persistent-collections

Theodus:
[Rendered](https://github.com/ponylang/rfcs/blob/more-persistent-collections/text/0000-more-persistent-collections.md)

### DIFF
--- a/text/0000-more-persistent-collections.md
+++ b/text/0000-more-persistent-collections.md
@@ -41,67 +41,67 @@ class val HashSet[A: Any #share, H: HashFunction[A] val] is Comparable[HashSet[A
         Check whether the set contains the value.
         """
 
-    fun val add(value: val->A): Set[A]
+    fun val add(value: val->A): HashSet[A, H]
         """
         Return a set with the value added.
         """
 
-    fun val sub(value: val->A): Set[A]
+    fun val sub(value: val->A): HashSet[A, H]
         """
         Return a set with the value removed.
         """
 
-    fun val op_or(that: Set[A]): Set[A]
+    fun val op_or(that: (HashSet[A, H] | Iterator[A])): HashSet[A, H]
         """
         Return a set with the elements of both this and that.
         """
 
-    fun val op_and(that: Set[A]): Set[A]
+    fun val op_and(that: (HashSet[A, H] | Iterator[A])): HashSet[A, H]
         """
         Return a set with the elements that are in both this and that.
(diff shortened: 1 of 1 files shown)


jemc:
The name `Vec` might be problematic, since @lukecheeseman's fork with dependent types was using the name `Vector` to be refer to a fixed-size array-like type in `builtin` (https://github.com/lukecheeseman/ponyta/blob/typevalues/packages/builtin/vector.pony).

Is there another term we could use here?

Theodus:
@jemc I see the possible confusion, but this form of persistent data structure is only referred to as vector everywhere that I know of (Clojure, Scala, Haskell, Javascript, Racket).

jtfmumm:
@Theodus I'm not sure how you were thinking of approaching the vector-or-whatever-it-will-be-called, but my original persistent map implementation was a HAMT.  It exists in the history but also in my Pony functional data repo

Theodus:
The implementation will be similar to the current Clojure implementation, including the tail optimization. Most of the implementation details have been handled in my branch: https://github.com/Theodus/ponyc/tree/persistent-vec

agarman:
Why not implement Vector using Map.pony? Is HAMT better than CHAMP at append?

Theodus:
@agarman the compression used in the persistent map wouldn't make much sense as the vector uses the index as the hash.

Theodus:
Also, the vector has an optimization where lookups and modifications to the tail are O(1) rather than O(log32(n)).

jemc:
@Theodus, yes the term "vector" is definitely quite widely known in many programming languages to have this meaning (all the way down to C++). However, I think the fixed-size meaning comes from mathematics, specifically vector algebra.  Kind of sucks that programming languages widely adopted "vector" to be "resizable array" when mathematical vectors are actually fixed-size.

There's definitely some value in consistency in terminology with other programming languages, to be sure, though Pony seems also to have an affinity for terminology from mathematics (like "partial functions", for example), so I'm not sure how best to balance it.  Would be interested to hear others weigh in on this.

agarman:
@jemc Vector is an unbounded random access sequential collection. There's tuple and arrays for random access collections where size is fixed at time of definition or construction.

Theodus:
Should set methods with `that: Set[A]` accept an `Iterator[A]` or should we include union, intersect, etc. from the mutable set API?

PR #90, opened 2017-05-18
Add RFC for string concatenation syntax.

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/bee61dbb40e9db5c60ebd4b29517775c73a587b8/text/0000-string-concatenation.md).

### DIFF
--- a/text/0000-string-concatenation.md
+++ b/text/0000-string-concatenation.md
@@ -1,0 +1,97 @@
+- Feature Name: string-concatenation
+- Start Date: 2017-05-17
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+A new syntax for concatenating `String`s (and `Stringable`s) that generates efficient code for creating the final `String`.
+
+# Motivation
+
+When writing code, what often comes most naturally (and succinctly) is to concatenate `String`s together using the `add` method sugar, as in this modified example from the standard library:
+
+```pony
+primitive DefaultLogFormatter is LogFormatter
+  fun apply(msg: String, loc: SourceLoc): String =>
+    loc.file() + ":" + loc.line().string() + ":" + loc.pos().string() + ": " + msg
+```
+
+However, this approach is not very performant, for the reasons explained in the Pony Pattern named ["Limiting String Allocations"](https://github.com/ponylang/pony-patterns/blob/master/performance/limiting-string-allocations.md). For those reasons, the *actual* standard library example looks like this:
+
+```pony
+primitive DefaultLogFormatter is LogFormatter
+  fun apply(msg: String, loc: SourceLoc): String =>
+    let file_name: String = loc.file()
+    let file_linenum: String  = loc.line().string()
+    let file_linepos: String  = loc.pos().string()
+
+    let output = recover String(file_name.size()
+    + file_linenum.size()
+    + file_linepos.size()
+    + msg.size()
+    + 4) end
+
+    output.append(file_name)
+    output.append(":")
+    output.append(file_linenum)
+    output.append(":")
+    output.append(file_linepos)
+    output.append(": ")
+    output.append(msg)
+    output
+```
+
+This pattern is far more verbose and cumbersome, so will only likely be used when users are *very* concerned about performance, rather than being used by default as the most common way of concatenating strings.
+
+Part of the appeal of Pony is that often the most natural way to do something in the language is also very performant. This is not currently true of string concatenation.
+
+# Detailed design
+
+A new symbol will be added to Pony syntax that represents string concatenation, allowing use like the following:
+
+```pony
+primitive DefaultLogFormatter is LogFormatter
+  fun apply(msg: String, loc: SourceLoc): String =>
+    loc.file() <> ":" <> loc.line() <> ":" <> loc.pos() <> ": " <> msg
+```
+
+This syntax would translate to the same code as the long-form example above, where any non-`String` operands are converted to `String`s via their respective `string` methods, the total byte size of all operands is calculated to allocate an appropriately-sized `String` for the final buffer, and all operands are appended to the final buffer.
+
+The compiler may even be able to do some additional optimization, such as using `push` calls instead of `append` for very short string literals in a concatenation.
+
+All operands in a string concatenation must be a `String`, or be a type that has a `string` method that takes no arguments. The `String` for each operand (or the return type of their `string` method) must be readable (that is, it must be a subtype of `String box`).
+
+The `string` method will not be called more than once for each operand, and it will not be called at all if the operand is already a `String`.
+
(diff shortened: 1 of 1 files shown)


Praetonus:
I like this RFC.

I think a precision about character literals should be added, specifying that they should be treated in the same way as string literals, e.g. with `"abc" <> 'd'` being sugared to `"abcd"` instead of `String.>append("abc").>append(U8(100).string())`. This is important because it will require some modifications to the lexer in order to make character literals visible until the sugar pass instead of producing numeric tokens.

jemc:
@Praetonus - I thought about this, but given that I mentioned that string literals with a small number of characters can be optimized to use `push` calls instead of `append`, is there any reason to use a character literal in a string concatenation?

Note that under the rules I wrote in the RFC so far, a concatenation composed entirely of string literals is sugared to a single string literal, so `"abc" <> "d"` would already be sugared as `"abcd"`.

Now, if we're talking about a concatenation with some non-literals, say if we made a variable called `abc` which held the string `"abc"`, you could write `abc <> "d"`, and it would be sugared to `recover String(abc.size() + 1) .> append(abc) .> push('d') end`, due to the optimization I mentioned where a string literal in a concatenation that has a small number of bytes (in this case, 1), could be sugared to a static series of `push` calls instead of an `append` call.

Praetonus:
> is there any reason to use a character literal in a string concatenation?

None that I can think of, but I think it's appropriate to mention it in the RFC for completeness.

SeanTAllen:
I wish we could do this without new syntax.  I'd like to explore ideas on how we could do this while keeping the "+" syntax.

jemc:
@SeanTAllen - I'm not sure that's feasible - as I mention in the RFC, this would have to be implemented in such a way that it would make `<>` be a sugar for this specific construct, rather than sugar for an "operator method" like `+` is for `add`.

Even if we can think of a feasible way to implement it as such, I don't necessarily want to say `+` is sugar for calling the `add` method, except for in these special circumstances, in which it does this other construct with string allocation and `append`. It's more consistent and easy to understand what's going on if the two concepts are kept separate rather than sharing the same syntax.

Note that in this RFC, adding `<>` doesn't create two ways to do the same thing, since it also specifies that `String.add` should be removed.  So the `+` sugar would just be for numeric types (and any other custom types that have an `add` method), but not for strings, which you would now *concatenate* instead of *sum*.

ivanov:
> Even if we can think of a feasible way to implement it as such, I don't necessarily want to say + is sugar for calling the add method, except for in these special circumstances, in which it does this other construct with string allocation and append. It's more consistent and easy to understand what's going on if the two concepts are kept separate rather than sharing the same syntax.

I agree that having `+` stand in for string allocation and append would be inconsistent, but I also think that should not preclude `+` from being kept around for "naive" pairwise concatenation. Python does just this:
```python
>>> 1 + 2 
3
>>> "1" + "2"
'12'
>>> (1,) + (2,)
(1, 2)
>>> [1] + [2]
[1, 2]
```
The typical way of concatenating many strings in a performant manner in Python land is via `"".join((a,b,c,d))`. Couldn't that also be a recommended way of concatenation in Pony, since join will have access to all of the strings? The `DefaultLogFormatter` would then look something like:
```
":".join((loc.file(), loc.line(),loc.pos(), " "+msg))
```

I'm a relatively green Pony tourist and also not a programming languages person, but it seems strange to have syntax for a specific type. I'm not sure of the implementation details, but does it makes sense for such syntax to apply to Array concatenations, for example, or anything that implements Seq, since it has the same flavor of not having to create a bunch of temporaries. 

Another place where pesky temporaries pop up is in numerical codes: If you have element-wise arithmetic implement for n-dimensional arrays,  `a + b - c * d` would entail creating a bunch of temporaries, but you want to reach for something like a generalized version of what the `<>` syntax will do for string concatenation by not allocating a new temporary array for each binary operation.

malthe:
How about `String.nadd(array, count)` – ? The method gets an explicit count of the items and can use a low-level array such that loop unrolling may occur.

That is, keep "+" for concatenation, but if `nadd` is available then use it.

sylvanc:
@jemc is clearly right that we need to do this, in some form.

@kulibali and @SeanTAllen and I were chatting about this on the sync call, and they each came up with a good idea. @SeanTAllen came up with using Julia-like post-type-checking macros (i.e. "generated functions"). @kulibali came up with doing a codegen-time (rather than sugar-time) expansion.

I like both these ideas. I think the Julia-like approach needs more research, but could be widely useful. I think the codegen-time approach could accomplish this now, using the `+` syntax, fairly painlessly.

Julia generated functions:

https://docs.julialang.org/en/stable/manual/metaprogramming/#generated-functions

The codegen-time approach:

For an AST node that is a `String.add` call, check if that AST node is itself an argument to another `String.add` call further up. If so, transform the whole shebang into a series of operations that first counts the total size, then allocates a single string and copies all the args into it.

This would give us the `<>` approach, but using `+`. This is what Java does now, also as a codegen-time optimisation.

agarman:
@sylvanc +1 to generated functions or any metaprogramming that fits cleanly with pony. Though in this case I am unsure how generated functions handle string concat op `add` sugar like "a" + "b" + "c"?

Another option: sugar that collapses multiple ops of same type to a psuedo variadic operator of the same name... eg. instead of `"a" + "b" + "c"` being `add(add("a", "b") "c")` if there was a String fn `add(strings:Array[String]):String` it would be invoked instead.

SeanTAllen:
@agarman the idea is that in neither of the things we discussed are we doing sugar for `add`. In @kulibali's idea, rather than + on a string being sugar for `add`, we do codegen time generation of the required code that would be like hand writing code as seen in the [limiting string allocations](https://patterns.ponylang.org/performance/limiting-string-allocations.html) pony pattern.

agarman:
@SeanTAllen I understand what @kulibali was suggesting. I don't understand how generator functions would be used similarly.

Rather than making the compiler "smart" specifically about add() w.r.t. String, it'd be much better to allow any pony type to specify their own "variadic" add() that would be invoked by pony **and** have this apply to all infix_operators, not just add().

Then any operator for any type can be optimized as appropriate for the type. And the [limiting string allocations](https://patterns.ponylang.org/performance/limiting-string-allocations.html) pattern could become `fun add(strings: Array[String]):String` with two `for` loops over `values()` in strings: the first determines the allocation, the second does the `append()`.

SeanTAllen:
@agarman i'd suggest checking out the sync audio.

jemc:
Going to implement this as an invisible optimization at codegen time, as proposed by @kulibali and @sylvanc.

There are some other aspects to this RFC that we may want to make changes to the language in the future, but in the short term doing an invisible optimization will solve the main motivating problem for this RFC.

PR #91, opened 2017-06-03
Send arbitrary messages to notify classes RFC

SeanTAllen:


jemc:
@SeanTAllen I'm definitely one of the other folks who has expressed this same need. :hand: 

So I definitely agree with the motivation for this RFC.  In fact I #38 was an RFC I proposed that was motivated by many of the same needs.  I think the core idea behind this RFC is better than #38, However, I have some concerns about the presented design, and would like to propose an alternate design that I came up with while reading this RFC, which I see as being an improvement.

### Concerns about the presented design

* In Pony terms, a "message" is in general considered to be a method call with arguments that is delivered asynchronously to an object (with no return value). To a certain extent this RFC is trying to re-create that same concept - this RFC proposes creating objects that are analogous to native Pony messages (the object type is analogous to the method, the object fields are analogous to the arguments, and the `match` block on the handler is analogous to the mailbox mechanism that finds the correct method to execute), but without language support. I think in doing so we lose some of the benefits of native Pony messages, in addition to creating overloaded concepts. I would prefer a solution that lets us pass statically-typed native Pony messages (method calls with arguments delivered asynchronously) to the notify object.

* Whenever I see `Any` being used as an interface in Pony, I see it as a code smell. Passing your objects through a layer that removes all statically-typed information about them means that anything useful you do with the object later, will 1) require the perf cost of a `match` at runtime and 2) leave the compiler completely behind where it should be helping you confirm the correctness of that part of your program. To be clear, using `Any` in constraints is not a code smell, because it is not the final reified type that will be seen by the compiler - the type specificity is not permanently lost, it's just temporarily set aside within the parameterized type or method that you're working with, and the specific type is still known to the caller that gave it as a type argument. I would prefer a solution that avoids losing type information (and maybe you can guess at this point that my solution idea involved type parameters :wink:)

### Alternate design

I think the best way to approach this is as a modified version of the [Access Pattern](https://ponylang.gitbooks.io/pony-patterns/content/async/access.html). However, instead of wanting to expose an `access` behaviour on the actor to expose synchronous access to itself, we want the actor to expose synchronous access to the `notify` class it holds. If we could do this, we could call any public method of the `notify` class from outside the actor that holds it. That is, we'd be delivering native Pony messages (method calls with arguments delivered asynchronously).

The remaining challenge is that the actor (`TCPConnection`, for example) doesn't know the specific type of the notify object - it sees it only as being an unknown object that complies with the notify interface (`TCPConnectionNotify`, for example). Therefore, a naive implementation would only allow us to call methods of the `TCPConnectionNotify` interface using our access method, which isn't really all that useful, except maybe for simulating traffic.

However, this is where type parameters shine - we can make the actor accept a type parameter that can accept the more specific type of the notify object which exposes the methods that we need (either the concrete type itself, or just some more specific custom interface/trait).

That is, the changes to `TCPConnection` would look like this:

```pony
actor TCPConnection
  var _notify: TCPConnectionNotify
  // ...
```

```pony
actor TCPConnection[N: TCPConnectionNotify ref = TCPConnectionNotify]
  var _notify: N
  // ...
  
  be access_notify(fn: {(N)} val) =>
    fn(_notify)
  
  // ...
```

Note that the type parameter has a default of `TCPConnectionNotify`, so there's no need to specify an explicit type argument unless you need to use the `access_notify` method with a more specific type than that interface. As you said, this is an advanced feature, so having a default that doesn't require any type arguments for simpler use cases is a nice help. Also, note that once we have [Generic Type Inference](https://github.com/ponylang/rfcs/blob/master/text/0010-generic-type-inference.md), I don't think it will be necessary for anyone to have to specify a type argument.

With that in mind, an application that uses this advanced feature might look something like this:

```pony
use "net"

class MyNotify is TCPConnectionNotify
  // ...
  
  fun ref my_message(arg1: String, arg2: String) =>
    // ...

actor Main
  new create(env: Env) =>
    let conn =
      try
        TCPConnection[MyNotify](env.root as AmbientAuth, MyNotify, "", "8989")
      else
        env.out.print("Couldn't create the TCPConnection...")
        return
      end
    
    conn.access_notify({(notify: MyNotify) =>
      notify.my_message("Hello", "World")
    })
```

jemc:
@SeanTAllen One more thing worth noting - off the top of my head I don't think your design in this RFC will compile - I think you can only use `#send` in a type parameter constraint, so I don't think you can use `Any #send` as an argument type.

SeanTAllen:
I'm going to close this RFC. @jemc is opening an RFC with his much better than Sean idea.

Joe's idea requires a compiler change to allow reference capabilites on type parameters.

jemc:
I was coming back to write up this RFC, and I'm noticing that a lack of higher-kinded types is going to make my proposal a little more unwieldy.

That is, you'd have no way to specify "any TCPConnection" as a type, since every `TCPConnection` type is only going to have a subtyping relationship with others that have the same notify type parameter (`N`).

We could add an `interface TCPConnectionAny`, but that's not very nice...

PR #93, opened 2017-07-06
Feature: Add RFC for new gencap, `#write`.

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/feature/gencap-write/text/0000-gencap-write.md).

### DIFF
--- a/text/0000-gencap-write.md
+++ b/text/0000-gencap-write.md
@@ -1,0 +1,66 @@
+- Feature Name: gencap-write
+- Start Date: 2017-07-05
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Add `#write`, a new gencap for use in type parameter constraints, implying the capability set: `{iso, trn, ref}`.
+
+# Motivation
+
+We have gencaps to act as capability sets in type parameter constraints, grouped by the situation in which you might want to use them:
+
+| gencap   | `iso` | `trn` | `ref` | `val` | `box` | `tag` | summary                       |
+|----------|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|-------------------------------|
+| `#any`   |   X   |   X   |   X   |   X   |   X   |   X   | any possible cap              |
+| `#send`  |   X   |       |       |   X   |       |   X   | sendable                      |
+| `#alias` |       |       |   X   |   X   |   X   |   X   | alias as themselves           |
+| `#share` |       |       |       |   X   |       |   X   | alias as themselves, sendable |
+| `#read`  |       |       |   X   |   X   |   X   |       | alias as themselves, readable |
+
+This proposal adds the following new gencap to the table:
+
+| gencap   | `iso` | `trn` | `ref` | `val` | `box` | `tag` | summary                      |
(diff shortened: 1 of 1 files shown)


Praetonus:
:+1:

plietar:
Sounds reasonable. Just a few points,

Do we also need an extra gencap for `#write! = { ref, box, tag }` ? (I can't think of any meaningful name for that set)
https://github.com/ponylang/ponyc/blob/master/src/libponyc/type/cap.c seems like we would need it

`#read` does not allow `iso` and `trn` even though they are readable, but `#write` would. This could be surprising (I was really confused as to why `#read` did not allow these when I started looking into generics)

Do we want a way to write the following, where `this` could be one of `ref`, `trn` or `iso`, depending on the receiver capability ?
```pony
fun #write replace(x: A) : this->A^ =>
  this.contents = consume x
```

Praetonus:
> Do we want a way to write the following, where this could be one of ref, trn or iso, depending on the receiver capability ?

I think cases like that are already covered by automatic receiver recovery.

jemc:
> `#read` does not allow `iso` and `trn` even though they are readable, but `#write` would. This could be surprising (I was really confused as to why `#read` did not allow these when I started looking into generics)

Yeah, I remember being struck surprised by this as well. However, I'm not sure what could/should be done :thinking:

> Do we also need an extra gencap for `#write! = { ref, box, tag }` ? (I can't think of any meaningful name for that set)
https://github.com/ponylang/ponyc/blob/master/src/libponyc/type/cap.c seems like we would need it

Hm, I'm not sure where you think we need it. You're probably right, and I'm probably wrong, but I already implemented my first draft of this in a branch, and didn't use such a set.  We should go over this again in code review for the implementation.

plietar:
> > Do we want a way to write the following, where this could be one of ref, trn or iso, depending on the receiver capability ?

> I think cases like that are already covered by automatic receiver recovery.

I was thinking more about reflecting the receiver capability in the `this` viewpoint, but I've realised `ref` is the most flexible viewpoint anyway, so there's not much point doing that.

> > Do we also need an extra gencap for #write! = { ref, box, tag } ? (I can't think of any meaningful name for that set) https://github.com/ponylang/ponyc/blob/master/src/libponyc/type/cap.c seems like we would need it

> Hm, I'm not sure where you think we need it. You're probably right, and I'm probably wrong, but I already implemented my first draft of this in a branch, and didn't use such a set. We should go over this again in code review for the implementation

I'm talking about the `cap_aliasing` function, which takes a (gen)cap and returns its alias. Today, all existing gencap have a alias (`#any`->`#alias`, `#send`->`#share`, others don't change), but `#write` would not.

This can be seen as an implementation detail, with a `#writealiased` gencap if it is indeed necessary, which wouldn't have to be expose to source programs.

I've always seen this is the reason `#alias` exists. The documentation says something about it being only useful internally in the compiler.

jemc:
I'm putting this into final comment period.

Will address @plietar's concern in the implementation, and we can discuss it more in the PR - I don't think we want to add another user-facing gencap to accomodate `#writealiased` - if we need another such gencap it should be an implementation detail as he suggested.

PR #95, opened 2017-07-13
Adding RFC for a Match iterator

autodidaddict:
Adding an RFC to enable a match to be re-run against the same subject string until there are no more matches.

### DIFF
--- a/text/0000-match-iter.md
+++ b/text/0000-match-iter.md
@@ -13,13 +13,13 @@ Not all regular expressions can know the full extent or the total number of grou
 
 # Detailed design
 
-The implementation of this would be an addition to the standard library. We want to be able to jump from a validated regular expresion (e.g. the result of an `apply` on a `Regex`) to an iterator. One way to to this would be to add a `values()` method to `Match` that returns `Iterator[Match]`. 
+The implementation of this would be an addition to the standard library. We want to be able to jump from a validated regular expresion (e.g. the result of an `apply` on a `Regex`) to an iterator. One way to to this would be to add a `matches()` method to `Regex` that returns `Iterator[Match]`. 
 
-By implementing this iterator in the standard library, developers will have access to this functionality in a standardized, reliable fashion. Here is an example of how this might be used where a single (first) `Match` is converted into an iterator via the idiomatic `values()` method:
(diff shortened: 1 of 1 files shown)


text/0000-match-iter.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Detailed design
+
+The implementation of this would be an addition to the standard library. We want to be able to jump from a validated regular expresion (e.g. the result of an `apply` on a `Regex`) to an iterator. One way to to this would be to add a `values()` method to `Match` that returns `Iterator[Match]`. 
+
+By implementing this iterator in the standard library, developers will have access to this functionality in a standardized, reliable fashion. Here is an example of how this might be used where a single (first) `Match` is converted into an iterator via the idiomatic `values()` method:
+
+```pony
+let r = Regex("([+-]?\\s*\\d+[dD]\\d+|[+-]?\\s*\\d+)")
+for m in r("3d6+10+1d20").values() do
```
jemc: Looking at this code snippet, I'd actually expect it to be doing something other than what you've described.

Since `r("3d6+10+1d20")` returns a single `Match` object, that means you're calling `Match.values()` here, which implies to me that you'll be iterating over each of the capture groups that are accessible via `Match.apply(index: USize)`. That is, I'd expect it to be an `Iterator[String]` where each `String` is a capture group, instead of an `Iterator[Match]` where each `Match` is a subsequent match in the original string, as you've laid out in the RFC.

What I think we probably want is something like this:

```pony
let regex = Regex("([+-]?\\s*\\d+[dD]\\d+|[+-]?\\s*\\d+)")
or m in regex.matches("3d6+10+1d20") do
  // ...
end
```

jemc (commented):
As we discussed on IRC, I think this makes a great first RFC!

However, I did have a comment/suggestion to make about the API you proposed so far.

text/0000-match-iter.md
autodidaddict: I agree. I went looking for examples of how other objects convert into iterators and most of them used this `values()` idiom. My big concern with the code sitting in the RFC as it stands now was that it's just not clear what kind of values you're getting, as you mention.

I think I'll update the RFC to use a more self-explanatory syntax.

jemc (approved):

jemc:
This seems non-controversial, so we're moving it into final comment period.

PR #97, opened 2017-07-21
Add RFC for more Random methods.

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/feature/more-random-methods/text/0000-more-random-methods.md).

### DIFF
--- a/text/0000-more-random-methods.md
+++ b/text/0000-more-random-methods.md
@@ -48,7 +48,7 @@ The following array-oriented method will also be added to the `Random` trait.
 To support `Random.shuffle`, the following method will be added to the `Array` class (because otherwise it is not possible to swap two elements in place without aliasing them):
 
 ```pony
-  fun ref swap(i: USize, j: USize) ? =>
+  fun ref swap_elements(i: USize, j: USize) ? =>
     """
     Swap the element at index i with the element at index j.
     If either i or j are out of bounds, an error is raised.
--- a/text/0000-more-random-methods.md
+++ b/text/0000-more-random-methods.md
@@ -1,0 +1,76 @@
+- Feature Name: more-random-methods
+- Start Date: 2016-07-20
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This RFC proposes the addition of various methods to the `Random` trait of the `random` package. It also proposes an additional method to the `Array` class.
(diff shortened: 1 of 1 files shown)


Praetonus (commented):
I think we should also add `iN` methods for signed integers, for convenience.

Also, what do you think about using `swap_elements` instead of `swap` for the new `Array` method? I've been thinking about using `swap` for a general idiom on collections and other types with value semantics to swap two objects. It could be used to cheaply "reconstruct" an embedded field, for example. I'll write an RFC on that eventually.

text/0000-more-random-methods.md
```diff
@@ -0,0 +1,76 @@
...
+
+This RFC proposes the addition of various methods to the `Random` trait of the `random` package. It also proposes an additional method to the `Array` class.
+
+# Motivation
+
+The purpose of these additional methods is to increase the functionality of the `Random` trait.
+
+# Detailed design
+
+The following methods will be added to the `Random` trait, similar to the existing `u8`, `u16`, `u32`, `u54`, and `u128` methods:
```
Praetonus: There's a typo here. `u54` should be `u64`.

jemc:
I'll update this with @Praetonus' requested additional methods soon, before the vote.

mfelsche:
Just a small question: 

why is the int method specified for a exclusive range ``[0, max)`` and not ``[0, max]``. With the former it is not possible to generate values within the whole int range, which might be a nice feature for some use cases. Or is there a trivial way to convert a ``[0, max)`` to ``[0, max]``, keeping the distribution properties of the exclusive range?

But maybe this should be discussed in a different rfc.

jemc:
@mfelsche Note that if you need to use the whole int range, you can use one of the other methods that doesn't use the floating point arithmetic - they will be faster / more direct anyway.

That is, instead of using `rand.int[U32](U32.max_value())`, you can just use `rand.u32()`

mfelsche:
It turns out i misunderstood the documentation of ``random.Random`` which was perfectly clear on that.
Using the ``uX()`` and ``iX()`` methods (to be added) is all that is needed.
thanks for the clarification!

jemc:
@Praetonus - I added the `iX` methods in ae5a959.

Issue #48, opened 2016-10-19
Change name of MaybePointer in builtin.


jemc:
We need someone to come up with a less confusing name for the `MaybePointer` type, and champion that in an RFC.

See https://github.com/ponylang/ponyc/issues/1248 for more background.

mfelsche:
Suggestion: `NullableStruct[T]`

PR #98, opened 2017-07-26
RFC to change String.join to take an Iterable instead of a ReadSeq.

mikeb01:


SeanTAllen:
I don't understand the advantage of switching to iterator from ReadSeq. You state "The Iterator interface placed fewer requirements on the implementation when compared to ReadSeq". It's not clear to me what that means. 

Is the idea that you have a scenario where you only have an iterator and not access to the ReadSeq itself? I'm wary of this change. Some additional concrete examples of problems this solves would help me.

mikeb01:
> I don't understand the advantage of switching to iterator from ReadSeq. You state "The Iterator interface placed fewer requirements on the implementation when compared to ReadSeq". It's not clear to me what that means.

What I meant by that is that a `ReadSeq` interface requires that your implementation have random access (`apply(USize): T`), have a known size (`size(): USize`) and support sequential (ordered) access by returning an `Iterator` from the `values()` function.  Where as the `Iterator` only requires that the implementation have sequential access, so is a subset of the functionality if `ReadSeq`.  There are a number of cases where supporting an `Iterator` interface rather than `ReadSeq` would be desirable.  Supporting a database cursor or a stream from a network service are two that come immediately to mind.

The other case that crops up is compatibility with other parts of the pony stdlib.  The itertools package which provides a number of useful functions over unbounded streams of data, similar java streams library.  The specific case I was looking at taking a stream of data, mapping a transformation over it, then joining the result.  Itertools was the only part of the library that supported that sort of functionality, but didn't provide a join or reduce function.  I found the join function on String, which required that the call supply a `ReadSeq`, but internally only ever needs in `Iterator`.  It seems a simple and obvious change to make the `join` function take an `Iterator` making it a more useful function.

I only have the supplied specific example.  I could cook a few more, but they would just be variations on that same theme and not really add value.  I'm not sure what make you wary of the change.  The only really controversial part would be that it is an API breaking change.

SeanTAllen:
Ah, that explanation makes a lot of sense @mikeb01.  Can you update the RFC accordingly?

plietar:
Maybe we should have a `Iterable`interface like this :
```pony
interface Iterable[T]
  fun iterator(): Iterator[this->T]
```
and change the `join` method (and possibly other places in the stdlib) to take an `Iterable[String]` argument.

If all `ReadSeq[T]` implement `Iterable[T]`, then `String.join(["Hello, " ; "world!"])` still works.
By making `Iterator[T]` implement `Iterable[T]` we can now use the examples from the RFC.

Other types can also implement `Iterable[T]` without implementing either `ReadSeq[T]` or `Iterator[T]` (a tree for example).

We can also  change for-loop desugaring to use this interface, making this possible :
```pony
for x in ["Hello, " ; "world!"]
do env.out.print(x)
end
```

This interface exists in many other languages :
- [Iterable interface](https://docs.oracle.com/javase/7/docs/api/java/lang/Iterable.html) in Java
- [Sequence protocol](https://developer.apple.com/documentation/swift/sequence) in Swift
- [IntoIterator trait](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) in Rust

SeanTAllen:
@plietar there's an RFC open for this now, can you copy your comment over there?

https://github.com/ponylang/rfcs/pull/98

SeanTAllen:
Closing in favor of RFC: https://github.com/ponylang/rfcs/pull/98

SeanTAllen:
Oops I didnt mean to close this.

mikeb01:
Updated RFC with the explanation from above.

SeanTAllen:
@mikeb01 we have put this into "final comment period". from here there is one week for people to make comments at that point if everything looks good, we vote.

cquinn:
What was the outcome of incorporating @plietar's suggestion above? I really like that approach.

jemc:
I personally think that @plietar's suggestion should be its own RFC, since it's a change that is rather larger in scope.

I have some concerns about how it might play out in practical terms, but maybe we can discuss those in that other RFC?

plietar:
Sure, I can write an RFC for that

SeanTAllen:
Congrats @mikeb01! First accepted RFC. Thank you!

mikeb01:
Excellent.  The implementation is already done in the closed PR: https://github.com/ponylang/ponyc/pull/2076, should I just re-open it?

SeanTAllen:
That works @mikeb01. Can you use `git commit --amend` to update the commit comment to reference https://github.com/ponylang/rfcs/blob/master/text/0048-change-String-join-to-take-iterable.md

As well as explaining the motivation for the change. The motivation section from the RFC would make a pretty good commit body along with the link to the RFC.

Something like

-----

RFC #48 Change String.join to take Iterable

Implements RFC #48. For full details see https://github.com/ponylang/rfcs/blob/master/text/0048-change-String-join-to-take-iterable.md.

The Iterator interface places fewer requirements on the implementation when compared to ReadSeq.  The ReadSeq requires that your implementation have random access (apply(USize): T), have a known size (size(): USize) and support sequential (ordered) access by returning an Iterator from the values() function. Where as the Iterator only requires that the implementation have sequential access, so is a subset of the functionality if ReadSeq. There are a number of cases where supporting an Iterator interface rather than ReadSeq would be desirable. Supporting a database cursor or a stream from a network service are two that come immediately to mind.

The other case that crops up is compatibility with other parts of the pony stdlib. The itertools package which provides a number of useful functions over unbounded streams of data, similar java streams library. The specific case I was looking at taking a stream of data, mapping a transformation over it, then joining the result. Itertools was the only part of the library that supported that sort of functionality, but didn't provide a join or reduce function. I found the join function on String, which required that the call supply a ReadSeq, but internally only ever needs in Iterator. It seems a simple and obvious change to make the join function take an Iterator making it a more useful function. It is trivial to convert from a ReadSeq to an Iterator, however it is more difficult and has a higher cost to convert in the other direction.

PR #101, opened 2017-08-08
Add RFC for improved-itertools-api

Theodus:
[Rendered](https://github.com/ponylang/rfcs/blob/itertools/text/0000-improved-itertools-api.md)

### DIFF
--- a/text/0000-improved-itertools-api.md
+++ b/text/0000-improved-itertools-api.md
@@ -9,7 +9,7 @@ The goal of this RFC is to improve the usability and reduce the maintenance cost
 
 # Motivation
 
-The motivation for this RFC comes from concerns that I have gathered from my own experiences and from the concerns and recommendations of members of the Pony community. The primary concern that I have heard is inconvenience caused by `fold` being partial. Also, adapting iterators with any state is impossible since the arguments to methods such as `map` and `filter` are lambdas, which means that their `apply` methods have a capability of box. This lack of stateful adaptations also has the side effect of making the `Iter` class prone to code duplication to handle common tasks like stashing values in order to maintain the semantics of filtering. With the addition of methods like `flat_map` and `filter_map`, the problem of code duplication becomes a more noticeable issue for future additions to the API.
+The motivation for this RFC comes from concerns that I have gathered from my own experiences and from the concerns and recommendations of members of the Pony community. The primary concern that I have heard is inconvenience caused by `fold` being partial. Also, adapting iterators with any state is impossible since the arguments to methods such as `map` and `filter` are lambdas with a capability of box. This lack of stateful adaptations also has the side effect of making the `Iter` class prone to code duplication to handle common tasks like stashing values in order to maintain the semantics of filtering. With the addition of methods like `flat_map` and `filter_map`, the problem of code duplication becomes a more noticeable issue for future additions to the API.
 
 # Detailed design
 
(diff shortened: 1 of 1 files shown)


jemc:
I think the "Motivation" section needs to be fleshed out a bit more. This RFC should serve as a historical document that explains the "concerns that I have gathered from my own experiences" and the "concerns and recommendations of members of the Pony community" in detail, illustrated with examples.

There are some parts of this RFC that are fairly obvious to me why they are needed, and some that are really not obvious to me. Also, things that are obvious to me will not be obvious to everyone. This is a big RFC (touches many concerns), and It's best to explain the motivation for each concern.

Theodus:
@jemc I agree. Further explanation has been given.

jemc:
~~~I'm not sure yet that I'm getting the need for the `_adapter` variants yet.~~~

I'm going to plan to take another look at this soon and make some additional comments ~~~on that~~~.

text/0000-improved-itertools-api.md
```diff
@@ -0,0 +1,121 @@
...
+  """
+
+fun ref filter_map_adapter[B](adapter: IterFilterMapAdapter[A, B]): Iter[B]^
+  """
+  Allows stateful modification to the stream of elements from an iterator,
+  similar to `filter_map`.
+  """
+```
+
+Many of the existing methods will be re-implemented using these three methods in order to reduce the probability of logical errors related to coordination between the `has_next` and `next` methods to stash the next value of the iterator. Because of this, the return types of some methods will change from `Iter[A]` to `Iter[A!]`. A notable example is the `enum` method, where the return type was previously `Iter[(B, A)]^`:
```
jemc: This is a downgrade, right?

I'd prefer to see the signature remain the same - having easier maintainability is not worth downgrading the cap possibilities of the API.

text/0000-improved-itertools-api.md
```diff
@@ -0,0 +1,121 @@
...
+
+```pony
+fun ref enum[B: (Real[B] val & Number) = USize](): Iter[(B, A!)]^
+```
+
+The following methods will be added to the Iter class:
+
+```pony
+fun ref filter_map[B](f: {(A!): (B | None) ?} box): Iter[B]^
```
jemc: What does raising an error in the lambda do?

The same as returning `None`?

text/0000-improved-itertools-api.md
```diff
@@ -0,0 +1,121 @@
...
+This will break existing code that uses the current `fold` method. Also, the minor change in return type will break any code that attempts to access a val field from the return value of an `Iter` of some iso type. It should be noted that this will have no effect on any `Iter` of some iso^ type since `A^! -> A`.
+
+# Alternatives
+
+We may choose not to include the iterator adapters if the change in return type is not worth reducing the maintenance cost of the itertools package.
+
+# Unresolved questions
+
+- Should the `acc` argument to `fold` and `fold_partial` be the first argument? How may this affect partial application?
+- Should we keep the classes such as `MapFn` and `Take`? Users may be confused by some methods of the `Iter` class having external implementations while most do not.
```
jemc: Seems to me that `MapFn` and `Take` have no reason to stay around - they seem to be fully redundant with methods of `Iter`.

text/0000-improved-itertools-api.md
```diff
@@ -0,0 +1,121 @@
...
+
+This will break existing code that uses the current `fold` method. Also, the minor change in return type will break any code that attempts to access a val field from the return value of an `Iter` of some iso type. It should be noted that this will have no effect on any `Iter` of some iso^ type since `A^! -> A`.
+
+# Alternatives
+
+We may choose not to include the iterator adapters if the change in return type is not worth reducing the maintenance cost of the itertools package.
+
+# Unresolved questions
+
+- Should the `acc` argument to `fold` and `fold_partial` be the first argument? How may this affect partial application?
```
jemc: I vote yes. This makes for cleaner "trailing lambda" syntax, and also seems to be more convenient for partial application where you start by partially-applying the `acc`.

text/0000-improved-itertools-api.md
```diff
@@ -0,0 +1,121 @@
...
+
+interface IterFilterAdapter[A]
+  fun ref apply(a: A!): Bool ?
+
+interface IterFilterMapAdapter[A, B]
+  fun ref apply(a: A!): (B | None) ?
+```
+
+```pony
+fun ref map_adapter[B](adapter: IterMapAdapter[A, B]): Iter[B]^
```
jemc: I don't really follow the name choice for `map_adapter`... How would you feel about `map_stateful`, or `stateful_map`?

text/0000-improved-itertools-api.md
```diff
@@ -0,0 +1,121 @@
...
+  """
+  A partial version of `fold`.
+  """
+```
+
+The following "iterator adapter" interfaces will be added to the package along with corresponding methods of the `Iter` class. These interfaces will facilitate stateful modification to the output of the iterator.
+
+```pony
+interface IterMapAdapter[A, B]
+  fun ref apply(a: A!): B ?
```
jemc: I'd just note that these interfaces aren't strictly necessary - you can use the inline lambda syntax instead: `{ref(A!): B ?} ref`

Also, note that when lambda inference is complete, users will be able to give a lambda where this interface is expected, without having to specify `ref`-ness in their lambda literal (it will be implied by the antecedent type).

Theodus: I had no idea that this was possible, I'll likely rethink the adapter interfaces given this possibility.

text/0000-improved-itertools-api.md
Theodus: I prefer those names given the possibility of removing the adapter interfaces.

text/0000-improved-itertools-api.md
Theodus: No, an error will result in the `next` method raising an error.

Theodus:
@jemc After further investigation the type downgrade from `A` to `A!` is unnecessary in the cases where filter semantics are used since `fold_stateful` can return `Iter[A]` instread of `Iter[A!]`.

jemc (approved):

Theodus:
Is this good to move on to final comment period?

jemc:
Yes, I think so.

PR #103, opened 2017-10-05
Expose ponyc version to pony program

nisanharamati:
Being able to determine the version of ponyc an application was compiled with could be helpful when handling bug reports and in cases where there can be confusion.
It would also be helpful in determining whether a particular application is vulnerable or sensitive to bugs that were fixed in a later ponyc version.

[rendered](https://github.com/nisanharamati/rfcs/blob/6c7a82a7a4ec0fc4a68ced8fab3de6bd0a77ec99/text/0000-expose-ponyc-version-to-program.md)

### DIFF
--- a/text/0000-expose-ponyc-version-to-program.md
+++ b/text/0000-expose-ponyc-version-to-program.md
@@ -1,0 +1,40 @@
+- Feature Name: expose_ponyc_version
+- Start Date: 2017-10-05
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Ponyc's version will be exposed to a pony program, so the program could report the version of ponyc that it was built with.
+
+# Motivation
+
+This proposal will allow pony programs to access the ponyc version they were built with via a pony function. This will enable programs to include the ponyc version they were built with in their own version (if they were designed to report it), similarly to the way `ponyc --version` includes the version of LLVM it was compiled with.
+
+# Detailed design
+
+The version information already exists in the ponyc binary, since it can report its version and the compiler it was built with. This information could be exposed via a pony runtime function which would return a string of the ponyc version.
+Optionally, an additional function could return a string of the LLVM version ponyc was compiled with.
+
+# How We Teach This
+
+This feature should include documentation in the standard library documentation, and it would be helpful to include an example of how to get the ponyc version in your application version in the pony patterns cookbook, and the pony tutorial.
+
+# How We Test This
+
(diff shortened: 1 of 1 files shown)


aturley:
I was thinking about this last night and couldn't come up with a reason for exposing this information other than for checking it if something goes wrong. Consequently, I don't think the program needs to have access to it within the code.

I'm still not completely sold on whether or not this needs to be supported at all, but if we do it I think I'd rather add it as a command line option for the program, as you described in the "Alternatives" section.

## I could be wrong

I may not have thought through all the use cases, so if there are compelling reasons to make this available inside the program I'd be very interested in hearing them.

## If we do end up doing this ...

### Questions

Where would this function live? Do we need to think about how to control access to this function, or does everybody get access to it?

### Testing

Is this something that could be tested like the other test functions that use the JIT to compile Pony code at runtime? I don't know how all the pieces would fit together, but this seems like an approach that might work.

jemc:
The best use case I can dream up for this feature to be exposed to the program internally is in the case where you want to be able the program to send some form of automatic bug report by "phoning home" in the style that some applications seem to favor, and you wanted to include the ponyc version in the report.

Given Pony's style of error handling and lack of unhandled exceptions rising to the top of the stack, this may not actually be a pattern that anyone really wants to use in practice.

If we can't come up with a compelling use case that will actually benefit someone with a real world problem they're having, it might be prudent to leave it out until we do have that arise.

I really like the idea of having it as a `--ponyversion` CLI arg, because it uses established patterns we already have rather than breaking new ground.

@nisanharamati - does that alternative work for your use case?

If so, we might be able to bypass the RFC process for that, since it's not really going to have an effect on the standard library or language itself.

mfelsche:
The use case for which i would love to have the pony version exposed is for having conditionals in my code based on the pony version, most likely ifdefs. This means exposing the pony version during compile time

This would make it possible to maintain compatibility between different pony versions within a single codebase; in a somewhat limited fashion: e.g. a breaking change like explicit partial calls might lead to horrible ifdef hell, but something like the recent changes to the Iter class should be easily handled by such a conditional on the pony version.

I know this is a somewhat different story, as this requires the pony version to be exposed during compile time, not runtime and should be handled in another rfc. But i wanted to raise it here nonetheless.

nisanharamati:
@jemc yes, I think the `--ponyversion` alternative works for our use case!

Theodus:
@nisanharamati In that case, would you mind opening an issue on the ponyc repo and closing this RFC?

nisanharamati:
@Theodus Done: https://github.com/ponylang/ponyc/issues/2285

Theodus:
Thanks @nisanharamati

PR #116, opened 2018-01-07
Rename Logger in net/http to HTTPLogger

EpicEric:
[Rendered](https://github.com/EpicEric/rfcs/blob/12bbc3610e10fc9d5002eea72cbb8d3ddcf1d977/text/0000-rename-http-logger-interface.md).

As pointed out by ponylang/ponyc#2471.

### DIFF
--- a/text/0000-rename-http-logger-interface.md
+++ b/text/0000-rename-http-logger-interface.md
@@ -1,0 +1,50 @@
+- Feature Name: rename-http-logger-interface
+- Start Date: 2018-01-06
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Rename `interface val Logger` in the `net/http` stdlib package to `interface val HTTPLogger`.
+
+# Motivation
+
+Different libraries in the stdlib are (usually) expected by users of Pony to be conflict-free, without the use of aliases. However, the interface `Logger` in `net/http` currently clashes with the class `Logger` in `logger`. This means that the following code currently leads to an error regarding "existing type name classes" on compilation:
+
+```pony
+use "logger"
+use "net/http"
+
+actor Main
+  new create(env: Env) =>
+    None
+```
+
+# Detailed design
+
+As a matter of "principle of least surprise", the class in `logger` should remain named as `Logger`. This arises from the fact that the class provides a bulk of the core functionality of its homonymous package.
+
+Therefore, all uses of the `Logger` interface within and on uses of the `net/http` library should be renamed to `HTTPLogger`.
+
+This will allow for both packages to be used simultaneously in Pony programs, without any aliases.
+
+# How We Teach This
+
+The name of this interface will be updated in the stdlib documentation.
+
+# How We Test This
+
+Current `net/http` tests being run on CI should not break once all instances of the interface are renamed.
+
+# Drawbacks
+
+* Breaks any existing code that calls this interface explicitly, although the fix should be simple by following the updated docs.
+* Adds verbosity to the type name.
+
+# Alternatives
+
+Do not update the interface name, and instruct users to use a package alias as instructed in the Pony Tutorial.
+
+# Unresolved questions
+
+None.


SeanTAllen:
@EpicEric this is a very nice RFC. brief but covers everything really well. Thank you.

mfelsche (approved):

Praetonus (changes requested):
I'm personally opposed to this change, as package import aliases exist to solve that problem. I'd be more inclined to support an RFC to improve import aliases and make them less constraining to use. Some good candidate features could be to move the error point from the import to the actual name use, and to add optional default package aliases.

gin66:
@Praetonus Package import aliases may solve this kind of problems if non-stdlib packages need to be imported. From an user perspective a name clash in a stdlib is...well...sorry to say so...poor library design.
An alternative would be to kick out Logger package from stdlib, though.

@EpicEric Thanks for coming up with this good RFC and perfect solution.

SeanTAllen:
@gin66 you are one user. please don't speak for all users. we can have far more productive conversations if we talk about what we feel is important rather than taking on the mantel of being the spokesperson for everyone else.

There are pros and cons to each approach.

Praetonus:
@gin66 While I agree that having multiple packages with a `Logger` type is bad, I don't think the issue is on naming. I think it is on what a type named `Logger` means, and functionality duplication.

A logger is a pretty general concept. It constructs output from a data source and outputs it to some destination. `logger.Logger` is agnostic to both the data source and the output destination. `http.Logger` is agnostic to the output destination, but not to the data source: it can only log HTTP request/response pairs. From a type contract standpoint, this means that `http.Logger` is a subtype (not in the Pony type system sense) of `logger.Logger`: it provides a more specific contract. From that, it would make sense to rename `http.Logger` to `RequestResponsePairLogger` (or some less verbose term), but there are some other issues that make this solution not ideal.

Now if we look at the implementation of both types, some functionalities of `logger.Logger` are duplicated in `http.Logger`, namely the outputting actions. There is one more issue in `http.Logger`, which is that it has more than one responsibility. It handle both the output construction and the outputting action, while `logger.Logger` delegates the output construction to the [lambda](https://github.com/ponylang/ponyc/blob/master/packages/logger/logger.pony#L113) passed as a constructor parameter.

So instead of simply renaming `http.Logger` to `RequestResponsePairLogger`, I think `http.CommonLog` and similar should wrap `logger.Logger`. `RequestResponsePairLogger` can then be added as an helper type alias if need be.

jemc (approved):
I'm in favor of this RFC.

In this case, the name `Logger` is inconsistent with the other names in the `net/http` package - we have `HTTPServer`, `HTTPSession`, `HTTPHandler`, etc. So in that context, I think a name change to `HTTPLogger` is appropriate, unless we also want to change the names of all those other entities.

I agree with @Praetonus that scoped package imports can be used to solve this, and are a preferable solution in general, but I'd only advocate using that approach here if we also used it for all the other `net/http` types.

I don't agree with @gin66 that all name clashes in the stdlib are inherently bad.

mfelsche:
I am 100% with @Praetonus on this.

SeanTAllen:
I think we should discuss this as a general topic. Given that I am open an RFC to remove the HTTP server from the standard library and that has general consensus support, I think this specific case doesn't matter much.

Personally I would be in favor of changing the logger name in http to httplogger because it matches the naming convention in the http code, not because of the name clash. Basically, I'm with jemc on this. I think this one class is poorly named.

Praetonus:
@EpicEric @SeanTAllen What should be done about this RFC with regard to #117?

EpicEric:
With #117 being approved, there's no point in having an RFC for something that isn't in the standard library anymore.

Still, this would be simply a matter of sanity, as @jemc and @SeanTAllen pointed out. I don't know if I should: rewrite this with that goal + the eventual outcome of the other RFC in mind; or discard it and suggest a change in the new repo instead.

mfelsche:
@EpicEric to me, the changes mentioned by @Praetonus in https://github.com/ponylang/rfcs/pull/116#issuecomment-355970000 are definitely necessary, be it in a separate package or still in the ponyc stdlib. 

While I am in favour of having a clean stdlib, I can totally understand waiting with this change until the ′net/http′ package has been moved to a separate repository and thus dropping this PR here in favour of submitting this change to the new repo.

SeanTAllen:
Closing as HTTP is being removed from standard library. Any changes can be made in new repo without having to go through RFC.

PR #117, opened 2018-01-13
Remove HTTP libraries from the standard library

SeanTAllen:


text/0000-remove-http-server-from-stdlib.md
```diff
@@ -0,0 +1,59 @@
...
+
+CI should pass for ponyc repo after the change is made. CI should be set up for the new repo to run HTTP server tests and examples to verify that everything was correctly moved.
+
+# Drawbacks
+
+This is a breaking change for anyone using the HTTP server. They will have to go through a small upgrade to use Pony stable to fetch the new dependency from a new location. Additionally, package names will change as it will no longer be `net/http`. That package name makes sense in the standard library, it doesn't as a standalone.
+
+# Alternatives
+
+We could leave in the standard library and incur the drawbacks that come from people thinking this is the officially sanctioned HTTP server that we as a community want to go forward with.
```
mfelsche: What about keeping it, but not as it is, but with a substantial rework, addressing first and foremost performance and stability?

mfelsche:
How about the HTTP client? Why should this be allowed to stay in stdlib given the argumentation for removing the server? (asking for a friend)

SeanTAllen:
Sorry, I should update the title. I think we should remove the client as well. That is part of "net/http".

I'm not in favor of leaving it. Folks have had plenty of time to rework it. I think it should be removed. I don't think an HTTP server belongs in the standard library. I'm rather dubious about an HTTP client as well.

text/0000-remove-http-server-from-stdlib.md
SeanTAllen: I think it should be removed and if someone wanted to open an RFC to add a new reworked one, they could. I don't think I would be in favor of it, but that is something folks could propose and could be discussed.

winksaville:
If this functions and passes its tests, what about moving it to examples/ with caveats on its major faults. I'm sure there is something neophytes like myself will learn alot from if there are some explanations of why its "bad". Actually, I'll probably learn more from "bad" code with explanations then undocumented "good" code.

I'd even go as far as saying the "bad" code will become even more valuable when there is "good" code, because the neophytes will have something to compare against. I'll volunteer to add the explanations  if somebody can give me some bullet points on what's "bad".

SeanTAllen:
I never said "bad". I want to be clear on that. I said the performance is poor and the API ergonomics are lacking (this has been brought up my multiple users trying to get started with it).

At no point did I say "bad" or "good". I intentionally avoid characterizing code that way. I don't think the code is bad.

winksaville:
My apologies.

winksaville:
I didn't mean to offend anyone, and I apologise for the poorly spoken comment.

Praetonus:
@SeanTAllen Should this go to final comment period?

SeanTAllen:
@Praetonus i need to update it to be more clear about removal of server and client and what happens to each.

text/0000-remove-http-server-from-stdlib.md
```diff
@@ -0,0 +1,59 @@
...
+
+We, as a community, want a great HTTP server that we can use. Having a middling HTTP server in the standard library doesn't help the cause. It discourages others from potentially implementing their own. Further, folks will look at the various decisions made designing said HTTP server and assume that it is how you should write Pony code. Given its performance issues, I don't think we want folks copying it.
+
+There are many things that we can say really should be in a standard library. Strings, Arrays, Maps and other data structures that we want everyone to build on to make interoperation between libraries certainly fit the bill. The argument for having an HTTP server in the standard library is much weaker.
+
+# Detailed design
+
+Move the following to their own repository "example-http-server" under the Ponylang organization.
+
+- packages/net/*
```
aturley: Should this be "packages/net/http/*"?

SeanTAllen:
I've finally updated this.  It's in a state where I think its ready for final comment and I'm marking as such.

text/0000-remove-http-server-from-stdlib.md
```diff
@@ -0,0 +1,61 @@
...
+- examples/httpserver
+
+Update examples/main.pony to not include:
+
+```
+use ex_httpget = "httpget"
+use ex_httpserver = "httpserver"
+```
+
+In the new repo, the README for the http server and client, should make its lineage known. It should indicate that they are not an exemplar of how you would write high performance code. They can be used, with caveats about API ergonomics and performance. The README should strongly indicate that the Pony core team encourages community members to create their own, better HTTP server(s) and client(s) and get them adopted by the community at large.
```
mfelsche, line 38: I would be in favor of treating the new http repository as something we want to evolve to some high-performance stuff we can be proud of. So, i would suggest put some initial issues in the new repo that list defects and make suggestions on what to improve.

The rationale behind getting a new repo is that it can evolve independently of ponyc and is not subject to RFCs anymore, so people can jump on and contribute with less pain and ceremony. I think having a README that states the goal of this becoming some high-perfomance idiomatic pony code and some issues that state how to get there, would be more encouraging.

text/0000-remove-http-server-from-stdlib.md
```diff
@@ -0,0 +1,61 @@
...
+
+CI should pass for ponyc repo after the change is made. CI should be set up for the new repo to run HTTP server tests and examples to verify that everything was correctly moved.
+
+# Drawbacks
+
+This is a breaking change for anyone using the HTTP server or client. They will have to go through a small upgrade to use Pony stable to fetch the new dependency from a new location. Additionally, package names will change as it will no longer be `net/http`. That package name makes sense in the standard library, it doesn't as a standalone.
+
+# Alternatives
+
+We could leave in the standard library and incur the drawbacks that come from people thinking these is the officially sanctioned HTTP libraries that we as a community want to go forward with.
```
mfelsche, line 57: Keeping it in the stdlib and evolve it via RFCs (in case we need to) is not an alternative?

SeanTAllen:
@mfelsche i dont think "evolving" is reasonable here. they need to be rewritten from the ground up. I spent a couple hours looking at what would need to change and the answer was basically "everything".

PR #118, opened 2018-01-13
Remove existing case functions implementation from Pony

SeanTAllen:


Theodus:
I'm in favor of this. I really like the idea of Pony having case methods, like Erlang or Elixir, but every time I have tried to use them in Pony the ergonomics quickly drew me away.

Praetonus:
This seems to be pretty uncontroversial. @SeanTAllen Should this go into final comment period?

mfelsche:
I am also in favor of this, but i would like to have this RFC explicitly list the issues with case functions, so it will be clear to anybody why exactly it has been removed. It might also help anyone who wants to reimplement this feature, what exactly should be fixed.

It should be enough to list ponyc issues if there are any.

SeanTAllen:
@mfelsche that's hard to do, its basically "they don't work" as in literally outside of a few simple examples, they do not work. there's nothing wrong with the concept of case functions but the current implementation (sugar for match functions) is buggy and error-prone.

I don't really know what to add to that.

EpicEric:
Does "Step 2" include adding a specific compiler error (such as "case functions are not supported") or not ("can't reuse name 'my-function'")?

SeanTAllen:
We discussed @EpicEric's question at sync and decided that a "this method is already defined on this class" is a fine error message.

SeanTAllen:
This passed.

SeanTAllen:
@Theodus i have a busy day, do you have time to do the bookkeeping on this?

PR #119, opened 2018-01-15
Add RFC: improved-ponybench

Theodus:
[Rendered](https://github.com/ponylang/rfcs/blob/improved-ponybench/text/0000-improved-ponybench.md)

### DIFF
--- a/text/0000-improved-ponybench.md
+++ b/text/0000-improved-ponybench.md
@@ -1,0 +1,104 @@
+- Feature Name: improved-ponybench
+- Start Date: 2018-01-15
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The contents of this RFC will present concerns with the current implementation of ponybench and provide an upgrade path based on a working prototype at https://github.com/Theodus/pony-benchmark.
+
+# Motivation
+
+Ponybench, in its current form, has flaws that make it unusable in some cases:
+- Benchmark functions cannot be garbage collected.
+- The user has no way to specify actions that should occur for setup or tear-down.
+- Asynchronous benchmarks have a tendency to move to other threads due to the use of Promises, leading to high deviation in the results.
+
+The new version of ponybench will address these concerns. New functionality will also be added to produce statistics from samples and to output the sample data in CSV format so that users may further analyze the data collected.
+
+# Detailed design
+
+The API for the `PonyBench` actor and the `MicroBenchmark` interface will resemble that of `PonyTest` and `UnitTest`, respectively. They will be defined as follows:
+
+```pony
+actor PonyBench
+  """
+  PonyBench provides a microbenchmarking framework for synchronous and
+  asynchronous operations listed by a `BenchmarkList`.
+  """
+
+  new create(env: Env, list: BenchmarkList)
+
+  be apply(bench: Benchmark)
+
+interface tag BenchmarkList
+  fun tag benchmarks(bench: PonyBench)
+
+type Benchmark is (MicroBenchmark | AsyncMicroBenchmark)
+  """
+  A benchmark defines the operation that will be run repeatedly to measure its
+  runtime over multiple samples. Each sample will run the `apply` method
+  repeatedly to minimize noise that may be observed from single runs. Setup and
+  teardown necessary for the microbenchmark may be defined in the `before` and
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@Theodus is this ready for final comment period?

text/0000-improved-ponybench.md
```diff
@@ -0,0 +1,175 @@
...
+The contents of this RFC will present concerns with the current implementation of ponybench and provide an upgrade path based on a working prototype at https://github.com/Theodus/pony-benchmark.
+
+# Motivation
+
+Ponybench, in its current form, has flaws that make it unusable in some cases:
+- Benchmark functions cannot be garbage collected.
+- The user has no way to specify actions that should occur for setup or tear-down.
+- Asynchronous benchmarks have a tendency to move to other threads due to the use of Promises, leading to high deviation in the results.
+
+The new version of ponybench will address these concerns. New functionality will also be added to produce statistics from samples and to output the sample data in CSV format so that users may further analyze the data collected.
```
dipinhora, line 17: `The new version of ponybench will address these concerns` seems to imply this new version of ponybench will ensure that the benchmark runner actor will not be moved to another scheduler thread.

How are you ensuring this?

Theodus: I don't know of any way to ensure that the user will not have other actors that will be scheduled during a benchmark, or that the runtime won't move the actor. However, async benchmarks no longer rely on promises (or any other actor) to run the benchmark. Though the user will likely have their own actors being tested in this case.

dipinhora: Makes sense. Thanks for the clarification.

Theodus:
Note: the example charts have been updated to reflect @dipinhora's improvements to avoid measurements of GC.

dipinhora:
@Theodus do you by any chance have a different example of an `AsyncMicroBenchmark`? I'm trying to imagine how it would work but it's not entirely clear to me. An `AsyncMicroBenchmark` version of `_Fib` would be very helpful.

Theodus:
@dipinhora a new example has been added to the pony-benchmark repo: https://github.com/Theodus/pony-benchmark/blob/master/examples/async/main.pony

dipinhora:
@Theodus Thanks. A couple of comments:

* It seems we have to use a `Promise` after all to run the async benchmark.
* It seems that in order to be able to async benchmark something, one has to specifically set up an actor with an `apply` function that can run the `Promise` to indicate completion.

I'm wondering how one would go about testing the performance of a HTTP get request or a TCP send where one can't modify the actors involved. I'm likely missing something simple/obvious because I'm not very familiar with `Promise`s.

Theodus:
Those points actually aren't necessary. The promise and the actor are only there because the use case for an async benchmark would be to involve other actors. An example without actors would be as follows:
```pony
actor Main is BenchmarkList
  new create(env: Env) =>
    PonyBench(env, this)

  fun tag benchmarks(bench: PonyBench) =>
    bench(_Fib(5))
    bench(_Fib(10))
    bench(_Fib(20))

class iso _Fib is AsyncMicroBenchmark
  let _n: USize
  let _name: String

  new iso create(n: USize) =>
    _n = n
    _name = "_Fib(" + n.string() + ")"

  fun name(): String =>
    _name

  fun config(): BenchConfig =>
    BenchConfig(where
      samples' = 100,
      max_sample_time' = 1_000_000)

  fun apply(c: AsyncBenchContinue) =>
    DoNotOptimise[USize](_fib(_n))
    DoNotOptimise.observe()
    c.complete()

  fun _fib(n: USize): USize =>
    if n < 2 then 1
    else _fib(n - 1) + _fib(n - 2)
    end
```

I will work on a better example, possibly with TCP

Theodus:
Here's an example benchmark for TCP connections: https://github.com/Theodus/pony-benchmark/blob/master/examples/tcp-connect/main.pony

dipinhora:
@Theodus Thanks for the examples. Much appreciated.

SeanTAllen:
This passed.

SeanTAllen:
@Theodus I have a really busy day. Do you have time to create the issue and other bookkeeping for this?

Theodus:
@SeanTAllen sure, will do.

PR #120, opened 2018-02-18
Add compile-time expressions

lukecheeseman:


SeanTAllen:
Are these different than macros in some fashion?

Theodus:
If I'm not mistaken, this related to Luke's work on simple value-dependent types: https://vimeo.com/175746403

SeanTAllen:
@Theodus Related yes, it would seem be there's a certain macro-ness to it as well. How this would play with macros (which we have said we want) is somewhat interesting.

lukecheeseman:
@SeanTAllen I wouldn't really call this a macro system as it doesn't allow developer defined mappings from inputs to outputs. It does replace parts of the AST (so in this way there is a kind of macro re-writingness to it), but this is just the implementation I have used for implementing the constant evaluation (one could imagine using a different approach). What issues are you considering when you have both a macro system and constant evaluation? I would imagine that as long as you scheduled the evaluation after the macro expansions then this should be fine (but I'm happy to be told I'm wrong :) ).

@Theodus yes this is the compile-time evaluation component of that project, the two had gotten very mixed up together and were hard to consider at the same time so I've separated this chunk out to discuss individually.

jemc:
I think this RFC could use some clarification about if FFI expressions are to be allowed to be evaluated in a compile time expression. I'd expect that this would not be allowable, but I'd like to see it explicitly mentioned here.

lukecheeseman:
Ah yes, good point. Those definitely would not be allowed.

Praetonus:
I must say that I really like this RFC and that I'm really excited about your work on this and value-dependent types being integrated to the compiler in the future.

My opinion on the various questions you've pointed out in the RFC is:

- On capabilities: I agree that making all compile-time expressions `val` would be the most sane design. Allowing "mutable" compile-time expressions sounds like it will introduce a lot of nasty edge cases.
- On the behaviour of `error`: I'm personally in favor of `error` resulting in a compilation error when evaluated in a compile-time expression.
- On floating point computations: The LLVM floating points types that we use (`float` and `double`) are guaranteed to conform to IEEE 754 representations, so I don't think we'll need special consideration for those. If it turns out I'm wrong for whatever reason and we do need platform-specific handling, LLVM's `APFloat` can offer out-of-the-box support.

I'd also like to discuss how this feature and my ongoing work on separate compilation would interact. Since separate compilation implies no access to the original source code of methods in another package, evaluating expressions by rewriting the AST would be problematic. There are various ways to solve the problem but none of them are fully satisfactory. We could:

- Store the package AST in the compiled archive. This is problematic because the AST format isn't stable, takes a lot of space, and would be quite slow to parse.
- Store LLVM bitcode in the compiled archive. This is problematic because as with our AST format, this isn't a stable format. In addition, calls to the runtime are hardcoded in the generated bitcode. This latter problem could be mitigated by using the LLVM JIT and providing hooks for the runtime functions.
- Use the final machine code of the compiled archive. The main problem here is the hardcoded runtime calls.

To truly solve this problem, I think we need a new compiler stage between the AST and the LLVM IR. We could design our own IR, similar to what Rust did with MIR. If done right, it would result in a stable, compact, fast to parse format of a package, which could be evaluated on the fly by the compiler to produce compile-time expressions.

Since my work on separate compilation is still quite far down the road I don't think we should delay this RFC because of it, but I'm interested to hear people's thoughts on that problem.

jemc:
@Praetonus - I want to better understand your goals for separate compilation.

I've been picturing your work on this feature to be aimed at optimizing pony source build times by caching the previously compiled packages, but it sounds like you also aim to have the separately-compiled package be a real "deliverable" similar to a `*.o` file is for C?

Praetonus:
@jemc Yes, I'd like to have both features.

jemc:
We discussed @Praetonus' comment on the sync call - we don't want to block this RFC by concerns of separate compilation, and we have several solution avenues to explore later when this comes up down the road.

Issue #65, opened 2016-11-02
Type alias that creates a distinct type (a la newtype).


jemc:
We'd welcome an RFC that based on the discussion that was started here: https://github.com/ponylang/ponyc/issues/317.

PR #121, opened 2018-03-30
Subtyping exclusion

Praetonus:
[Rendered](https://github.com/Praetonus/rfcs/blob/6186ade2e796decde5e8d7003fa25c9071a39fd0/text/0000-subtyping-exclusion.md).


SeanTAllen:
What is `fun foo[A: Any](a: (A | Empty val))` supposed to mean?

It can be an Any or it can be an Empty val.

So I could have

`fun foo[Empty ref](a: (Empty ref | Empty val)`

Correct me if I'm wrong here but the only reason the `| Empty val` is needed is because of the need to note that its a `val`.

My initial reaction to the RFC is "this feels hacky" but I also don't feel like a fully understand the scope of the problem so, I don't feel very comfortable commenting. 

What are the scenarios when the fact that `Empty` is already covered by `Any` isn't sufficient?

jemc:
A question:
- Are you planning to apply `nostructural` to `None` in the standard library? If so, that should be included in this RFC. Personally, I think if we make a change like this, we should add a new type (maybe `Void`?) which has this restriction, rather than applying it to an existing type, like `None`.

A thought:
- I don't really like the idea of this being an annotation. I think of annotations as being "optional" or applying tweaks at runtime, but not affecting program compilability. If it affects the type system (like this feature would), I think it should be a first-class keyword (or other syntax construct) in the language rather than an annotation.

Praetonus:
@SeanTAllen With the functionality of this RFC, `Empty` would be annotated with `nostructural` in order to make the compiler accept the match inside of `foo`. So `foo[Empty ref]` would be a compiler error, as `Empty ref` would be outside of the constraint.

For a more concrete example of this functionality, there is `_MapEmpty` and `_MapDeleted` in the `collections` package. Currently, the code in `HashMap` works because of a compiler bug with unconstrained type parameters (see [ponyc/#2182](https://github.com/ponylang/ponyc/issues/2182), which is the same bug), but it is in fact broken. If you constrain the type parameters of `HashMap` with `Any`, you'll get a compiler error on an `as` expression, which is caused by the fact that `Any` and `(_MapEmpty | _MapDeleted)` aren't disjoint. This RFC would give the framework required to fix the code in `collections`.

@jemc 

No, I don't plan to apply the annotation to `None` I plan to apply it to private marker types only, like `_MapEmpty` and `_MapDeleted`. The future RFC on relative complements would allow programmers to exclude `None` from interfaces on a case-by-case basis with an `(Any \ None)` type expression.

I think annotations are a good fit for this kind of small and localised features, since it avoids reserving another keyword or finding some new symbol. In addition, it avoids any change to the structure of the AST. What we could do is to issue a compiler error if an annotation used by the compiler is found in an unexpected place. I think it would avoid any misuse of the functionality.

jemc:
As far as syntax goes, what about something like the following?

```pony
primitive _MapEmpty isnt Any
```

The above doesn't need any new keywords, and is still very intuitive.

We avoid misuse by allowing this syntax only if the right hand side was an empty interface, like `Any`.

We already treat empty interfaces to a few special cases when it comes to tuple subtyping.

Praetonus:
@jemc I think there would be a problem here with equivalent interfaces. Suppose the user defined an `interface MyAny` in their program, which is structurally equivalent to `Any`. Should `Foo isnt Any` imply `Foo isnt MapAny`?

If no, then the interfaces aren't equivalent anymore and it causes all sort of "action at a distance" problems. So I think the only reasonable answer would be to say that yes, having an `isnt X` in the provides list of a type means that it also `isnt` every equivalent interface.

But there is a problem with that. If our type provides other interfaces or traits, then these types should be excluded from `Any` and other equivalent types. We have an "action at a distance" problem.

Since the problem exists for traits as well, this made me realise that this RFC shouldn't only be about structural subtyping. It should be about both structural and nominal subtyping, i.e. a functionality to state that a type shouldn't be a subtype of anything (except `_` probably) and is forbidden from having anything in its provides list.

The annotation for that would probably be `nosupertype` (or some more formal term, if somebody has any idea.) The `isnt` syntax is problematic as a potential candidate since I feel like it doesn't convey the functionality very well.

jemc:
> Suppose the user defined an `interface MyAny` in their program, which is structurally equivalent to `Any`. Should `Foo isnt Any` imply `Foo isnt MapAny`?

Yeah, I think it definitely should - for reasons similar to the reasons you cited in the part of your comment that I'm not quoting here. If I say my type is not a subtype of an empty interface, and an empty interface is a supertype of all types (including concrete types, interfaces, and traits), then I'm definitely intended that my type is not a subtype of *anything*.

> But there is a problem with that. If our type provides other interfaces or traits, then these types should be excluded from `Any` and other equivalent types. We have an "action at a distance" problem.

I think it's nonsensical to say `primitive MyType is MyInterface isnt Any`. So I think we can just make it a syntax error and be done with it.

That is, if you have `isnt {empty interface}` as your provides list (and we don't allow any other type to be used on the right hand side of `isnt`), then there's no valid reason to have an `is {some other type}` provides list - every case of doing so will be a contradiction, so we can just treat it as invalid syntax to have both `is` and `isnt` in your provides list, and I think we avoid all of the action at a distance issues you mentioned.

Praetonus:
This was discussed during this week's sync call. First, I'm going to update the RFC to clarify its relation with the upcoming relative complement type operator RFC.

We also discussed the syntax alternatives. The main disagreement is on the basic semantic of annotations. The parallel with C++'s attributes (which were an inspiration for Pony's annotations) was raised, and @jemc made the very valid point that given the current standard attributes in C++, removing an attribute from a C++ program won't cause a compilation error.

So the main question here would be whether annotations should be limited to things like optimisation hints (`likely`) and interactions with FFI (`packed`), or if we open up the syntax to pretty much any semantic changes, including type system stuff. I'm personally inclined to say that we shouldn't impose restrictions (excepted in cases where new syntax would incontestably be more appropriate), as I think it would be wasteful for such a versatile syntax.

Also, one of my concerns with `isnt Any` that I forgot to mention during sync is that I feel like it reverses the intent of the functionality. I.e. instead of being "I'm not a subtype of anything so I'm not a subtype of `Any`", `isnt Any` feels like saying "I'm not a subtype of `Any` so I'm not a subtype of anything".

jemc:
> So the main question here would be whether annotations should be limited to things like optimisation hints (`likely`) and interactions with FFI (`packed`), or if we open up the syntax to pretty much any semantic changes, including type system stuff.

Revising my earlier position about annotations not causing compile errors, I do agree there are good use cases such as the `nodiscard` example we discussed. Another example I was discussing with @SeanTAllen earlier this week was having an `exhaustive` annotation you could add to a `match` to instruct the compiler to raise a compile error if the match was not exhaustive.

However, I still feel nervous about using an annotation for this, because the annotation is *required* to compile (the point we discussed on the sync call, and you mentioned earlier in your comment). I feel like annotations should be able to set up *additional* checks to enforce, but removing an annotation from an example that compiles should leave you with an example that still compiles.

SeanTAllen:
I like you Joe am worried about annotations that are required for a program to pass. I have concerns about the exhaustive annotation given that if the annotation is turned off, it compiles but could result when it otherwise might not have, but that's a different conversation.

I think spending some time talking about what we all imagine annotations should be used for would be a good idea.

plietar:
An alternative solution would be to discourage or even disallow type parameters to be used in unions.

Instead of using `(A|None)`, I think it would be preferable to use `Maybe[A]`, defined as 
```pony
primitive None
class Some[A]
    let value: A
type Maybe[A] is (Some[A] | None)
```

Unfortunately this adds another layer of indirection compared to `(A|None)`. `embed` can be used instead of `let`, but this does not allow actors and interfaces to be used.

We could have another field kind (or reuse embed) which specialises after reification to either embed or let.

Or we could have first class sum types in the language, but I don't know how these would be implemented either (fat/tagged pointers maybe?)

It would be interesting to see how prevalent unions which involve type parameters are outside of the `(A|None)` case.

rurban:
The biggest drawback of this hack is that the type system becomes unsound. To allow nonstructural subtyping you have to use nominal subtyping, i.e no fields. Casting any fields to other fields is a recipe for desaster, similar to the current FFI type problems, which break type safety and memory safety.

jemc:
> To allow nonstructural subtyping

Note that, as discussed in the comment thread so far, this feature would imply that the type is not a subtype of *anything*. So there isn't anything in this RFC that would allow nonstructural subtyping.

> you have to use nominal subtyping

as implied above, it would also be impossible to use nominal subtyping (traits) with this feature, because the "is not a subtype of the empty interface" fact implies that it is not a subtype of anything that is a supertype of that interface.

>The biggest drawback of this hack is that the type system becomes unsound

If you have any particular soundness concerns you want to discuss after you update your understanding of this RFC, we'd be interested to hear them. But from reading your comment, it's not clear to me you have a full understanding of what's being discussed here, so it's difficult to engage with your concerns.

rurban:
If you read the RFC there are several grave mistakes and omissions:

"Add an annotation specifying that a given type shouldn't be a structural subtype of any other type"

Non-structural subtyping is already done with field-less classes. Such as Empty. 
Causing Empty not to subtype from Any is worthwhile fix, but not with the proposed syntax, and esp. not with the proposed wording.

The biggest problem is with that annotation you could subtype from divergent fields, breaking the type system. Thus the RFC needs to add that the `nostructural` type hint is illegal with classes with fields. 

Worse: The title and the wording.
Breaking the type system on purpose to allow weak nominal subtyping is nothing we want to even mention in any doc or type hint. The particular problem is something completely different, and should be fixed with a differently named hint. Like `empty`, but I have no better idea.

Praetonus:
@plietar That's an interesting idea but I don't think we'd be able to implement it without a runtime cost, since as you mention it it would add a layer of indirection. `embed` is also unusable since an embedded field can only be initialised from a constructor, which is problematic for collections.

@rurban It is unclear to me why you think that this feature would be problematic with types with fields since fields don't participate in subtyping resolution, only methods do. Do you have a code example of what you'd consider unsafe? Also, as mentioned in the comments above, I realised after some discussion that this feature should cover **both** structural and nominal subtyping (i.e. resulting in types that aren't subtypes of anything) but I haven't had the time to update the RFC yet.

plietar:
@Praetonus if we were to add tagged unions as a language feature I think it could be done without any additional costs in the most common cases.

Take for example:
```pony
type Maybe[A] is (Some(A) | None)
let y = Some(x)
match y
| let Some(z): Some(A) => ....
| None => ...
end
```

During codegen (and after reification), if A is disjoint from `None` the type `Maybe[Foo]` is equivalent to `(Foo | None)`. Matching is done using the type identifier embedded in the object, just like it is today.

In the case where A is `None` then an extra discriminant is required to distinguish between the A case and the `None`. For small unions one, the discriminant can be stored in the low bits of the pointer (they are usually zero because of alignment), or high bits on 64bit. For large unions, either an extra layer of indirection or two word pointers would work.

Praetonus:
@plietar That's interesting. However, since this RFC (and the upcoming one on relative complements) don't require code generation changes, I feel more comfortable moving forward with these features. Though it would be interesting to see if there are some patterns that aren't covered by my proposed feature(s) but are covered by tagged unions.

Praetonus:
I've updated and renamed the RFC based on the discussion here.

plietar:
Tagged unions would be a much more general feature. In addition to solving the same problem as this RFC, it allows differentiating variants which the user wants to be distinct even if they happen to have the same type, such as `(Foo(Baz) | Bar(Baz))`, or make `Maybe[Maybe[A]]` actually be three cases, not two.

Overall the current RFC (and the set difference constraints) feels like an odd case in the type system, and doesn't compose very well.

That being said, as you mentioned and unlike tagged unions, the solution proposed here is simple to implement, and solves a very real problem users are hitting. We could use it as a stopgap measure, but in the long run tagged unions are I think a cleaner approach.

I'm currently working on the regions RFC, but I could propose one for tagged unions which would i hope help make things clearer.

jemc:
@plietar Would tagged unions solve the use case discussed in https://github.com/ponylang/rfcs/issues/65?

SeanTAllen:
@plietar How does `Maybe` solve the "Empty" problem? Give that we want "Empty" not "None"?

Praetonus:
Are there any objections to moving this forward into final comment period?

jemc:
Based on @plietar's answers to those questions, I might want to recommend we adopt this RFC as a temporary measure and repeal it later if tagged unions form a better solution to this problem (and others).

However, I'm okay with going into final comment period.

plietar:
@jemc There's definitely some overlap between tagged unions and newtype (#65). I still need to figure out the details. Adopting this as a temporary measure sounds good to me.

@SeanTAllen I'm not sure what you mean. If we had tagged unions you wouldn't need an `Empty` type at all anymore. `Maybe` would be usable anywhere you'd expect.

----

I've realised I've not really given any arguments against this RFC,  other than "it feels odd". A big concern I have is that `Empty` would not satisfy the bound `X: Any #any`, but it would satisfy `X: X`<sup>1</sup> (the equivalent of an unbounded type variable). This used to be the case of tuples too, and it caused a lot of confusion (ponylang/ponyc#1767)

While the example from the RFC would compile:
```pony
fun foo[A: Any](a: (A | Empty val)) =>
  match consume a
  | let a': A => None
  end
```

This would not, since `A` could be `Empty`.
```pony
fun foo[A](a: (A | Empty val)) =>
  match consume a
  | let a': A => None
  end
```

Of course it can be worked around by adding an `: Any` bound, but the different behaviour between the two would be very surprising.

----

<sup>1</sup> Unless Empty is not even a subtype of itself, which would be crazy and probably break most things. (Technically subtyping isn't completely reflexive even today, but that's a weird edge case ponylang/ponyc#1810).

SeanTAllen:
@plietar why wouldn't empty be needed?

I should be able to store a None and have it not trigger being "empty".

What would be be (Some[None] | None) ?

Praetonus:
@plietar Would it make sense to modify the type system so that constraining a type parameter to itself is equivalent to constraining it to `Any #any`?

jemc:
@Praetonus I think we've discussed that before on a previous sync call, and the main thing holding us back was that tuples were not a subtype of `Any`.

Hm... actually, we might have issues with `struct`s as well, right?

Praetonus:
Structs cannot be used as type arguments (except for `Array` and some other builtin types, which are special-cased in the compiler) so I don't think it would be problematic for them.

jemc:
Yeah, as long as we extend the special case to handle the fact that types like `Array` will now have a constraint on their type parameter of `Any #any`

Praetonus:
We agreed to move this RFC to final comment period during the last sync call.

SeanTAllen:
I vote yes for this as a temporary measure.

shelby3:
@Praetonus, the link in the OP is 404 dead.

Note even though I don’t entirely understand this issue being discussed here, I can glean from the comments here that likely my [recent post](https://github.com/ponylang/ponyc/issues/2182#issuecomment-444590444) may subsume this RFC proposal. But I await further discussion and clarification to determine if I am missing some understanding.

jemc:
I've edited the OP to fix the link, thanks.

And yeah, this RFC is definitely related to the other discussion.

PR #122, opened 2018-04-20
Formal Viewpoint Adaptation (George Steed's model)

jemc:
[Rendered](https://github.com/ponylang/rfcs/blob/feature/formal-viewpoint-adaptation/text/0000-formal-viewpoint-adaptation.md).

### DIFF
--- a/text/0000-formal-viewpoint-adaptation.md
+++ b/text/0000-formal-viewpoint-adaptation.md
@@ -1,0 +1,97 @@
+- Feature Name: formal-viewpoint-adaptation
+- Start Date: 2018-04-19
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Adopt [George Steed's formal model for viewpoint adaptation][steed], including a new syntax for distinguishing between extractive and non-extractive viewpoint types.
+
+[steed]: http://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf
+
+# Motivation
+
+1. George Steed's formal model for viewpoint adaptation is the result of rigorous academic work, and provides formal rules to justify the resulting cap tables, in contrast to the current cap table we use, which is justified in an ad-hoc example-based way.
+
+2. Adopting this model will allow the compiler to be more permissive, while remaining safe. That is, this model will allow the compiler to recognize a larger subset of safe programs as being safe, and allow them to compile as such.
+
+3. Using formal rules for viewpoint-adapted capabilities will allow us to employ rule-based exhaustive testing of the capability type system for viewpoint adaptation, as we already are able to do for other areas of the capability type system.
+
+# Detailed design
+
+Currently Pony has one form of viewpoint adaptation, which is applied to test type safety of both extractive and non-extractive operations. After adopting the new model, Pony will have two distinct forms of viewpoint adaptation to represent the distinction between capability safety requirements for these two distinct kinds of operations. To quote [George Steed's paper on the subject][steed]:
+
+> [Pony as it currently exists] defines just a single operator `[operator 1]` for both reading and writing the value of fields. We now split this definition in two as this allows us more room to optimise the definition and simplify the requirements of each operator independently: we reuse the original operator `[operator 1]` as non-extracting viewpoint adaptation (the capability obtained on field read) and define `[operator 2]` as extracting viewpoint adaptation (the capability of the old value of a field or variable returned by overwriting it).
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I'm very excited to see this happen!

SeanTAllen:
Small thing to add to the RFC. Sylvan and I would like this see this implemented where we:

- Do a release before the implementation of this is merged
- Merge the implementation of this
- Release this on its own

shelby3:
I think George Steed might have an error in wording or a typo in his paper, and I can’t find an email address for him, so I will place my comments here. Hopefully one of you can forward to him if I have raised is a valid issue.

I quote from [_§2.2.1 Pony Capabilities_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=15):

> as well as allowing modifiers to denote the property that a value is unaliased by adding a caret after the capability, such as `Baz tagˆ`

But the tutorial section [Ephemeral types](https://tutorial.ponylang.io/capabilities/aliasing.html#ephemeral-types) says:

> That is, it's a type for a value that currently has no name (it might have a name through some other alias, but not the one we just consumed or destructively read).

The tutorial says the ephemerally referenced value is not always unaliased. Only the consumed (or destructively read) _named_ reference loses its alias which is transferred to the ephemeral reference. I [understand](https://github.com/ponylang/pony-tutorial/issues/322#issuecomment-445173855) that the ephemeral reference is unnamed until it is assigned to a name, but non-exclusive reference capability types may have aliases other than the ephemeral reference.

Perhaps George intended to write ‘reference’ instead of ‘value’? So maybe it’s just a typo and not an error in conceptualization? Although wouldn’t the correct way to write that be, _“…as well as allowing modifiers to denote the property that a reference is an unnamed alias by…”_ I have not completely read the paper yet, so I don’t know if I am incorrect for raising this issue or if that is just a typo or [very unlikely] if it is a conceptual error that carries through the paper.

EDIT: perhaps George meant “_mutably_ unaliased” because I read in  [_§3.3.1 Ephemeral Modifiers_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=37) that he is only interested in ephemeral modifiers on `iso` and `trn`?

---

In [_§2.2.2 Viewpoint Adaptation_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=17) the “and ω to refer to an object” doesn’t agree with Figures 1 and 2, which instead are labeled with `ι`.

---

Another typo in [_§2.4 F-Bounded Polymorphism_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=20):

> […] it is impossible to return any class more specialised that `Cloneable`.

I believe he meant “than” but that would still be misleading. As it would seem to imply that `Cloneable` is the returned class, but it’s `Object`. So better to write “than the return type in”.

---

Another typo in [_§2.6.4 Modelling Challenges_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=26):

> […] which allows expressions [with] potentially dangerous combinations of capabilities to be constructed.

Missing ‘with’ as shown.

---

Another typo in [_§3.4.3 Compatibility with Ephemeral Capabilities and Types_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=40):

> Now that we have a definition of ~~what~~[which] capabilities can co-exist with one-another […]

Incorrect word ‘what’ as shown.

---

Another typo in [_§3.11 Viewpoint Adaptation_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=47):

>  It is for this reason we now proceed [to] define viewpoint adaptation […]

Missing ‘to’ as shown.

---

Another typo in [_§3.11.2 Non-Extracting Viewpoint Adaptation_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=47):

>  Non-extracting viewpoint adaptation expresses the capability at which an actor _α_ may see through an object _ι_ to a field [object] _ι'_ […]

Missing ‘object’ as shown.

---

I’m wondering why in [_§7.6 Recovery_](https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/GeorgeSteed.pdf#page=113) to correct a bug in Pony, he chooses to not recover the capability for any of the elements of a tuple, when it seems to me that the programmer could identify one of the elements for which the capability could be safely recovered?

PR #125, opened 2018-05-23
Add operators for partial integer arithmetic

mfelsche:
[Rendered](https://github.com/ponylang/rfcs/blob/partial-arithmetic/text/0000-partial-arithmetic.md)

### DIFF
--- a/text/0000-partial-arithmetic.md
+++ b/text/0000-partial-arithmetic.md
@@ -1,0 +1,54 @@
+- Feature Name: partial_arithmetic
+- Start Date: 2018-05-23
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+This RFC suggests adding partial versions of basic Integer arithmetic operators (`+`, `-`, `*`, `/`, ...)
+that error on overflow or underflow or division by zero. These new operator should combine the base-operators with a `?`:
+`+?`, `-?`, `*?`, `/?`, ...
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


mfelsche:
I know this is kinda controversial, but i wanted to give it a shot anyway.

Fight!

winksaville:
I like the general idea of allowing the users to write safe/correct code and the consistency this proposal provides for all operators is nice.

text/0000-partial-arithmetic.md
```diff
@@ -0,0 +1,54 @@
...
+
+# How We Teach This
+
+The new methods added need to have docstrings explaining their behavior. The docstrings of the other methods should mention the other method of doing e.g. addition.
+
+The tutorial should introduce the different ways of doing arithmetic in greater detail, *checked* (`addc` etc.), *unchecked* (`add` etc.). *unsafe* ( `add_unsafe` etc.) and *partial* (`add_partial`). So that users know what options they have and what the drawbacks might be (e.g. performance against correctness).
+
+# How We Test This
+
+How do we assure that the initial implementation works? How do we assure going forward that the new functionality works after people make changes? Do we need unit tests? Something more sophisticated? What's the scope of testing? Does this change impact the testing of other parts of Pony? Is our standard CI coverage sufficient to test this change? Is manual intervention required?
```
jemc: I think this section of the RFC template still needs to be replaced with content?

mfelsche:
@jemc yeah, i forgot that section. I updated the RFC. thanks for checking.

sylvanc:
I like this. If I'm not here when a vote happens, I vote yes.

Also, in the context of how Matthias explains these operators, do we want to revisit calling the `~` operators `_unsafe` ?

codec-abc:
I would like to add that using BigInteger and some static types check may be an alternative since they can't overflow/underflow and a smartly enough compiler may be able to prove when the denominator may be different than 0 to safely allow a division. The disadvantage of this technique is that is mostly shift the problem to the point where it may be needed to convert a BigInteger to a different size with a smaller capacity.

PR #127, opened 2018-05-30
Initial draft for RFC ssl-alpn

patroclos:
[Rendered](https://github.com/patroclos/pony_rfcs/blob/ebcfc2e19c6d03ba4084ad1fcf1bec0117a118b4/text/0000-ssl-alpn.md)

### DIFF
--- a/text/0000-ssl-alpn.md
+++ b/text/0000-ssl-alpn.md
@@ -1,0 +1,173 @@
+- Feature Name: ssl-alpn
+- Start Date: 2018-05-28
+- RFC PR: 
+- Pony Issue: https://github.com/ponylang/ponyc/issues/2564
+
+# Summary
+Add server and client support for the Application-Layer Protocol Negotiation (ALPN) TLS extension to the `net/ssl` package.
+
+# Motivation
+Right now, for anybody who wants/needs to use the ALPN extension, there is no way around reimplementing or modifying the existing `net/ssl` package.
+This is a particularly nasty limitation, since pony seems to be at least somewhat affine for web technologies (going from the `HTTPServer` that is/was part of the distributed packages).
+
+And yet, without support for ALPN, there can be no viable http/2 server implementation using the
+standard packages distributed like they are today, because modern browsers depend on the ALPN extension to determine weather to use http/1.1 or http/2 defaulting to 1.1.
+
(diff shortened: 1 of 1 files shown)


mfelsche (approved):
It is not in final comment period or anything but i'd like to express my full approval anyhow.

SeanTAllen:
@patroclos this looks good. one minor nit. the issue field should be blank. at would be filled out with an issue once the RFC is accepted (we create a new issue for each accepted RFC). If you want to reference that earlier issue, the body of the RFC would be the place to do it.

text/0000-ssl-alpn.md
```diff
@@ -0,0 +1,173 @@
...
+
+```pony
+type ALPNProtocolName is String val
+primitive ALPNNoAck
+primitive ALPNWarning
+primitive ALPNFatal
+
+type ALPNMatchResult is (ALPNProtocolName | ALPNNoAck | ALPNWarning | ALPNFatal)
+
+interface ALPNProtocolResolver
```
jemc, line 51: Small nit. This could be `interface box`, since that seems to be the default usage of the type.

text/0000-ssl-alpn.md
```diff
@@ -0,0 +1,173 @@
...
+It is a bit tedious to have to implement a custom resolver every time one uses ALPN,
+although the resolver pattern helps to abstract all the pointer manipulation away from the user while staying highly flexible.
+
+To address this concern, there is an `ALPNStandardProtocolResolver` that works akin to the OpenSSL function `SSL_select_next_proto`:
+It takes a list of protocols expected by the server and tries to resolve to the first protocol in that list that is also advertised by a client.
+If no such protocol can be found, the resolver can optionally settle for the first protocol advertised by the client.
+
+Reference implementation:
+```pony
+class ALPNStandardProtocolResolver is ALPNProtocolResolver
```
jemc, line 99: Similar to my previous comment, this should probably be `class val`

sylvanc:
This is great. If I'm not at whatever sync this gets voted on, please count me as a Yes vote.

jemc:
My comments don't need to be blockers for merging the PR, since I waited so long to give them - just something to keep in mind for the implementation.

PR #129, opened 2018-09-24
Function argument alias

malthe:
Pony's shadowing prohibition means that using an object field or method name as a function argument is not allowed. This is a proposal to add an optional clause to function arguments that gives the argument a local alias.

[Rendered](https://github.com/malthe/pony-rfcs/blob/function-argument-alias/text/0000-function-argument-alias.md)

### DIFF
--- a/text/0000-function-argument-alias.md
+++ b/text/0000-function-argument-alias.md
@@ -27,7 +27,7 @@ obvious argument for that method is `name`.
 However, if we want to allow public access to that field then we're
 out of luck due to the shadowing prohibition. The only solution in
 this case is to use a different name for the function argument such as
-`new_name`, `_name` or `name'`, not of which are ideal.
+`new_name`, `_name` or `name'`, none of which are ideal.
 
 This problem applies for any function, whether it's a constructor or a
 method. It's been brought up occassionally on the mailing list with
--- a/text/0000-function-argument-alias.md
+++ b/text/0000-function-argument-alias.md
@@ -1,0 +1,120 @@
+- Feature Name: Function argument alias
+- Start Date: 2018-09-24
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Pony's shadowing prohibition means that using an object field or
+method name as a function argument is not allowed. This is a
+proposal to add an optional clause to function arguments that gives
+the argument a local alias.
+
+# Motivation
+
+Function arguments are part of the public API of a class. They appear
+in the documentation, in error messages and are used explicitly when
+calling a function using named arguments.
+
+In a variety of situations, a function argument will naturally
+correspond to a class field or method, for example when a constructor
+provides a way to directly pass an initial state, or when a method
+replaces a field with a new value. For ergonomics and ease of use,
+it's ideal to be able to use the most simple and obvious argument name
+here. That is, if a method updates the field `_name` then the most
+obvious argument for that method is `name`.
+
+However, if we want to allow public access to that field then we're
+out of luck due to the shadowing prohibition. The only solution in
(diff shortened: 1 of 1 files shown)


jemc:
We discussed this on today's sync call. We discussed a lot of different options, but @sylvanc suggested (and @kulibali, @mfelsche, and I agreed) that a good alternative to solve this set of problems is to allow locals to shadow members (fields and methods), with the understanding that members will always be accessible by prefixing with `this.`

This would make the `as` workaround unnecessary, and would solve several other problems, while still preventing the worst of shadowing issues.

I mentioned that we'd have to make sure that error messages would try to detect and suggest disambiguation issues when you forget to use `this.`

@mfelsche mentioned that it would be important to update the tutorial to reflect that change.

malthe:
That seems like a good option. The main reason I went with `as` was that the `this.p()` notation was discussed previously: https://github.com/ponylang/ponyc/issues/305#issuecomment-141509314.

But I think `this.p()` ultimately is the better compromise. It's less verbose and allows writing clearer code, albeit with the risk of `p` having that ambiguity in what it actually refers to.

mfelsche:
Could we restrict the rule mentioned above a bit more and say, that only parameters should be able to shadow members, not `let` or `var` variables?

I have to admit, having slept a night over it, i still have mixed feelings about it.

SeanTAllen:
I'm really uneasy about allowing any sort of shadowing. The only real time I have ever felt the pain of this is when using "where = BLAH" in a function calling and not being able to use the name I wanted because it was the same as a public field of the object.

I think malthe's idea from my understanding addresses this one pain point. I don't like the idea discussed on the sync call of allowing locals to shadow fields and access via `this`.

EpicEric:
I personally prefer an explicit solution (i.e. the proposed RFC) than an implicit one that might lead to refactoring bugs.

That said, I am in favour of this RFC, as it'd be helpful specially for named arguments on a constructor, avoiding using primes both on the API and in the actor/class fields.

chalcolith:
I am generally not in favour of this RFC, as I think the use of `as` is confusing.  I would prefer only relaxing shadowing for method names, and perhaps even only in constructors.

The only annoyance I've found with Pony's shadowing rules is when I want to have a parameter in a constructor be the same as a getter method.

jemc:
I'm in general not in favor of this RFC as currently written. I'm mentioned in the text of the RFC as recommending that primes be optional when invoking with named arguments, and I still believe this, as I agree with @SeanTAllen that this is the biggest problem to solve here. The RFC points out that this wouldn't solve the problem of primes showing up in documentation and such, but I don't really see that as being a big issue.

I find the `as` syntax in this RFC to be a really verbose way of solving a problem that is mostly about convenience for the caller. Inconveniencing the function author via verbosity for the convenience of the caller doesn't seem like a great solution, and because it's optional, it's likely that function authors won't bother to do it and the callers will still be left with using primes in their named arguments.

jemc:
This was discussed in depth on today's sync call. We're not in favor of this approach to solving the problem, for the reasons already discussed. There are a few alternatives that we discussed that remain on the table for further discussion:
- Solve the issue by removing named arguments and adding something like named tuples or other value types, which would probably be a great feature for the language, but there's a lot that needs to be thought through with this approach and it's unlikely to land soon.
- Solve the issue by allowing a field name to be used as a parameter name, which will implicitly assign the value to that field at the start of the method body.
- Solve the issue by allowing the prime to be optionally omitted from the call site when using named arguments. @SeanTAllen is not in favor of the inconsistency this would bring.
- Do nothing to solve this issue.

PR #130, opened 2018-10-16
Create a Graph Implementation RFC

SCKelemen:
This RFC defines data structures for Graphs, and will form the basis for many graph algorithms. 

This RFC addresses [ponyc issue 2905](https://github.com/ponylang/ponyc/issues/2905).

### DIFF
--- a/text/0000-Std-Graph-Impl.md
+++ b/text/0000-Std-Graph-Impl.md
@@ -1,0 +1,74 @@
+- Feature Name: Graph Implementation
+- Start Date: 2018-10-16
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Create an implementation of a graph. A Graph, in Graph Theory, is defined as an object containing two sets, a vertex set and an edge set. This RFC currently only contains the data structures. Algorithms will be added once a structure is formalised.
+
+# Motivation
+
+Graphs are useful to a whole host of interests in mathematics and have many applications. Graphs could be useful in implementing other algorithms and data structures, and could be useful in representing networks.
+Since graphs have a basis in many fields, it makes sense to have a standard implementation so that future third-party libraries can't easily interact with eachother. 
+
+# Detailed design
+
+Three classes will need to exist:
+ - Graph
+ - Vertex
+ - Edge 
+
+ As previously mentioned, a Graph is defined as an object containing two set: A Vertex set, and an Edge set. 
+ Thus, we can define a Graph in the following pseudo code:
+  ```
+    class Graph {
+      Set<Vertex> Vertices
+      Set<Edge> Edges
+    }
+    
+  ```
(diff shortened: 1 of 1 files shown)


SCKelemen:
I have created a Set RFC, which will provide the base set for the Graph implementation. #131

text/0000-Std-Graph-Impl.md
```diff
@@ -0,0 +1,82 @@
...
+ As previously mentioned, a Graph is defined as an object containing two set: A Vertex set, and an Edge set. 
+ Thus, we can define a Graph in the following pseudo code:
+  ```
+    class Graph {
+      Set<Vertex> Vertices
+      Set<Edge> Edges
+    }
+    
+  ```
+  *note: Set is used in the Set theory sense, meaning a collection of objects, not a HashSet.
```
mfelsche, line 31: why not a HashSet?

text/0000-Std-Graph-Impl.md
```diff
@@ -0,0 +1,82 @@
...
+      Vertex Source
+      Vertex Destination
+    }
+  ```
+  An Edge can be undirected or directed, and can be weighted or unweighted. 
+
+# How We Teach This
+
+In Graph Theory, the entity of the graph is known as a Vertex. LinkedLists, Trees, Tries, etc are all examples of a graph. In these domain examples from computer science, the entities are often known as Nodes. `Node` could be an alternative name to Vertex.
+This feature add only to the math standard library. It does not add to the learning curve of Pony, and does add any increased complexity to the language or compiler. It's simply a library that can be used at the descretion of the developer.
```
Clyybber, line 52: I think you meant "... and does not add any increased complexity ..."

text/0000-Std-Graph-Impl.md
SCKelemen: Depending on the implementation, it could be a HashSet. I want to separate the theoretical abstract idea of the set from it's ponyc implementation in this RFC.

text/0000-Std-Graph-Impl.md
SCKelemen: I sure did. Thanks, I'll update the PR.

SeanTAllen:
Hey @SCKelemen, 

Just a note. Thanks for opening the RFC. Seeing new folks come in with good RFCs is one of the the things that keeps me motivated and working on Pony.

-S-

text/0000-Std-Graph-Impl.md
SeanTAllen: Nice.

SeanTAllen:
I think this makes more sense to start as a 3rd party library where it can be evolved without the constraints of the RFC process and then once its more mature, adding it to the standard library via the RFC process.

SCKelemen:
I'm okay with that. Eventually, I'd hope to have more language constructs and syntactic sugars for working with such structures, but at the moment, I don't think I have enough experience with the Pony to support this in the std lib anyway.

PR #131, opened 2018-10-17
Standard Lib: Math/Set

SCKelemen:
This RFC contains details regarding an implementation of a Set as it pertains to Set Theory, and a few methods related to sets. This RFC is basis for the Graph Implementation in RFC #130 .

### DIFF
--- a/text/0000-Std-Set-Impl.md
+++ b/text/0000-Std-Set-Impl.md
@@ -160,34 +160,100 @@ In the case that two sets share no common elements, then the intersection of the
 
 With these formal definitions, we have sufficient information to begin describing the implementation.
 
+## Data Structures
+
+Set:
+```pony
+  class Set
+    """
+     Docs
+    """
+    create // todo
+```
+
+Empty Set:
+```
+  static Set Empty = Set{ } 
+```
+## Methods
+
+Equality:
+```
+  Bool is_equal(set: Set)
+```
+
+Cardinality:
+```
+  uint cardinality()
+```
+
+Equivelance:
+```
+  Bool is_equivalent(set: Set)
+```
+
+Subsets:
+```
+  Bool is_subset(set: Set)
+  Bool is_proper_subset(set: Set)
+```
+
+Supersets: 
+```
+  Bool is_superset(set: Set)
+  Bool is_proper_superset(set: Set)
+```
+
+Empty:
+```
+  Bool is_empty_set()
+```
+
+Complementary:
+```
+  Set get_complementary_set(set: Set)
+  Bool is_complementary_set(set: Set, complement: Set)
+```
+
+Unions:
+```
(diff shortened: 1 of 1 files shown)


text/0000-Std-Set-Impl.md
```diff
@@ -0,0 +1,264 @@
...
+
+# Drawbacks
+Adding this to the standard library will have some maintainence cost associated. 
+
+# Alternatives
+Some alternatives could be other naming conventions. Some Set Theory terminology can seem quite nuanced to novices, especially with terms like "Equal" and "Equivalent" having very different meanings. Another alternative would be completely leaving this structures and methods out of the standard library and using it as a free-standing thrid-party library.
+
+# Unresolved questions
+What is the most idiomatic way to structure this class?
+Can operators be overloaded?
```
mfelsche, line 264: The existing operators can be overloaded. See https://tutorial.ponylang.io/expressions/infix-ops.html

mfelsche:
There are several questions this RFC should answer:

How would such a `Set` as you proposed differ from [`collections.HashSet`](https://stdlib.ponylang.io/collections-HashSet/) and [`collections.persistent.HashSet`](https://stdlib.ponylang.io/collections-persistent-HashSet/)?

Wouldn't these existing sets suffice for the use case you imagine? If not why?

Do you propose a new syntax for constructing a set using curly braces?

SCKelemen:
Set.Set could be implemented with  collections.HashSet or collections.persistent.HashSet, in which case this would define a few extension methods. Some of the methods I defined do not have corresponding implementation in those libraries, such as Equivalent and Complementary sets, and `properness` of subsets and supersets. 

I think the curly brace syntax would be a nice collection initializer approach. I am not sure that such a language change should not have its own RFC, though.

SeanTAllen:
> I think the curly brace syntax would be a nice collection initializer approach. I am not sure that such a language change should not have its own RFC, though.

I think that should have its own RFC.

SCKelemen:
> > I think the curly brace syntax would be a nice collection initializer approach. I am not sure that such a language change should not have its own RFC, though.
> 
> I think that should have its own RFC.

I agree. I can go ahead and write that RFC, if you think it make sense for this language.

SeanTAllen:
@SCKelemen I think it would make sense to start with this as an external library, outside the stdlib and eventually bring it in. That would allow for a lot more flexibility changing the library and evolving it. Changes in the standard library have to go through the RFC process. In general, I'm in favor of libraries starting outside the standard library, getting to some level of maturity, and then being brought in.

SCKelemen:
I'm okay with that. Eventually, I'd hope to have more language constructs and syntactic sugars for working with such structures, but at the moment, I don't think I have enough experience with the Pony to support this in the std lib anyway.

PR #135, opened 2018-11-12
Add RFC for modulo operator

mfelsche:
and connected floored division for all integer types.

[Rendered](https://github.com/ponylang/rfcs/blob/mod-operator/text/0000-modulo-operator.md)

### DIFF
--- a/text/0000-modulo-operator.md
+++ b/text/0000-modulo-operator.md
@@ -57,9 +57,11 @@ The `mod` function will be defined on all integer and floating point types and w
 
 ## Integers
 
+For Integer types this operator will have the partial and unsafe variants: `%%?` and `%%~`. There will also be a `modc` function being added, returning a flag for overflow/underflow and the actual result.
+
 `mod` will be defined as the remainder from floored division (rounded towards negative inifinity).
 
-As the current integer division operation is truncated towards zero, an additional floored division function needs to be defined on all integer types. This one will be called `floored_div`, a short version `fld` or `fldiv` is also possible. This function will have no operator symbol mapping to it. For unsigned integer types, floored division will be the same as `div`. For signed integer types it needs to be implemented differently.
+As the current integer division operation is truncated towards zero, an additional floored division function needs to be defined on all integer types. This one will be called `fld`. This function will have no operator symbol mapping to it. For unsigned integer types, floored division will be the same as `div`. For signed integer types it needs to be implemented differently. To stay consistent there will also be `fld_unsafe`, `fld_partial` and `fldc`.
 
 Floored division for signed integers will be implemented as follows:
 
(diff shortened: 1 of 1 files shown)


text/0000-modulo-operator.md
```diff
@@ -0,0 +1,109 @@
+- Feature Name: modulo_operator
+- Start Date: 2018-11-09
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add a modulo operator `%%` for integer types that is defined as the remainder after floor division. Most importantly the result always has the sign of the Divisor, in contrast to `rem` (`%` operator), whose results always have the result of the dividend.
```
jemc: Small typo:

> whose results always have the result of the dividend

Should be:

> whose results always have the *sign* of the dividend

aturley:
@malthe are you ready for final comment period on this?

malthe:
I think you mean @mfelsche

mfelsche:
I am, lets do this

jemc:
For reference, see also:
https://github.com/ponylang/ponyc/pull/2888
https://github.com/ponylang/pony-tutorial/pull/323

(in which `%` was renamed as `rem` from previously being named `mod`)

SeanTAllen:
I think this RFC needs to define what happens for floating point. For example, `rem` exists as a method on F64. Will `mod` exist on F64? Will `%%` work on it?

See https://playground.ponylang.io/?gist=0f84691aa4de5367ff7cb7021a9a8cec to see that `rem` works on F64 now.

I think this RFC makes sense but we would 

- want floating point and integer to have the same operations available
- need to define for `%%` and `mod` what happens with floating point operations.

From sync, Sylvan suggests that for floating point we should use divisor.

mfelsche:
Yes, this RFC totally needs to also speak about floats, i totally overlooked them also having a rem method. Gonna add that within the next days

aturley:
Removing "final comment period" tag so @mfelsche can add stuff about floats.

mfelsche:
Stuff about floats has been added, namely: mod and fld (floored division) will also be added to floating point types with mod being the remainder after division with the same aign as the divisor and without truncation (except normal IEEE floating point truncation).

I think, thus that this is read for final comment period.

jemc:
@mfelsche - did you forget to push the commit?

mfelsche:
Yep, sorry, gonna push tonight. Have been in a rush...

mfelsche:
pushed.

jemc:
:+1:

aturley:
@SeanTAllen do you have any input on this?

skull-squadron:
Will there be `%%?` and `%%~` to stay consistent?

mfelsche:
Yes, i will add those. Gonna update this RFC to make this clear.

jemc:
Thanks @mfelsche!

PR #138, opened 2019-01-11
Remove nominal typing from Pony

SeanTAllen:


jemc:
This has been discussed on several sync calls.

I am pretty much apathetic about this change. I don't really think that having `trait`s in the language causes any major problems, but I also don't think that *removing* them would cause any major problems either, so if others want to remove them in the interest of simplicity, I am fine with it.

:+1:

text/0000-remove-traits.md
```diff
@@ -0,0 +1,71 @@
...
+* format/prefix_spec.pony
+* logger/_test.pony
+* ponybench/_runner.pony
+* ponybench/benchmark.pony
+* ponytest/_group.pony
+* ponytest/test_list.pony
+* ponytest/unit_test.pony
+* random/random.pony
+
+Additionally, we need to update tutorial, website, and patterns sites to use interface instead of tutorial.
```
cquinn: I think you mean "instead of trait."

text/0000-remove-traits.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Motivation
+
+As it currently stands with Pony. Traits and Interfaces are almost identical.  There's nothing you can do with one that you can't do with the other.
+
+New users coming to Pony often end up in IRC asking what the difference between the two are. We generally explain that traits are nominal subtyping and interfaces are structural subtyping and this is mostly true except, that you use the `is` keyword when implementing an interface to get compile time checks to make sure you are implementing said interface.
+
+All in all, its rather confusing. 
+
+Where possible, we should be aiming to provide the minimum number of concepts in Pony to allow a user to solve the problem they have at hand. There is no problem that traits solves that can not be solved with interfaces. There are however things you can do with structual subtyping and interfaces that you can't do with traits.
```
blitzerr, line 18: For completeness of the RFC, can we mention a couple of examples attributing to the fact that "there are however things you can do with interfaces that one can't with traits" ?

malthe:
A drawback would be, as the tutorial points out, "But traits are a powerful tool as well: they stop accidental subtyping."

SeanTAllen:
Im going to be opening up a 2nd RFC to go along with this one to add a facility for doing enums in Pony (a common usage of traits in Pony) without code reuse aspects that traits have.

SeanTAllen:
Just a note that I need to update this.

text/0000-remove-traits.md
SeanTAllen: updated.

text/0000-remove-traits.md
SeanTAllen: @blitzerr  i've expanded upon that.

jtfmumm:
If we decide to remove nominal subtyping from Pony, I’d like to suggest that we keep the name `trait` for what we currently call `interface`.  In general, “interface” indicates a logical specification of a module, and can be used in a variety of contexts in software engineering. The well-known counterpart to “interface” is “implementation”, so there's plausibly a background assumption that interfaces are logical specifications that require implementation. In Pony, both `trait` and `interface` allow default implementations of methods. In my view, this much more closely tracks the meaning of “trait” than “interface” (both in the way those terms are used in programming-language-neutral conversations and the way they have been used as keywords in programming languages).

First, consider the way the `interface` keyword is used in other languages.  Go interfaces do use structural subtyping, so in that respect they’re similar to Pony’s, but they notably do not permit default/partial implementations. This point about default implementations was true of Java interfaces as well until the “defender method” concept was introduced in Java 8.  But in that case, you still use a `default` keyword on methods with default implementations, indicating that it is nonstandard. I've also been reading through discussions about default method implementations for the interface-like construct in various languages, and found that it was not uncommon to use the terms “trait” and “trait-like” to refer to the property of allowing (or emulating) default method implementations.

Second, consider the way the keyword `trait` is used in other programming languages.  In both Scala and Rust, the interface-like construct is called “trait”.  And in both these cases, traits allow default implementations of methods.  In early Scala documentation, this was explicitly called out as marking a distinction from Java interfaces where default implementations were not allowed.  And at least according to some informal discussions (note that I’m not sure if this is true!), there used to be something called “interface” in Rust that was superseded by “trait” in large part for these very reasons.

If you go back to the [paper](http://scg.unibe.ch/archive/papers/Scha03aTraits.pdf) introducing the “trait” concept, you’ll find that a trait includes a combination of already implemented methods and so-called “required” methods (i.e. methods without default implementations that must be implemented). The paper suggests that classes can be composed from traits, and that traits can be composed from other traits.  Both of these properties are in line with Scala and Rust traits, as well as both Pony concepts of `interface` and `trait`.  There is some mismatch here, since in the paper it seems to be assumed that a trait would always implement at least some behavior, but I think the similarities to the Pony concept currently called `interface` (as well as the similaries to `trait` in Rust and Scala, and dissimilarities to `interface` in Java and Go, among other languages) outweigh this mismatch. A Pony `interface` can be used both as a purely logical specification and as a trait with default implementations. I think the latter point is the more interesting one from the standpoint of understanding how it can be used (and placing it in the context of how these concepts are used in other programming languages).

I understand that at first this switch would be confusing for current Pony programmers, but as someone who has written Pony code regularly for a few years now, I don’t personally think it will involve too much of an adjustment. After all, each user has probably favored one over the other, and so some of us will have to make an adjustment in either case.  On the plus side, a global replace of `interface` with `trait` should work seamlessly as far as I can tell in any Pony codebase (ignoring the fact that someone might use “interface” in comments to refer to a logical specification, a fact that I feel wouldn’t actually hurt my overall argument here).

jemc:
Agreed with the points raised by @jtfmumm.

Additionally, I'll add that `trait` is fewer letters to type and also, being a five-letter keyword like `class` and `actor`, it aligns aesthetically with the nested three-letter keywords like `fun` and `new`:
```pony
trait val Trait
  fun val method()

aturley:
I think @jtfmumm's points are good. I agree that what we have in Pony could more properly be called a "trait" than an "interface.

My biggest concern is that there are existing blog posts and other documents that may never get updated and will float around forever, potentially confusing people. But hey, maybe that's a good excuse for everybody to write a new blog post about the move from "interface" to "trait".

SeanTAllen:
I am going to update this to include @jtfmumm suggestion. We will be removing nominal typing from Pony, but what remains will be renamed from interface to trait.

Additionally, we will do a check in the compiler for `is BLAH` where it will error if `trait BLAH` has no methods as it is meaningless and not what someone would want. That check will catch any usage of trait for "open enumerations" which is no longer a supported affordance.

SeanTAllen:
I'm closing this because I'm no longer in favor of this.

I've spent the last year considering something more along the lines of "only traits can have default implementations". Currently to get a default implementation from an interface, you have to use `is` and therefore nominal typing already.

I will probably open an RFC for that change. It would allow for both `trait` and `interface` but only `trait` can appear in a `class Foo is SOMETHING` and only `trait` could have default methods.

PR #142, opened 2019-01-31
Standard lib prime proposal

adri326:
This is a proposal to add some prime number utilities in the `math` package of the standard library. A draft can be found at [pony-primes](https://github.com/adri326/pony-primes)

### DIFF
--- a/text/0000-stdlib-prime-proposal.md
+++ b/text/0000-stdlib-prime-proposal.md
@@ -1,0 +1,144 @@
+- Feature Name: stdlib-prime-proposal
+- Start Date: 2019-01-31
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add some prime maths to the standard library, within the `math` package: primality and coprimality tests, prime factorization, prime iteration and GCD/LCM.
+
+# Motivation
+
+As the standard library entry for the `math` package states, this package needs to be populated with functionalities.
+My proposal would be to have prime utilities added to it, as I believe it would fit nicely with the current `fibonacci` class.
+
+There are many potential use cases, they show a lot of interesting characteristics, like their chaotic patterns.
+This feature may also be used for benchmarking, to show how fast Pony crunches down numbers into other numbers.
+
+# Detailed design
+
+A draft/proposal of the feature is up at [pony-primes](https://github.com/adri326/pony-primes).
+Most of the features will be contained in a `Prime` primitive, with the only exception of the prime iterator, which has its own class, `PrimeIterator`.
+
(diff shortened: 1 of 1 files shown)


text/0000-stdlib-prime-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+As this utilities will be in the `math` package for the standard library, their explanations should be done within the `stdlib` documentation.
+
+Mathematical yet easy to understand and to apprehend terminology should be used.
+Words like `coprimes`, `gcm` and `lcm` could have links to articles explaining them.
+
+If a tutorial for the `math` package comes up, this feature could be introduced or evoqued, but I do not feel there is a need to.
+
+# How We Test This
+
+Unit testing can easily be implemented: a good part of them can already be found in the `/test` directory of [pony-primes](https://github.com/adri326/pony-primes).
```
SeanTAllen, line 126: So, if I understand, pony-primes is an example of what would be done but more work would be required for inclusion in the standard library? For example, additional tests?

text/0000-stdlib-prime-proposal.md
```diff
@@ -0,0 +1,144 @@
...
+The algorithms used here are common and so can easily be fixed if they happen to break.
+
+None of the code here have an impact on the other parts of the stdlib and only the `builtin` library is used.
+
+# Drawbacks
+
+This code would sit next to fibonacci's code and should not require much maintenance (the fibonacci code hasn't been updated in 2 years).
+
+However, this feature is not important an important one, and only implements common algorithms which are rarely implemented in other languages' standard library.
+It could thus find a better home in `main.actor`.
```
SeanTAllen: You can update this to say that it could remain as an independent library. `main.actor` is going to be a documentation hosting site and package discovery site.

text/0000-stdlib-prime-proposal.md
adri326: Yes, it is an implementation of what I described above. If the RFC was to be accepted without any modifications needed, then we could move code from this repository to the `ponyc` repository.
More work might be needed, this is one of the unresolved questions.

text/0000-stdlib-prime-proposal.md
adri326: Done!

text/0000-stdlib-prime-proposal.md
SeanTAllen: I think pony-primes should be complete in terms of tests before inclusion in the standard library. I'd prefer to not accept the RFC until "pony-primes" library is "done and ready for inclusion".

adri326:
This has kind of died off. I shall let this stay as an external library, while we prepare for the new additions to the math stdlib.

PR #144, opened 2019-03-02
Proposal to access environment variables and values via a Map

joncfoo:
This proposal aims to introduce a method of accessing environment variables and their values via a `Map`.

### DIFF
--- a/text/0000-env-vars-map.md
+++ b/text/0000-env-vars-map.md
@@ -1,0 +1,57 @@
+- Feature Name: env-vars-map
+- Start Date: 2019-03-02
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Environment variables corresponding values should be exposed via a `Map` in
+`Env`.
+
+# Motivation
+
+Just about every programming supports accessing environment variables via some
+method of lookup - either via a function or a data structure.
+
+Currently in Pony, the program's environment is provided as `Array[String] val`
+via `Env.vars` on which the user has to iterate and perform some string
+operations in order to a) check if an environment variable was specificied and
+b) check its corresponding value.
+
+The current method of checking and retrieving an environment variable's value is
+duplicated in every Pony program that requires such functionality. Let's move
+this into the standard library so users can easily access environment variables
+and their values.
+
+# Detailed design
+
+`Env` would be modified to have a public field `vars_map: Map[String, String]
+val` and in constructors this map would be created to reflect the key-value
+pairs present in `vars: Array[String] val`.
+
+# How We Teach This
+
+This new addition can be publicized in the next release's changelog.
+
+# How We Test This
+
+A code review should be sufficient as the implementation is trivial.
+
+However, tests can be written to ensure that the contents of `Env.vars`' are
+indeed present in `Env.vars_map`.
+
+# Drawbacks
+
+Introducing this change means users now have two ways to access environment
+variables.
+
+# Alternatives
+
+The alternative is to not implement this.
+
+# Unresolved questions
+
+As mentioned in [#Drawbacks](Drawbacks) there would now be two ways of accessing
+environment variables which seems questionable. Perhaps a future RFC can
+introduce the deprecation of `Env.vars` or perhaps even its removal so that
+there is only one simple way of accessing environment variables.
--- a/text/0000-env-vars-map.md
+++ b/text/0000-env-vars-map.md
@@ -29,29 +29,46 @@ and their values.
 val` and in constructors this map would be created to reflect the key-value
 pairs present in `vars: Array[String] val`.
 
+Additionally, in order to support `vars_map`, `Map` and its dependencies would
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@joncfoo `Env.vars` is an array because everything in the builtin package can only use either things in the builtin package. For this to work, `Map` would have to be moved into the builtin package. Bringing in `Map` from collections, is probably going to bring a lot of other stuff with it.

There might be another solution other than pulling `Map` and its dependencies into `builtin`, but that is the most "straightforward" way to do this. I think you should note this as 1 possibility in the RFC and layout what needs to move from collections into `builtin` if we went that route.

SeanTAllen:
It looks like `map.pony` and `hashable.pony` would need to move into builtin. But don't quote me on that.

SeanTAllen:
I think the existence of 

```pony
primitive HashByteSeq is
  (HashFunction[ByteSeq box] & HashFunction64[ByteSeq box])
  """
  Hash and equality functions for arbitrary ByteSeq.
  """
  fun hash(x: ByteSeq box): USize =>
    @ponyint_hash_block[USize](x.cpointer(), x.size())

  fun hash64(x: ByteSeq box): U64 =>
    @ponyint_hash_block64[U64](x.cpointer(), x.size())

  fun eq(x: ByteSeq box, y: ByteSeq box): Bool =>
    if x.size() == y.size() then
      @memcmp[I32](x.cpointer(), y.cpointer(), x.size()) == 0
    else
      false
    end
```

in `hashable.pony` is a fairly strong argument that `hashable.pony` should be in `builtin`.

joncfoo:
I've updated the text to point out the dependency issues.

mfelsche:
It feels strange to just move `Map` into builtin and leave all other collections aside, just for having more convenient access to the env-vars. Although i agree the access to the env vars should be more convenient as it is now.

Another alternative comes to mind: Add a new package, and move `cli.EnvVars` to it. Access to environment variables is nothing `cli` specific imho.

SeanTAllen:
@mfelsche Array which is a collection is already part of builtin.

jemc:
This was discussed on today's sync call, in which @sylvanc shared some relevant historical info that led us to suggest an alternative to this RFC.

At one time in Pony history, Map used to be a part of builtin for this very reason, and it turned out that the often very large number of environment variables present on most systems was dominating the performance of some Pony programs, and we had the undesirable situation of the number of environment variables changing the runtime performance of the application. Due to that problem, which also occurred with similar effects on a smaller scale for Array, env vars were changed to be lazily evaluated in https://github.com/ponylang/ponyc/commit/dedf0702954ca453cc30f0b0a41aecc53aab66ec.

Later on, in https://github.com/ponylang/ponyc/commit/b1bf144cc5c84751556de454f5ba3261c2f64f2f, this was reverted back to being an array, which brought back some of this same issue.

What we discussed doing on the call was to:
- bring it back to being fully lazy - don't precalculate a Map or an Array, and keep it internally as a pointer
- provide a method that returns an `Iterator[(String, String)]` based on the pointer
- provide a "find env var by name" convenience method that returns the value of a given env var by iterating over the pointer until it is found (or raises an error if not found).
- keep support for artificially-constructed `Env`s by having the env vars be stored as a union type of the pointer format and the array format, using the pointer format for "real" `Env`s and the array format for artificial `Env`s.

While the "find env var by name" convenience method won't have the lookup performance that a Map would, the laziness of it avoids the runtime cost of building the map eagerly. The assumption is that Env var access is to be relatively simple, sparse, and not in the hot path. In cases where you need to do something more complicated or performant, you can use the `Iterator[(String, String)]` to build a Map or persistent Map or whatever other kind of data structure you need, but at least we have covered the basic convenience case of "give me the value of a particular environment variable" without going to that level of user-side complexity.

After some discussion, we want to avoid bringing Map into builtin for a few reasons:
- `builtin` is ideally meant to be the minimal set of types required by the compiler or to start the program
- having a `Map` in `builtin` would be a breaking change for the persistent Map data structure - we'd have to call it something else because `builtin`'s `Map` would always be loaded and no type with the same name could be defined.

mfelsche:
Should the RFC now be reformulated? Or do we do the changes discussed in the sync as is in ponyc? In this case we should add them as issues there.

jemc:
@mfelsche - we discussed in sync that we could probably go direct to issues in ponyc repo, deprecating this RFC in favor of that.

jemc:
There's an issue with the plan discussed above:
> keep support for artificially-constructed Envs by having the env vars be stored as a union type of the pointer format and the array format, using the pointer format for "real" Envs and the array format for artificial Envs.

A `Pointer` can't be used in a union type. How do we want to proceed?

mfelsche:
As the Env created from an array is only used for testing, would it make sense to transform it into the envp format: Pointer[Pointer[U8]] ? No union needed in that case. Would we get into trouble because the arrays underlying the strings we point to might be garbage collected?

jemc:
That should work.

PR #147, opened 2019-04-01
added range_redesign rfc

chobeat:
RFC for the range redesign. All the details are in text/0000_range_redesign.md


SeanTAllen:
Markdown formatting seems a little off based on the highlighting that GitHub is doing

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+A Range reimplementation that includes the features from collections.Reverse together with a more comprehensive and coherent API. A reference implementation can be found here: https://github.com/chobeat/pony-range-proposal
+
+# Motivation
+
+The existence of Range and Reverse as separate classes, each one with their own limitations, to cover a use case that in most languages is addressed by a single API highlights the possibility for an improvement. Having a single, well-tested implementation to generate numeric ranges will improve the quality of life of new developers and possibly reduce unexpected behaviors for a core piece of the stdlib.
```
SeanTAllen, line 12: Can you elaborate on what the limitations of the current APIs are? It's unclear to me what specific problems would be solved. There's a short "for example". Current vs New API examples would be helpful here.

SeanTAllen:
Looking at your reference code, its more explicit about inclusive and exclusive but reverse is now implicit. The reverse API is non-obvious in that fashion to me. I think this should be addressed in the RFC.

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+
+In addition to that, the current implementation doesn't allow to express a range spanning over a whole data type since it makes assumptions on the exclusiveness of the right bound.
+
+# Detailed design
+
+The new implementation should fullfil the following design principles: 
+
+* be able to generate all the possible ranges, increasing and decreasing, for all the data types, without incurring in overflow errors.
+* offer a unified, usable, coherent API with the possibility to specify inclusive or exclusive bounds. To achieve this, the definition of the bounds and the range should be independent. The Range class should offer helper methods to cover the most common cases without explicitely instancing the bounds.
+* never raise errors but instead return an empty list. This is done to make the class more usable in day to day usage. A version of the class with explicit errors can be considered.
```
SeanTAllen: can you elaborate on why this is a design goal?

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+The existence of Range and Reverse as separate classes, each one with their own limitations, to cover a use case that in most languages is addressed by a single API highlights the possibility for an improvement. Having a single, well-tested implementation to generate numeric ranges will improve the quality of life of new developers and possibly reduce unexpected behaviors for a core piece of the stdlib.
+For example in the existing implementation it is possible to generate infinite ranges without any explicit control. 
+
+In addition to that, the current implementation doesn't allow to express a range spanning over a whole data type since it makes assumptions on the exclusiveness of the right bound.
+
+# Detailed design
+
+The new implementation should fullfil the following design principles: 
+
+* be able to generate all the possible ranges, increasing and decreasing, for all the data types, without incurring in overflow errors.
```
SeanTAllen, line 21: how does the reference implementation achieve this goal?

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+* offer a unified, usable, coherent API with the possibility to specify inclusive or exclusive bounds. To achieve this, the definition of the bounds and the range should be independent. The Range class should offer helper methods to cover the most common cases without explicitely instancing the bounds.
+* never raise errors but instead return an empty list. This is done to make the class more usable in day to day usage. A version of the class with explicit errors can be considered.
+
+The reference implementation solves these problems in the following way:
+
+* implements a trait `Bound[T]`, implemented by two classes: `Inclusive[T]` and `Exclusive[T]` to represent the bounds. Each one of them will be able to return the value to be used to define the range in the case that they are used as an upper or lower bound.
+* implements a class `Range` with a default constructor that respects the existing API
+* implements a more flexible constructor with the following signature: `
```
SeanTAllen: the backticks here breaks github formatting and appears to need to be on the same line when using single backticks.

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+  new define(b: Bound[T], e: Bound[T], step: T=1)`. 
+* Defines Step so that it cannot be negative or zero. In that case the range is considered empty. 
+**** Supports the notion of range direction. If b<e the range is defined as forward, otherwise backward. If b==e the range is empty.
+* implements two additional constructors `incl` and `to`, to support the case where both bounds are inclusive and the case where the range starts at 0. Additional constructors can be considered.
```
SeanTAllen: can you give an example of this case?

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+
+# How We Teach This
+
+The API of the `Range` class is rather similar to the API of the `Range` implementations in many other languages. Explaining the basic behavior should be straightforward. The Bounds logic might deserve a dedicated explanation since it's necessary to use the default constructor and might be unexpected.
```
SeanTAllen: We need to have specific "how we teach" steps that can be used as part of "is all the work done for this" on merge of the implementation. For example, is this purely documentation as part of the stdlib? Is there anything else where this needs to be documentated? 

What should be part of the documentation for the standard lib documentation for it to be accepted.

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+
+The API of the `Range` class is rather similar to the API of the `Range` implementations in many other languages. Explaining the basic behavior should be straightforward. The Bounds logic might deserve a dedicated explanation since it's necessary to use the default constructor and might be unexpected. 
+
+The error management and the logic of the parameter `step` also deserve a dedicated explanation: while being similar to the logic seen in many other languages, in a quick survey we understood that the design of these behaviors have a much bigger variance and diversity across different languages. 
+
+# How We Test This
+
+* Test that the constructors actually set the fields as expected
+* Generate partial and full ranges, forward and backward, for a given data type and verify that the returned iterator contains the expected values
+* Verify that all the error conditions actually return an empty iterator (negative step, invalid bounds)
```
SeanTAllen: can you list all the error conditions you are aware of?

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+* Explicitely test the behavior of bounds around the min and max values to avoid overflow errors
```
SeanTAllen: Are there tests you are aware of now that are needed for this? 

It might not be you implementing the RFC, so the more detail the better.

jemc:
Can you include the important points of the API design in the RFC, along with a few usage examples?

I know you linked a reference implementation, but it's good to have the API design in the RFC itself for posterity, since the reference implementation link may become invalid later. Also, seeing usage examples would make this easier to talk about.

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+# Drawbacks
+
+* Some behaviors might differ compared to the old Range implementation 
+* The "empty list error" logic instead of explicit errors might surprise a fraction of the users
```
SeanTAllen: This will break existing code so that is a drawback as well.

text/0000-range_redesign.md
```diff
@@ -0,0 +1,67 @@
...
+
+* to raise errors
+* not to support inclusiveness/exclusiveness
+* to support a different range definition logic, i.e. (begin, step, number of elements)
+
+The main alternative to the redesign as a whole is to keep the existing implementation.
+
+# Unresolved questions
+
+Some behaviors with float arithmetic are not really covered by the reference implementation and should be discussed.
```
SeanTAllen: what behaviors are those?

text/0000-range_redesign.md
chobeat: because it's less verbose for something that shouldn't have runtime errors when the use defines the range with known numbers. When you're using runtime values, after the definition of the range you can know if it's valid or not and handle both cases anyway by checking this condition.

text/0000-range_redesign.md
chobeat: First of all there are explicit tests to verify this. I would say that the main problem before directly addressing this goal was that some parts of the implementation (very similar to the existing one), when the bounds was the max_value or min_value of some data type, were overflowing when checking if the bound was reached, basically creating an infinite range. Now this is handled explicitely: when the next value would reach the bound and if that bound would have overflow problems, then we just end the range there.

text/0000-range_redesign.md
chobeat: Incl might be used when for example you want a range spanning a whole data type because you want to include the last value. It's quite niche use case so if you want to make the API minimal, it can be removed.

I feel "to" is much more useful: when you want to generate ranges of positive numbers only you want to avoid to specify 0 as the starting value.

chobeat: Should I justify each of them in the RFC?

text/0000-range_redesign.md
chobeat: Maybe I will add an example to the RFC comparing the two versions

chobeat:
I've addressed all the points you raised and I think the document is ready for another round of review.

text/0000-range_redesign.md
```diff
@@ -0,0 +1,88 @@
...
+
+# Detailed design
+
+The new implementation should fullfil the following design principles: 
+
+* be able to generate all the possible ranges, increasing and decreasing, for all the data types, without incurring in overflow errors.
+* offer a unified, usable, coherent API with the possibility to specify inclusive or exclusive bounds. To achieve this, the definition of the bounds and the range should be independent. The Range class should offer helper methods to cover the most common cases without explicitely instancing the bounds.
+* never raise errors but instead return an empty list. This decision arises from the assumption that a majority of ranges are defined with bounds known at compile time while declaring ranges using runtime values is less common. The error handling in that case has to be done by calling the `is_invalid()` method. A version of the class with explicit errors can be considered.
```
SeanTAllen, line 23: "majority of ranges are defined with bounds know at compile time" isn't my experience. if the majority of ranges were defined with bounds at runtime, would that change the "never raise errors" idea?

chobeat: yes, because the assumption is that ranges defined with hardcoded value don't need error handling. The range will be either always valid or always invalid. If this is not the most common scenario, raising an error would be more idiomatic than checking `is_invalid`. The idea is to renounce an idiomatic interface to save the user a lot of unnecessary error handling.

jemc:
It was mentioned on today's sync call that this implementation adds some allocation overhead that will significantly slow down use of `Range` in, say, a for loop.

Having `Inclusive` and `Exclusive` be a class containing a number means that they will be allocated as objects. Having them behind the trait `Bound` means that we will use virtual calls and not reified, which precludes inlining of the function and stack-allocation of the class.

To fix the performance issue, `Inclusive` and `Exclusive` could each be a `primitive` with no state, and the number is kept in a separate field. Making it use a primitive and a number separately means no allocation for that data. Additionally, they should be type parameters instead of runtime parameters, and `iftype` used to check the type parameter, so that the inclusivity/exclusivity control flow paths can be known at compile-time instead of runtime.

From a usability standpoint, it is suggested that we make sure the defaults for those type parameters are matching the current behaviour of Range for compatibility purposes.

sylvanc:
Another issue for me is that dynamically determining forward or backward would interfere with high performance code patterns. For example, writing a loop that was intentionally a forward loop but that might (dynamically) intentionally perform zero iterations (by setting the `begin` as higher than the `end`). This is common in, for example, matrix multiplication (a scenario of particular interest for me).

Much like @jemc says above for inclusive and exclusive, forward and backward could be encoded as primitive type parameters, allowing zero-cost reification.

jemc:
Removing from final comment period until the above comments are addressed.

PR #152, opened 2019-07-30
RFC: rename_maybe_pointer

### LINKED ISSUE #48: Change name of MaybePointer in builtin.

Theodus:
[Rendered](https://github.com/ponylang/rfcs/blob/maybe-pointer/text/0000-rename-maybe-pointer.md)

Closes #48


SeanTAllen:
We agreed at Sync that NullablePointer is the name we would like to use.

PR #153, opened 2019-08-03
LLVM command line options integrations

mshockwave:
The proposed feature allows users of `ponyc` to supply LLVM-specific command line options from `ponyc`'s command line interface in dev builds (i.e. Debug build). This should make development of `ponyc` easier by improving accessibilities of internal LLVM features during debugging and fine-grained tuning.

------

**from the Pony core team**:

Everyone who has an opinion on Option 1 vs Option 2, please leave a reaction to this comment where you do:

hooray reaction if you favor option 1 and rocket reaction if you favor option 2.

### DIFF
--- a/text/0000-llvm-cli-opts-integration.md
+++ b/text/0000-llvm-cli-opts-integration.md
@@ -1,0 +1,83 @@
+- Feature Name: llvm-cli-opts-integration
+- Start Date: 2019-07-28
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+The proposed feature allows users of `ponyc` to supply LLVM-specific command line options from `ponyc`'s command line interface in dev builds (i.e. Debug build). This should make development of `ponyc` easier by improving accessibilities of internal LLVM features during debugging and fine-grained tuning.
+
+# Motivation
+
+`ponyc` uses LLVM heavily for code optimizations and code generation. There are many cases where it requires developers to inspect some parts of the LLVM pipeline. For example, when debugging a crash induced by violation of LLVM's own verifier rules. It would be great to use LLVM's `-print-after-all` command line option to dump LLVM IR after each pass. However, currently there is no way to turn on this option directly from `ponyc`'s command line interface. Thus developers have no choice but modifying LLVM source tree to overwrite those flags by hard-coded values, which is absolutely a bad practice.
+
+Other cases like turning on advanced mode in some LLVM passes (e.g. the `-basicaa-recphi` flag in BasicAA to recursively analyze PHI nodes) during performance tuning, or using LLVM's built-in bisection features (i.e. `-opt-bisect-limits`) will also be benefit from this proposal. One should expect this new functionality to be:
+1. Handy when they're developing or debugging the LLVM part in `ponyc`.
+2. Working well with existing `ponyc` command line options.
+3. Not available in release build, as we want to keep minimum exposure of internal LLVM part to normal users.
+
+# Detailed design
+
+One of the goals of this feature is to keep LLVM-specific command line options separated from existing `ponyc` command line options. There are two ways to achieve it.
+## Option1 - Clang style options prepending
+For every LLVM-specific command line option, one need to prepend a special token command line options before it. For example, to pass `-basicaa-recphi` , a LLVM-specific option, to `ponyc`, instead of running command like this:
+```
+ponyc -basicaa-recphi <rest of the ponyc options>
+```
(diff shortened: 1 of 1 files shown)


dipinhora:
Julia's solution seems like a good alternative: https://docs.julialang.org/en/v1/manual/environment-variables/index.html#JULIA_LLVM_ARGS-1

mshockwave:
@dipinhora thanks for the feedback, I'll put into the alternative section.

SeanTAllen:
Everyone who has an opinion on Option 1 vs Option 2, please leave a reaction to *this comment* where you do:

`hooray` reaction if you favor option 1 and `rocket` reaction if you favor option 2.

SeanTAllen:
Thanks for this @mshockwave. 

We are going to discuss this on August 20th after trying to collect some feedback on the two options.

dipinhora:
I'd prefer option 3: either an environment variable or a command line argument passed in using quotes (`"to handle spaces"`) so that the entire blob can be passed off to LLVM directly without any pre-processing or thought about how to structure the arguments (to prefix them all with `-mllvm` or which separator to use instead of ` `).

From what I can tell, the Julia environment variable implementation boils down to 1 line of code (https://github.com/JuliaLang/julia/blob/master/src/codegen.cpp#L7613) calling into an existing LLVM function (https://llvm.org/doxygen/namespacellvm_1_1cl.html#ae5a3364c95fc71013039ef5fd7aa7fe0). I'm guessing there's a similar function for parsing all LLVM arguments from a single string also.

jemc:
I like Dipin's suggestion.

mshockwave:
Actually I see no problem implementing all of the options, including option 3 proposed by @dipinhora . Option 1 and option 2 can exist at the same time. If user also supply environment variable in option 3, we just concat it before or after those which are supplied from command line.

SeanTAllen:
I'm not in favor of multiple ways to do this. I want 1 way to do it. Definitely not 2 or 3 possible ways.

SeanTAllen:
Ok, let's try this again, based on early feedback.

We have two ideas that seem to be favored.

The first idea is labeled as "option 2" in the current RFC text.
The second idea is listed as an alternative in the current RFC text.

If you are in favor of "option 2" please respond to this comment with the "hooray reaction". If you are in favor of the Julia inspired environment variable option, please respond with the "rocket reaction".

slfritchie:
Looking at the https://docs.julialang.org/en/v1/manual/environment-variables/index.html#JULIA_LLVM_ARGS-1 link, I don't see anything there that tells me what the contents of the `JULIA_LLVM_ARGS` env var is supposed to contain.  Is it verbatim LLVM compiler args?  Or something else.  Some clarification and an example would be really helpful.

mshockwave:
@slfritchie I think they simply use [this API](https://llvm.org/docs/CommandLine.html#the-cl-parseenvironmentoptions-function). Here is an usage example:
```
export MY_PONY_LLVM_ARGS="-debug-only=mem2reg,sroa -stats"
./ponyc <pony args>
```
somewhere in `ponyc`, we call the API like this:
```
llvm::cl::ParseEnvironmentOptions("ponyc", "MY_PONY_LLVM_ARGS");
```

In short, we put exactly the same CLI option string as we passed to other LLVM tools into the environment variable, and we can choose our own environment variable name.
However, as mentioned in the LLVM command line doc above: **the API does not support quoting**. That is, `-some-args="foo bar"` will be seen as two separated options `-some-args=foo` and `bar` with this API.

SeanTAllen:
Ok, we are down to vote time. 3 possibilities. Please vote as a reaction to this comment:

- Command line option 2 as listed in the RFC add a :rocket:
- Environment variable listed in comments here and in alternatives :heart:
- Allow for both of these :eyes:

aturley:
The voting on this will close on September 10.

SeanTAllen:
This has been accepted in principal. @mshockwave can you update the RFC to make option 2 as "the option" and have the others as alternatives that we didn't accept, but did exist?

Once that is done, we can merge and officially accept.

Thanks.

mshockwave:
@SeanTAllen done. 
Also just curious, after this RFC is accepted, are we going to proceed with the implementation?

SeanTAllen:
@mshockwave when accepted, an issue will be created on the ponyc that anyone can pick up and start working on. if you want to do the work, please do.

SeanTAllen:
This RFC has been accepted.

SeanTAllen:
@mshockwave there's an issue open for this now: https://github.com/ponylang/ponyc/issues/3327

mshockwave:
@SeanTAllen awesome!

PR #155, opened 2019-11-15
Pony RT options passing

alexsnaps:
# Pony runtime tweaks through arguments

## The existing

* `--pony*`  arguments are provided to the `ponyc` outputted executable
* We’ve enhanced bad input handling to _existing_ arguments (required option value existed, now we also check for boundaries)
* Added `--ponyhelp` as the input grew 
* Existing recognized `--pony*` arguments are _not_ exposed through `Env.args` (which I think is a wise decision, I only see pros to it, but in case some think this might be revisited as part of this, I guess it could)

## Issues with the existing

* When renaming `--ponythreads` to `--ponymaxthreads` , an upgrading user could miss this change, and this would just be silently ignored. In this very case resulting in ignoring the user specified value and falling back to default. Also resulting in the old argument now being exposed to his  `Env.args`
* Once GA (i.e. post release 1.0) how can we deprecate arguments? How can we even add one, without possibly break existing code (e.g. someone using that arg already)?

---

This RFC tries to address the above. Mostly trying to get the ball rolling again. I've had some personal things going on that came in between me and this… 

Moving forward I'll start implementing and filling the `Detailed design` section in. But wanted to get this discussion going again… in a more "formal" way than on Tulip. So please have at it! 🙇 

[Formatted version here](https://github.com/alexsnaps/rfcs/blob/0065/text/0000-runtime-opts.md)

### DIFF
--- a/text/0000-runtime-opts.md
+++ b/text/0000-runtime-opts.md
@@ -1,0 +1,72 @@
+- Feature Name: runtime-opts
+- Start Date: 2019-11-02
+- RFC PR: 
+- Pony Issue: 
+
+# Summary
+
+This feature is a proposal to consolidate how we deal with these options, providing a way to avoid colissions with user defined options, upgrade and deprecation path across Pony releases. 
+
+# Motivation
+
+The options (`--pony*`) provided to a compiled Pony program to control its runtime environment have evolved overtime, with the latest instance being the `--ponythreads` to `--ponymaxthreads` [rename](https://github.com/ponylang/ponyc/issues/3318). With the current approach, this change would silently ignore the old option. It could possibly fail further parsing within user's application, as recognized `--pony*` arguments are _not_ exposed through `Env.args` the old, yet the now "unused" option would be exposed.
+The opposite being just as true: a user specified option could be elided if the runtime would consume it in a subsequent release. Nothing prohibits a user from using the `--pony` prefix. 
+
+# Detailed design
+
+Building upon [RFC-0064](0064-llvm-cli-opts-integration.md), as we probably should avoid introducing a new pattern, the proposal is to have a single option, namely `--pony_opts`, to pass all options as a single, comma seperated string, for the runtime to consume, e.g.:
+
+```
+$ ./helloworld --pony_opts="minthreads=2,pin,maxthreads=4" 
+Hello, world.
+```
+
+Providing an unknown "key" within that string would result in an error:
+
+```
+$ ./helloworld --pony_opts="minthreads=2,pin,maxthreads=4,answer=42" 
+Unrecognized option answer, see --pony_opts=help for more information
+$ echo $?
+255
+```
+
+Also enabling for deprecation by providing a warning (on stderr?):
+
+```
+$ ./helloworld --pony_opts="minthreads=2,pin,threads=4" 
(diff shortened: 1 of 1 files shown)


text/0000-runtime-opts.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Motivation
+
+The options (`--pony*`) provided to a compiled Pony program to control its runtime environment have evolved overtime, with the latest instance being the `--ponythreads` to `--ponymaxthreads` [rename](https://github.com/ponylang/ponyc/issues/3318). With the current approach, this change would silently ignore the old option. It could possibly fail further parsing within user's application, as recognized `--pony*` arguments are _not_ exposed through `Env.args` the old, yet the now "unused" option would be exposed.
+The opposite being just as true: a user specified option could be elided if the runtime would consume it in a subsequent release. Nothing prohibits a user from using the `--pony` prefix. 
+
+# Detailed design
+
+Building upon [RFC-0064](0064-llvm-cli-opts-integration.md), as we probably should avoid introducing a new pattern, the proposal is to have a single option, namely `--pony_opts`, to pass all options as a single, comma seperated string, for the runtime to consume, e.g.:
```
jemc, line 17: `--pony-opts` might be more idiomatic for CLI than `--pony_opts`?

That is, dashes are probably more idiomatic than underscores.

SeanTAllen:
Hi @alexsnaps,

We discussed this at sync. No one has strong feelings about what should replace what we have (in terms of your alternatives) beyond the "-" instead of "_".

We think either of your alternatives is better than what we have so we are good with moving forward with them.

Thank you for this RFC. We have put it into a final comment period and will be voting on it soon. Can you update to change the `_` to `-`; assuming you agree with that request.

redvers:
Does it make sense to have a call to provide the software author the ability to:

1. Potentially disable the options.
2. Have visibility of the options if they choose?  like env.ponyargs?

SeanTAllen:
@redvers What options do you imagine the software author would want to disable? What does "disable" mean to you? 

Can you give some scenarios you are thinking of?

redvers:
1. has been implemented already by the looks of it.
2. Visibility of the options would be good for logfiles so we can see in logs if our end-users alter any of the runtime values.

rhagenson:
Looks good to me. I agree on the `-` rather than `_` change.

Personally, I think the namespacing alternative is superior to the environment variable route. The [Pony philosophy](https://www.ponylang.io/discover/#the-pony-philosophy) includes Correctness as its first tenet and Consistency as the second to last. Managing environments in addition to executables is a hassle and decreases the likelihood of correctness, and running without the same environment would break consistency. A given program should run correctly and consistently wherever one chooses to run it.

My thought on a sort-of halfway to using environment variables (where one gets to always use their preferred defaults) is to bake the desired runtime defaults into the executable via compile-time options. People would get to set their preferred defaults when they compile the program, but it executes with the same semantic defaults no matter where it is run despite the environment. Configuring these personal preferences might be a good feature for the dependency manager `corral` since it would allow compiling for different semantic environments such as `prod` and `dev` (and perhaps one day `distributed`, where environments would be a royal mess to manage).

jemc:
Removing from final comment period until the `-` vs `_` point is addressed in some way.

PR #156, opened 2019-12-23
Behaviour call syntax

Theodus:
[Rendered](https://github.com/ponylang/rfcs/blob/behaviour-call-syntax/text/0000-behaviour-call-syntax.md)

### DIFF
--- a/text/0000-behaviour-call-syntax.md
+++ b/text/0000-behaviour-call-syntax.md
@@ -13,7 +13,15 @@ A call site distinction between functions and behaviours on an actor would give 
 
 # Detailed design
 
-A new syntax of `a~b()` is used for calling a behaviour (`b`) on an actor (`a`). The previous notation of `a.b()` will now only be allowed at a function call site.
+- A new syntax of `a~b()` is used for calling a behaviour (`b`) on an actor (`a`). The previous notation of `a.b()` will now only be allowed at a function call site.
+
+- Any actor function with sendable parameters may be called asynchronously as behaviours using the new `~` syntax. Doing so will cause the return type of the function to become `None` when called as a behaviour.
+
+- Apply sugar on behaviour calls will take the form of `a~()` as the asynchronous alternative to `a()`. So `a~()` will be sugar for `a~apply()`.
+
+- Similarly, `A~()` will be the new create sugar for actor constructors that execute asynchronously as they do now. `A()` will now be syntactic sugar for calling the `create` constructor of any class or actor synchronously. Synchronous actor constructors will be a new language feature.
+
+- Partial function application will use the `$` operator instead of the current `~` operator in order to remove ambiguity from these two disparate language features.
 
 As an example, the following `TCPConnectionNotify` has its `received` function executed within a `TCPConnection` behaviour. In this example, "stuff" will never be written if "done" has been received. This is because `conn~write("things")` is called as a behaviour and `conn.close()` is called as a function. Using the new syntax, it is immediately obvious that the execution of `write` is not occurring in-line with the rest of the function body.
 ```pony
@@ -62,20 +70,8 @@ This would be a major breaking change, and would likely break most existing Pony
 
 # Alternatives
 
-Alternative operator for async call syntax (currently `~`) that doesn't conflict with the [partial applications](https://tutorial.ponylang.io/expressions/partial-application.html).
+- Alternative syntax for partial function application, instead of `~` or `$`.
 
 # Unresolved questions
 
-- How should partial application syntax be modified to support this change? I believe that behaviour calls are much more common than partial application, and should therefore take the more "understandable" and terse syntax. But this is very subjective and suggestions for alternative syntax for either are welcome.
-
-- Should apply sugar still be allowed for actors? If so, we would have to differentiate behaviour calls with something like `a~()`. For reference, here is a list of all public actors in the stdlib with `apply` behaviours:
-  - `builtin.InputStream`
-  - `builtin.Stdin`
-  - `bureaucracy.Custodian`
-  - `ponybench.PonyBench`
-  - `ponytest.PonyTest`
-  - `promises.Promise`
-  - `term.ANSITerm`
-  - `time.Timers`
-
-- May any actor function with sendable parameters be called, using the new syntax, as behaviours? If so, any return type on the function would become `None` when called this way.
+- Should the `be` keyword continue to exist to enforce that a function may be called asynchronously (and therefore has only sendable parameters)?
--- a/text/0000-behaviour-call-syntax.md
+++ b/text/0000-behaviour-call-syntax.md
@@ -19,7 +19,11 @@ A call site distinction between functions and behaviours on an actor would give 
 
 - Apply sugar on behaviour calls will take the form of `a~()` as the asynchronous alternative to `a()`. So `a~()` will be sugar for `a~apply()`.
 
-- Similarly, `A~()` will be the new create sugar for actor constructors that execute asynchronously as they do now. `A()` will now be syntactic sugar for calling the `create` constructor of any class or actor synchronously. Synchronous actor constructors will be a new language feature.
+- Update sugar on behaviour calls will take the form of `a~(key) = value` as the asynchronous alternative to `a(key) = value`. So `a~(key) = value` will be sugar for `a~update(key, value)`.
+
+- `A~()` will be the new create sugar for actor constructors that execute asynchronously as they do now.
+
+- The compiler will no longer treat `fun tag` as equivalent to `be` for the purposes of structural or nominal subtyping.
 
 - Partial function application will use the `$` operator instead of the current `~` operator in order to remove ambiguity from these two disparate language features.
 
@@ -72,6 +76,8 @@ This would be a major breaking change, and would likely break most existing Pony
 
 - Alternative syntax for partial function application, instead of `~` or `$`.
 
-# Unresolved questions
+# Unresolved questions for future RFCs
 
 - Should the `be` keyword continue to exist to enforce that a function may be called asynchronously (and therefore has only sendable parameters)?
+
+- Should we allow synchronous actor constructors?
--- a/text/0000-behaviour-call-syntax.md
+++ b/text/0000-behaviour-call-syntax.md
@@ -1,0 +1,81 @@
+- Feature Name: behaviour_call_syntax
+- Start Date: 2019-12-22
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+Modify the language to differentiate between synchronous function calls and asynchronous behaviour calls at the call site.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


jemc:
> May any actor function with sendable parameters be called, using the new syntax, as behaviours? If so, any return type on the function would become `None` when called this way.

I'll go on record as saying "yes" to this question and that this is the main reason why I support this change.

As I mentioned in last week's sync call when I brought up this idea, I think the main advantage here is being able to avoid API sprawl for any actor type that allows `ref` access to itself (via the [Access Pattern](https://github.com/ponylang/pony-patterns/blob/126616e7374a272e183430a759202377b7f80cd5/content/async/access.md), the [Notifier Pattern](https://github.com/ponylang/pony-patterns/blob/master/content/code-sharing/notifier.md), or similar). Today, any such actor type is pressured to create both synchronous and asynchronous methods of the same API to cater to the varying needs of different callers.

For me, the primary advantage of having an explicit syntax for asynchronous calls _at the call site_ is that it should allow us to remove the distinction _at the definition site_, and allow any eligbile method (that is, one that can be called without non-sendable arguments) to be issued either synchronously or asynchronously, based on the needs and circumstances of the caller.

jemc:
> How should partial application syntax be modified to support this change? I believe that behaviour calls are much more common than partial application, and should therefore take the more "understandable" and terse syntax. But this is very subjective and suggestions for alternative syntax for either are welcome.

I would agree, and I would propose the symbol `.&` to replace the current syntax for partial application.
- Mnemonically, the `&` in `.&` can be read as "and" in a way that can remind you of the idea that you're taking the receiver object (on the left side of the symbol) *and the* method name and arguments (on the right side of the symbol), and packaging them up into a new callable object.
- For C and C++ comers, the ampersand conjures up ideas of "by reference", which is *sort of* what's going on here - you're sort of taking a reference to the function without actually calling it.
- For Ruby and Crystal comers, the ampersand is associated with grabbing a reference to a method without calling it, and creating an object that "wraps" that method to make it callable later.

jemc:
> Should apply sugar still be allowed for actors? If so, we would have to differentiate behaviour calls with something like `a~()`.

I don't care much about this point, but I think the suggestion of `a~()` makes sense in the context of the apply sugar we already have.

However **something potentially more far-reaching** that this brings to mind for me is the question of `create` sugar as well.

Recall that actor constructors are executed asynchronously. Are we also proposing to mark all actor constructor calls with the `~` symbol, and if so, how does that interact with the `create` sugar?

SeanTAllen:
> Recall that actor constructors are executed asynchronously. Are we also proposing to mark all actor constructor calls with the ~ symbol, and if so, how does that interact with the create sugar?

@jemc that is an excellent point. I'm going to think about that for a while.

rhagenson:
I am in support of the proposed `~` for behavior calls to differentiate them from function calls. The other symbol I could see using is `@` (subtlety suggesting "**a**synchronous" and that the call is "at" the actor, not "here"). Visually `conn@write("stuff")` does not throw me, but I am unaware of `@` usage in other languages which might cause undo confusion for folks familiar with its use elsewhere.

> I would agree, and I would propose the symbol .& to replace the current syntax for partial application.

Even if the behavior call symbol does not become `~` I am in favor of this partial application symbol change as having an entirely dissimilar symbol from `.` for partial application threw me the first time I saw it.

rhagenson:
Additional point: using an `@` I would initial expect it to be `<behavior>@<actor>`, but I think that order being backwards as `<actor>@<behavior>` is not an unreasonable leap in thinking given it will match up with `<actor>.<function>` which would seem like nonsense in its reverse order `<function>.<actor>`.

SeanTAllen:
> Recall that actor constructors are executed asynchronously. Are we also proposing to mark all actor constructor calls with the ~ symbol, and if so, how does that interact with the create sugar?

I think that we would need to remove create sugar. I've never been a fan of the confusion that create/apply sugar creates for people new to the language (and me when I'm switching languages). So, at the moment, I'd be ok with removing `create` sugar but leaving `apply` sugar.

SeanTAllen:
A point to raise...

if `be` was removed and only `fun` existed then, partial functions need to be dealt with.

In particular, I think that it would have to be a compile-time error to try to call a partial function async.

I was discussing this with @jemc. 

To quote him:

"I think that a partial `fun` just wouldn't be eligible. And we'd have an error message that clearly explains why. Just like any fun that requires non-sendable arguments wouldn't be eligible"

adri326:
I would rather have a dot-like character sticking around; `@` tends to easily hide among other alphabetical characters. I thus propose `.@` (e.g. `conn.@write("stuff")`), which would be thought of as "call (`.`) at (`@`)".

rhagenson:
I can see what you mean about `@` hiding among alphabetical characters. Using `.@` for behaviors still seems reasonable to me: 1) achieves the differentiation from synchronous functional calls, 2) has similarity to the current and carried over function call syntax `.`, and 3) does not class with an already symbol like switching to `~` (`@` is used in FFI, but there it always starts the line so does not contextually clash -- if "contextual clashing" is even a recognized thing).

jemc:
After [some discussion in the Zulip chat](https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/synchronous.20actor.20constructors), I tentatively propose that we use this `.` vs `~` syntax to allow the caller to choose whether to call the actor constructor synchronously or asynchronously.

That is, `MyActor.create()` would execute the constructor synchronously at the call site, while `MyActor~create()` would execute the constructor asynchronously, via a message passed to the new actor telling it to do so. The requirement of all-sendable arguments would be the same for both cases.

Under this system, I'd propose that the `create` sugar would result in a synchronous constructor call - that is, `MyActor` would be equivalent to writing `MyActor.create()`, as it is now - it's just that it would become a synchronous execution by default. For asynchronous execution, there would be no shortcut sugar - you would need to write `MyActor~create()`. (Please note that I'd also be okay with the alternative of just getting rid of `create` sugar entirely, but that's more controversial and would preferably be bundled with some other changes I have in mind in order to make it worthwhile).

Theodus:
The following updates have been made:
1. Apply and create sugar is included, using the `a~()` syntax. This introduces synchronous actor constructors using the current `A()` create sugar on actors.
2. Any actor function with sendable parameters may be called asynchronously as behaviours.
3. Partial function application will use the `$` operator. This will likely require further discussion and I only chose it over the other suggestions for now since it seems unlike other operators used in the current syntax.

rhagenson:
My responses to the updates:
1. No strong opinion on the sugar. I think my preference would be for actor construction to by async by default, but no good reason for that preference other than "that is what I would expect" and adding one additional character to make it async is hardly a barrier
2. No complaint on this. Would this disallow having a behavior and function with the same name or would the sync/async call determine which is used at the call site?*
3. Choosing a different character is reasonable. One of my growing concerns with Pony is the number of special characters it uses, but I would certainly prefer more special characters than ambiguous or confusing limited special character use

*: I have not tested having a function and behavior of the same name, but the new syntax would encourage defining more functions as opposed to behaviors (as functions can be made async based on parameters with this RFC, but behaviors cannot be made sync). My question is more would there be an option to define a behavior with the same name as a function that is instead called when the "function" is called asynchronously?

Edit: Before it slips my mind, if a function and behavior can share a name then I recommend adding to the style guide to have methods on a actor be alphabetical so that these two equal name methods are placed next to one another in source.

jemc:
> Would this disallow having a behavior and function with the same name or would the sync/async call determine which is used at the call site?*

I don't yet have a strong opinion on this, but my initial inclination is to lean away from it. It seems to hold too much potential for confusion, for not a lot of benefit. But I'm open to discussion about it.

> if a function and behavior can share a name then I recommend adding to the style guide to have methods on a actor be alphabetical so that these two equal name methods are placed next to one another in source.

I think we can have the latter without the former. In general I dislike rules about alphabetizing code and prefer style conventions like "the most closely related functions should be adjacent to eachother in the source". This still achieves the goal you're intending with this statement, without placing a strict total order on of your code (which can really break up those logical associations).

rhagenson:
On the first point, I see this as a future RFC as it is a much longer discussion only semi-related to the behavior call syntax. My initial inclination is opposite yours by leaning toward it. I see its most common use cases as breaking up large/long computations by kicking off more messages to itself or having the behavior variant return on a message to the sender. I.e. the function returns in place while the behavior takes a tag to the sender and sends a message back in response.

On the second point, I see it as reasonable to not enforce a total order to the methods. I personally try for some unhappy middle ground in general rules of thumbs such as alphabetizing which nobody really likes, it can spread related code out, but can always be applied as a general rule to the same result. Prevents opinion from influencing/changing the order of methods unnecessarily. I could also see a style of having functions together, behaviors together, and function/behavior pairs together in source, creating effectively four "blocks" in each actor: 1) fields and initializers, 2) lone functions, 3) lone behaviors, 4) function/behavior pairs. But personal preference and the official style guide are hardly the same.

SeanTAllen:
Niclas raised a very interesting point here: https://ponylang.zulipchat.com/#narrow/stream/190591-sync/topic/12.2F17.2F2019/near/184075534

> I have a comment on behave/async RFC; At first I thought this sounded like a good idea, just like aliases defines whether a type is immutable or not (rather than traditional way of the declaration), but there is a downside that I think needs to be addressed. The author of the called code may make assumptions about call order, which seems to be potentially upset with this feature, if any method is both fun and be at the same time. It could make design a lot harder to do as one would have to imagine all possible right and wrong sequencing in the API. So, perhaps there need to be a way for the target code to control whether sync and/or async calls are possible, maybe per method or maybe for the entire type. Example; Queue ordering in a collaborative (extensions, multi-author) library can no longer be guaranteed, and greedy extensions could perhaps bypass the async ordering, and jump a queue.

rhagenson:
@SeanTAllen I think that point by Niclas is why I had the dispatch on fun/be in my head. Perhaps we can simplify the design by requiring functions, even if they have all sendable parameters, must be synchronous, but a behavior of the same name may be defined? This would prevent any "oh, but this function is async because..." confusion and let the called code author determine if they will support both sync and async calling of a method.

SeanTAllen:
@rhagenson 

if `be` and `fun` both still exist then at that point I think we are talking about some sugar so that you don't need to do:

```pony

be async_version() =>
  sync_version()

fun sync_version() =>
  None
```

I'm not in favor of "behavior with the same name must be defined". Pony doesn't allow overloading operator names and I think this is a good thing.  However, if that is sugar for something else, eh, maybe, but that gets into some weird name mangling which... not so sure about that. Given you can already do the above pattern. I would be hesitant to get clever. I'm definitely not a fan of clever.

rhagenson:
Sounds good. Name mangling is never really a good idea, but felt obligated to toss it into consideration as a solution to the method-level called code concern raised by Niclas.

adri326:
I think behaviors should be considered as message-passing-ready methods (instead of message-passing-only methods), this would be especially useful when writing libraries.

jemc:
When this is implemented, it might be nice for someone to write a script that can auto-rewrite code based on the error output.

Here is a script I wrote back when we did the explicit partial calls change that broke many call sites, just like this one will: https://gist.github.com/jemc/95969e3e2b58ddb0dede138c737907f5

Theodus:
We have voted this in during the `4/28/2020` sync

Theodus:
Continued in https://github.com/ponylang/rfcs/pull/163

PR #157, opened 2020-01-06
Move Hashable to builtin package and make it easier to make objects Hashable

mfelsche:
[Rendered](https://github.com/ponylang/rfcs/blob/move-hashable-builtin/text/0000-hashing-builtin.md)

Suggested changes: https://github.com/ponylang/ponyc/compare/hashing-builtin?expand=1#diff-3fcbdccf4a6378f88bdb211e3838307bR9

### DIFF
--- a/text/0000-hashing-builtin.md
+++ b/text/0000-hashing-builtin.md
@@ -59,7 +59,7 @@ This is a cranky, crufty and ugly, hand-crafted hash-function for the class abov
 
 ```pony
   fun hash(): USize =>
-    var result = USiz(17)
+    var result = USize(17)
     result = (result * 37) + _field1.hash()
     result = (result * 37) + _field2.hash()
     result
--- a/text/0000-hashing-builtin.md
+++ b/text/0000-hashing-builtin.md
@@ -1,0 +1,177 @@
+- Feature Name: move_hashable_to_builtin
+- Start Date: 2019-12-19
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Move `Hashable` to the builtin package and make it easier to make objects `Hashable`
+by defining convenience functions that use a common good and fast way of
+combining hashes of object fields into a single hash value. Also extend documentation
+and tutorial with information on how to make objects hashable.
+
+# Motivation
+
+It is possible to make any object implement the interface `Hashable` or `Hashable64` but
+if we want to state that by explicitly extending it we need to add a `use "collections"`,
+although we do not intend to use any pony collection:
+
+```pony
+use "collections"
+
+class MyClass is Hashable
+  fun hash(): USize => 42
+```
+
+The main point here is that `Hashable`s main use case is its usage in `HashMap` etc. but
+it is not limited to that. The notion of Hashability is generic enough to earn a rank
+among other interfaces like `Equatable` which already reside in the `builtin` package.
+
+In fact, there exists a fundamental relationship between `Equatable` and `Hashable`: If
+a class `C` is both `Equatable` and `Hashable`, if two instances `c1` and `c2` are considered
+equal (as far as structural equality is concerned), their hash should also be the same.
+
+This RFC also suggests to add a primitive called `Hashing` which provides convenience methods for
+defining fast and good (enough) `fun hash(): USize` methods for arbitrary objects, combining hashes of the objects fields
+with a good hash function ([Siphash-2-4](https://131002.net/siphash/)).
+
+Consider the following example:
+
+```pony
+
+class val MyClass
+  let _field1: String
+  let _field2: U32
+
+  new val create(f1: String, f2: String) =>
+    _field1 = f1
+    _field2 = f2
+```
+
+To make this class implement `Hashable` or `Hashable64`, we need to find a way to
+combine the hashes of both fields if these should make up the identity for this class.
+In order to make it more convenient for people to write useful classes
+it is favorable to advertise a common way to define hash functions for objects and offer a convenient way
+to implement them without too much boilerplate.
+
+This is a cranky, crufty and ugly, hand-crafted hash-function for the class above
+(adapted from how most IDEs in java auto-create hashCode methods):
+
+```pony
+  fun hash(): USize =>
+    var result = USiz(17)
+    result = (result * 37) + _field1.hash()
+    result = (result * 37) + _field2.hash()
+    result
+```
+
+A hash function implemented using a convenience function suggested by this RFC
+would look as nice and clean as this:
+
+```pony
+  fun hash(): USize =>
+    Hashing.hash_2(_field1, _field2)
+```
+
+# Detailed design
+
+## Move Hashable to Builtin
+
+We move the `Hashable` and `Hashable64` interfaces to the `builtin` package.
+
+## Make hashing objects convenient
+
+A primitive called `Hashing` will be introduced into the `builtin` package, alongside with a pure pony `Siphash` implementation.
+Currently libponyrt contains code for hashing byte-sequences with a C-implementation of Siphash
+in https://github.com/ponylang/ponyc/blob/master/src/libponyrt/ds/fun.c#L92 . As the implementation is already iterating over
+the array in 8-byte (4-byte for 32-bit platforms) increments, we can also use it in a streaming fashion, feeding it consecutive U64.
+This way we can feed it hashes of object fields and compute the sip-hash-2-4 of all the hashes of the object fields.
+
+When hashing objects based on their fields, we must take all those fields into account. They might be primitive types, objects with fields themselves,
+their hash might be using the fields identiy using the `digestof` operator etc. etc. So generating a hash boils down to a clever way to combine the hashes of all fields
+that should be considered. The hashes of the fields are chosen as a common denominator.
+And, ideally, those hashes should reflect all of the fields contents. 
+So if we build a hash of those hashes using a good enough hash function, we should, by transitivity, also get a good enough hash of the whole object in return.
+
+As far as I know, [SipHash-2-4](https://131002.net/siphash/) can be considered a good enough hash-function.
+It is also already in use for byte sequences and strings in Pony.
+
+This `Hashing` primitive will contain methods for hashing `ByteSeqs`, as e.g. `Array[U8]` does not implement `Hashable` but can be easily hashed,
+in fact all Arrays of primitive numeric types can and should be very easily hashed.
+It will contain methods for hashing `ReadSeq[Hashable #read]` by combining the element hashes into one hash value 
+using the [SipHash-2-4](https://131002.net/siphash/) implemention described above.
+It will also contain methods for conveniently combining hashes from different objects, like:
+
+```pony
+
+primitive Hashing
+  fun hash_3(a: Hashable box, b: Hashable box, c: Hashable box): USize =>
+    """
+    convenience function for combining the hashes of three object implementing `Hashable`
+    into one hash value.
+    """
+    ifdef ilp32 then
+      let sip = HalfSipHash24Streaming.create()
+      sip.update(a.hash().u32())
+      sip.update(b.hash().u32())
+      sip.update(c.hash().u32())
+      sip.finish()
+    else
+      let sip = SipHash24Streaming.create()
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Is there anything in the standard library that is outside of the collections package that implements these interfaces?

mfelsche:
Only builtin classes like the numeric types, String and Pointer implement a `hash` or `hash64()` function. Nothing else inside of stdlib.

text/0000-hashing-builtin.md
```diff
@@ -0,0 +1,177 @@
...
+    else
+      let sip = SipHash24Streaming.create()
+      sip.update(a.hash().u64())
+      sip.update(b.hash().u64())
+      sip.update(c.hash().u64())
+      sip.finish()
+    end
+```
+
+It will contain such functions for 2 arguments up to 12 arguments, maybe more.
```
jemc, line 128: One alternative to this approach that I want to mention is something like what Crystal does.

Every object in Crystal has a `hash` function that takes one argument of type `Hasher`. By default, that argument is a new instance of `Hasher`, which is a class similar in concept to what you've proposed here with the `Hashing` primitive. The difference is that it is implemented not as a primitive, but as an immutable/persistent data structure (similar to those `collections/persistent` in Pony).

If we took a similar approach, someone could implement hashing with an example looking something like this:

```pony
class MyClass
  let foo: Foo
  let bar: Bar
  let baz: Baz

  fun hash(): U32 =>
    hash_with(Hasher.create()).result()

  fun hash_with(hasher': Hasher): Hasher =>
    var hasher = hasher'
    hasher = this.foo.hash_with(hasher)
    hasher = this.bar.hash_with(hasher)
    hasher = this.baz.hash_with(hasher)
    hasher
```

The advantage to this kind of approach is that it won't require us to define N different `hash_1`, `hash_2`, `hash_3`, [...] functions as proposed here.

This approach also adds the ability to hash a dynamic number of inner fields to hash, rather than being limited to a static function signature with N arguments. For example, an implementation of a `List` style of data structure might want to hash everything present in the list, and in this approach it could be as simple as this:

```pony
  fun hash_with(hasher': Hasher): Hasher =>
    var hasher = hasher'
    for item in values() do
      hasher = this.foo.hash_with(hasher)
    end
    hasher
```

mfelsche: I see the beauty in this approach, but it also feels like a weird indirection for the user.

I would not see a requirement for keeping the Hasher immutable/persistent, but would rather make it mutable, as i did with `Siphash24Streaming` here: https://github.com/ponylang/ponyc/compare/hashing-builtin?expand=1#diff-3fcbdccf4a6378f88bdb211e3838307bR9

```
fun hash_with(hasher: Hasher ref): Hasher ref =>
  this.foo.hash_with(hasher)
  this.bar.hash_with(hasher)
  this.baz.hash_with(hasher)
  hasher
``` 

Whatever is faster here.

So, this approach would change the interface `Hashable` to also include the `hash_with` method and would require to add a `Hasher` interface and implementation.

Also notice, that my `Hashing` implementation could conveniently solve the case of dynamic number of fields by adding a `hash_iter(iter: Iterator[Hashable box])` method.

mfelsche: In general i think both approaches are interesting and would be willing to rewrite the RFC to incorporate @jemc's suggestion  as the RFCs main suggestion, if people tend towards it.

jemc: One note on performance - I'm not sure how the other differences would affect perforamance, but one difference that is worth noting is that the `Hashing.hash_2(x, y)` approach would need virtual calls, due to the calls being done on the `Hashable` interface arguments without a concrete type. The workaround for this would be to use type parameters to specialize for different types, but that leads to a much more verbose call site (i.e. `Hashing.hash_2[TypeX, TypeY](x, y)`).

I believe that the `Hasher` class approach wouldn't need any virtual calls, but maybe I'm missing something.

jemc: Regarding mutable vs immutable `Hasher` class, I am not strongly opinionated, and would in general lean toward whichever approach is more performant. Probably the mutable approach is more performant in current Pony, since immutable objects need to allocated many more times, and in the absence of Heap-to-stack optimization, that's kind of a big deal.

Crystal addresses this by using a `struct` instead of a `class` for this type. In Crystal, a `struct` is passed "by value" instead of "by reference", so it's something like a Pony tuple, but with callable methods for it. This also explains why it is implemented as an immutable data structure, since that's the only sane way to deal with a "by value" object without major headaches.

jemc: Now that I mentioned If we wanted to do absolutely zero-allocation (and still zero-virtual-call), we could use a tuple ourselves. Something like:

```pony
type HashState is (U32, U32) // or whatever other kind of state we need...
primitive Hashing
  [...]
```

```pony
class MyClass
  let foo: Foo
  let bar: Bar
  let baz: Baz

  fun hash(): U32 =>
    Hashing.finish(hash_with(Hashing.begin()))

  fun hash_with(state': HashState): HashState =>
    var state = state'
    state = this.foo.hash_with(state)
    state = this.bar.hash_with(state)
    state = this.baz.hash_with(state)
    state
```

This would probably be the most performant solution (if I had to guess), while also being generally nice in terms of developer experience.

mfelsche: I like where this is going, @jemc !

Using SipHash-2-4 needs a state of `(U64, U64, U64. U64, USize)` in 64 bit systems, and `(U32, U32, U32, U32, USize)` on 32 bit systems. So we need some dispatching, or we work with USize all the way down. Anyway, it seems very much doable.

mfelsche:
During sync it was mentioned, that keeping hashing free from allocation is very important as hashing often happens in hot-paths.

It was discussed that moving `Hashable` etc to `builtin` has some use cases outside of the `collections` package. A suggestion was made to move hashing related logic (Hashable etc. except HashMap) into a separate `hashing` package and use it from `collections`.
But as the RFC discussion in the comments already went along, given we use the `hash_with` approach, we would need the `Hashing` and `HashState` primitives/aliases available in builtin, if we want to apply this new approach to the builtin classes like numerics and String.

Will update the RFC accordingly.

mfelsche:
I did a POC of this RFC in on the [hashing_builtin](https://github.com/ponylang/ponyc/tree/hashing-builtin) branch of the pony repo and created a small benchmark to compare the different approaches to hashing (main difference is in hashing numbers and using a pure pony siphash implementation everywhere): https://gist.github.com/mfelsche/da13a70d6e4048bd70800a46f23b7ef4

TL;DR: hashing numbers is around 2 times slower, but that actually doesnt affect e.g. `HashMap.update`. Hashing Strings seems to be a bit faster.

If those benchmarks seem odd to you, I can happily redo them.
If those timings are not a blocker for anyone, I will go on and finalize this RFC and add the finishing changes.

jemc:
Do we have an understanding of what causes the perf difference for hashing numbers?

Is it more about the overhead of keeping the state, or is it more about per-operation overhead (e.g. safe arithmetic vs [unsafe arithmetic](https://tutorial.ponylang.io/expressions/arithmetic.html#unsafe-arithmetic))?

text/0000-hashing-builtin.md
```diff
@@ -0,0 +1,177 @@
...
+In order to make it more convenient for people to write useful classes
+it is favorable to advertise a common way to define hash functions for objects and offer a convenient way
+to implement them without too much boilerplate.
+
+This is a cranky, crufty and ugly, hand-crafted hash-function for the class above
+(adapted from how most IDEs in java auto-create hashCode methods):
+
+```pony
+  fun hash(): USize =>
+    var result = USiz(17)
```
damon-kwok: USiz -> USize

mfelsche:
@jemc the simple reason is that it now uses siphash as does string and array hashing (it also did before but was using a C implementation in libponyrt). Previously numbers where hashed by this small function: 

```pony
    var x = u64()
    x = (not x) + (x << 21)
    x = x xor (x >> 24)
    x = (x + (x << 3)) + (x << 8)
    x = x xor (x >> 14)
    x = (x + (x << 2)) + (x << 4)
    x = x xor (x >> 28)
    x = x + (x << 31)
    x
```

I am unable to comment on the hashing quality of this.
Whereas now it does full siphash 2-4 which consists of 6 rounds of this: https://github.com/ponylang/ponyc/compare/hashing-builtin#diff-3fcbdccf4a6378f88bdb211e3838307bR272-R286 and some more xoring and stuff.

As all hashing should operate on the same HashState, i didnt dare to mess with the given tuples and basically do the same as above on, say the first element of the tuple. I am not experienced enough to judge on what i am doing to the hashing quality of the used siphash in that case.

mfelsche:
Let me argue in favor of this RFC in case it comes to voting and i am absent:

So, while we hit a 2x performance penalty on hashing numbers, we gain a little on strings (i think because of llvm optimizations which might not be there in the compiled runtime), but most importantly we gain a means to do composite hashing of classes that has good quality, is not ad-hoc and error-prone, that is easy to use and teach, and, most importantly to me, it enables us to implement something like `value classes` or `case classes` (i can elaborate on that if necessary), with some simple syntax sugar.

jemc:
Given that `HashMap` and other collection types allow specifying a custom hash function, I'm tentatively okay with the idea of moving ahead with this RFC despite the potential performance hit for number-hashing cases - those applications for which this turns out to be a problem could use a more simplified / weaker number hashing function if they found that to be appropriate.

However, if anyone on hand feels qualified to comment, I'd love to figure out if we could somehow get the best of both worlds, and do something like this thing that you said you didn't dare to do, due to lack of knowing how hashing quality would be affected:

> As all hashing should operate on the same HashState, i didnt dare to mess with the given tuples and basically do the same as above on, say the first element of the tuple. I am not experienced enough to judge on what i am doing to the hashing quality of the used siphash in that case.

I personally would not be qualified to make a decision like this either, without more time spent researching and understanding the topic.

SeanTAllen:
@Theodus @sylvanc can either of you weigh in on this?

@Theodus I'm not sure if you know much about hasing strategies, but I think you might.

SeanTAllen:
Based on discussion during sync, I think that the "how to teach it" needs to be updated to include documenting how to change the hash function for a given object so that if the number hashing is an issue (or hashing in general is an issue).

I'm not sure what that should be. It might be part of stdlib documentation, it might be part of the tutorial. But I think it is important as part of the change given possible performance or quality impacts.

I'd like to see this change made to the RFC and fleshed out before moving this to final comment period,

jemc:
Something else to help mitigate the performance issue that we discussed in the sync, which can also be used to help with teaching efforts:

We could add a new standard library package that includes something like `SimpleHash` and `SimpleHashIs` (or whatever other name we choose), which will use the old logic for number hashing (and by extension, object identity hashing), which anyone who observes a performance regression could easily switch to using in their code. And our teaching example could show how to use this package if you want to use the "simple hash" logic.

Note that the `SimpleHash` function should be able to be implemented as a primitive with a type parameter, using `iftype` on the type argument to switch between blocks of code based on which integer type was used.

mfelsche:
I had a look at how rust does it. There the `hash` method takes a trait `Hasher` as argument. Rust's default implementation uses SipHash 2-4 https://doc.rust-lang.org/std/hash/struct.SipHasher.html as we do. The key advantage of Siphash is that it can be initiated with a random key and thus safeguard against hash flooding attacks by hashing to different values depending on the initial random key used for the hasher. Rust addresses the problem of slow hashing on numbers and in general short sequences of bytes by making it possible to switch the `Hasher` to use when creating a HashMap which then passes it into the `hash` method of the thing to create a hash from. Rust namely advertises [FNV](https://crates.io/crates/fnv) to be used for hashing numbers. See here for some performance comparisons: https://cglab.ca/~abeinges/blah/hash-rs/

I would honestly just do roughly the same, and have the `Hasher` as a type parameter that defaults to SipHash. As discussed above, we should add a SipHasher and a FNVHasher and clearly document both in stdlib and in the tutorial when to use which. I am not sure, if we can have those hashers (that define their own state to drag along through calls of `hash` or `hash_with`) be primitives, but I would try to design it that way if possible.

I am sorry for not being able to participate in a tighter and more regular frequency.

PR #158, opened 2020-01-07
Add IP address classes to net package

mfelsche:
and return them from `NetAddress` instead of their raw representation.

[Rendered](https://github.com/ponylang/rfcs/blob/ip-address-classes/text/0000-ip-address-classes.md)

Fun fact: Making the `IPv6Address` class Hashable was the actual motivation for #157

### DIFF
--- a/text/0000-ip-address-classes.md
+++ b/text/0000-ip-address-classes.md
@@ -1,0 +1,104 @@
+- Feature Name: ip_address_classes
+- Start Date: 2020-01-07
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add classes representing IPv4 and IPv6 addresses to the `net` package.
+
+# Motivation
+
+Right now, the only way to interact with IP-addresses is either via string 
+e.g. when constructing a `TCPConnection` or `TCPListener`, or when dealing with
+`NetAddress`, which is wrapping the C `sockaddr` and is thus actually a representation
+of the address of a (connected) socket. It is possible to get the contained IP address
+right now only as `U32` in case of IPv4 or as `(U32, U32, U32, U32)` in case of IPv6.
+Those representations are both very inconvenient to work with. E.g. it is especially hard to
+create a proper string representation from them.
+
+This RFC tries to overcome those inconveniences by introducing proper usable classes
+but keep efficiency in memory-representation in mind. Trade-offs...
+
+Having IP address classes will give us proper IP-address validation, easily created String representations, inspection of IP address properties.
+
+# Detailed design
+
+A class `IPv4Address` is added. Its internal state is a `U32` in network byte order.
+
+```
+class val IPv4Address is (Equatable[IPAddress] & Stringable & Hashable & Hashable64)
+  let _addr: U32
+
+  ...
+```
+
+A similar class `IPv6Address` is added. Its internal stats is a 4-element embed `Array[U32]`.
+
+Those classes have constructors for creating instances from String and convenience constructors for some special addresses, like loopback, any or broadcast addresses.
+
+They have methods for detecting if they are within a certain IP range (e.g. multicast, private etc.).
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
During sync, I asked @mfelsche to update this with an example of what is currently problematic about using the current interface.

mfelsche:
I am gonna update this RFC to not break the current 'NetAddress' methods.

mfelsche:
I updated this PR and deliberately decided for introducing a breaking change in `NetAddress`. It just didn't seem reasonable to still return the raw representation from `NetAddress.ipv4_addr()` and not the new classes. Instead I rooted for introducing new methods: `fun ipv4_addr_raw(): U32` and `fun ipv6_addr_raw(): (U23, U32, U32, U32)`.

SeanTAllen:
I would like to see 

"How we teach this" and "How we test this" expanded on. They feel very handwavey at this moment. What specifically should be tested? What specifically should be documented? Should we be updating  any package level docs with documentation? Should we add something to any of the programs in examples? Perhaps add some examples?

SeanTAllen:
I don't really know what I am signing up for with the detailed design. You appear to have a solid idea. I couldn't implement this based on the detailed design. Do you have a more solid idea of what you would do? Can you add more Pony code to demonstrate?

SeanTAllen:
From the RFC, I don't understand the point below. Why is this an alias? You seem to be talking about something that sounds like an interface to me. Can you provide some example code of what you mean and how it would be used?


A type alias `IPAddress` will be created:

```pony
type IPAddress is (IPv4Address | IPv6Address)
```

this way it is possible to handle a generic `IPAddress` using the methods common to both `IPv4Address`. and `IPv6Address`. It might be a valid alternative to use an interface `IPAddress` instead.

text/0000-ip-address-classes.md
```diff
@@ -0,0 +1,108 @@
...
+
+Proper class and method docstrings should be sufficient.
+
+# How We Test This
+
+Unit tests for constructing, parsing, validating ip addresses and range checks should be enough.
+
+# Drawbacks
+
+* Making use of an IP address requires an additional object allocation.
```
SeanTAllen, line 72: How likely are such object allocations? I think we need to do more than note that they will happen. Do we expect that this code end up being used in hot code path? IE, will this happen a lot? If yes, this is a huge drawback. If no, then its probably minor.

text/0000-ip-address-classes.md
```diff
@@ -0,0 +1,108 @@
...
+* Breaks existing code with regards to `NetAddress`.
```
SeanTAllen, line 73: I think the RFC should document what changes folks will have to do to make the change. This will better help us understand the extent of the drawback.

text/0000-ip-address-classes.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Alternatives
+
+Create primitives for IP address handling that work directly on the raw representations `U32` for IPv4 and `U128` for IPv6 for which type aliases can be created. This way no allocations are needed, but we have a very non-OO and non-intuitive API and representation.
```
SeanTAllen, line 77: I don't find non-OO to be an issue. The number 1 Pony Philosophy point is "performance", not "OO". I also don't see how the API would be "non-intuitive". Can you expand on what would be "non-intutitive" about it? Examples would favorable. 

Given the option of "no allocations", I will want no allocations unless you can show me why the drawbacks outweigh the removal of said allocations.

text/0000-ip-address-classes.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Unresolved questions
+
+Should the way for handling a generic IP address (IPv4 or IPv6) be via a type-union or an interface?
```
SeanTAllen, line 81: How is this unresolved? The detailed design says type union not interface. Seems like this is more of an alternative than an unresolved question.

PR #159, opened 2020-01-12
Change type of Env.root to AmbientAuth

mfelsche:
For more context see also [related Zulip discussion](https://ponylang.zulipchat.com/#narrow/stream/189985-beginner-help/topic/Main.20boilerplate)

[Rendered](https://github.com/ponylang/rfcs/blob/env-root-not-optional/text/0000-env-root-not-optional.md)

### DIFF
--- a/text/0000-env-root-not-optional.md
+++ b/text/0000-env-root-not-optional.md
@@ -42,6 +42,7 @@ actor Main
       env.err.print("no auth available.")
       env.exitcode(1)
     end
+```
 
 [`Env.root`](https://stdlib.ponylang.org/builtin-Env#let-root-ambientauth-val-none-val) is of type `(AmbientAuth | None)`
 and can thus possibly be `None`. This is for enabling people to create artificial `Env` instances that not need an `AmbientAuth` e.g. for testing
--- a/text/0000-env-root-not-optional.md
+++ b/text/0000-env-root-not-optional.md
@@ -1,0 +1,95 @@
+- Feature Name: env_root_not_optional
+- Start Date: 2020-01-20
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+[`Env.root`](https://stdlib.ponylang.org/builtin-Env#let-root-ambientauth-val-none-val) currently has type `(AmbientAuth | None)` to allow for
+creation of artificial `Env` instances without `AmbientAuth`.
+Yet, there is no observable usage of this feature. 
+As it is cumbersome to use, as on always needs a pattern match 
+or an `as` expression with a surrounding `try`, 
+this RFC wants to change its type to `AmbientAuth` and make `AmbientAuth` required when constructing
+an artificial `Env` instance.
+
+# Motivation
+
+In a [discussion on Zulip](https://ponylang.zulipchat.com/#narrow/stream/189985-beginner-help/topic/Main.20boilerplate) it was mentioned that having to always use `as` with surrounding `try` or a pattern-match
+for accessing `AmbientAuth` via [`Env.root`](https://stdlib.ponylang.org/builtin-Env#let-root-ambientauth-val-none-val) is cumbersome and it turned out the reason for this was
+kind of hard to justify.
+
+```pony
+
+actor Main
+  new create(env: Env) =>
+    try
+      let auth: AmbientAuth = env.root as AmbientAuth
+      // do something with auth
+    else
+      env.err.print("no auth available.")
+      env.exitcode(1)
+    end
+```
+
+```pony
+actor Main
+  new create(env: Env) =>
+    match env.root
+    | let auth: AmbientAuth =>
+      // do something with auth
+    else
+      env.err.print("no auth available.")
+      env.exitcode(1)
+    end
+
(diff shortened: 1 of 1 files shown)


jemc:
Just to add a summary of one of my comments in the Zulip thread:

While it may be desirable as a general statement to think about programs created with access to a more limited form of the `Env`, I don't see it as desirable for such a theoretical feature to make this an all-or-nothing choice between having full ambient authority and having none.

Instead I believe that to support such use cases we should have an established pattern of building some kind of a custom set of fine-grained capabilities, and passing that to your program after constructing the limited set from the `Main` actor. That is, to assume that `Main` always gets full authority, but you can build a rest of your program that uses a reduced set of capabilities, and that program could be invokable from a full-authority wrapper in Main.

To summarize the summary, while I believe in the theoretical goal of building programs that use a "principle of least privilege" reduced set of capabilities, I don't believe that the current approach of making `Env.root` optional is a good way of accomplishing this, so I have no objection to removing this bit of complexity and user friction.

SeanTAllen:
I dont have a problem with removing the friction. However, my read on the issue on Zulip was that capabilities themselves were "boilerplate". That, I do not agree with. I don't think our goal should be to make things easier for the sake of easier. After all, we can make lots of things easier by making performance harder or security harder.

I thought the AmbientAuth | None was a bit of a hack at the time, but I didn't have an issue with it. I still don't have any issues with it. I also think, it is a hack and we can and should do better. I'm interested in exploring allowing Pony programs to be started with less than AmbientAuth from the get-go. I think this would be a good thing. Something like "do not allow this program to make changes to the file-system or the network". However, we have a long way to go to get there.

I'm fine with removing the `| None`. I just want to make sure it is understood that I think that eventually allowing for Pony programs to be started with less than ambient authority means that what was called "boilerplate" will become a reality. You might not have ambient authority. You will probably have to check and see what you have and if you can accomplish what you need to as a program.  That, I'm ok.

In general, I agree with @jemc's statement:

> To summarize the summary, while I believe in the theoretical goal of building programs that use a "principle of least privilege" reduced set of capabilities, I don't believe that the current approach of making Env.root optional is a good way of accomplishing this, so I have no objection to removing this bit of complexity and user friction.

Except, I would say that making Env.root `(AmbientAuth | None) does nothing towards accomplishing what I think we should be working towards.

SeanTAllen:
I'd like to see this updated to not link out to the Zulip conversation in such a load-bearing way. I don't think this does a great job of standing on its own. I think some more of the Zulip conversation should be brought into the RFC so it can stand completely without that conversation.

Also formatting wise, our standard and I believe it applies to RFCs as well is no 80 column limit for markdown. I find the line-breaks in the RFC to feel random and it makes it hard to read. I'm not in favor of adopting this until the formatting is cleaned up.

SeanTAllen:
I'm also concerned about the statement that a single match or try is "cumbersome". I'm not willing to go on record voting for an RFC that includes that statement. I will go on record with "this | None" doesn't accomplish its goals. However, a single try or match as cumbersome sets a really bad precedent that I do not agree with.

I also can't agree to calling the AmbientAuth | None matching "added ceremony/boilerplate."

text/0000-env-root-not-optional.md
```diff
@@ -0,0 +1,96 @@
...
+
+Next to substantial changes in stdlib documentation and tutorial, examples and patterns, adapting the examples to how `Env.root` can be accessed now, 
+a small change in the docstrings of `Env`, correcting how to create an `Env` should be enough.
+
+# How We Test This
+
+# Drawbacks
+
+* Breaks existing code
+* Makes creating artificial `Env` instances a little harder.
```
SeanTAllen: How would you create an artificial env without this? "a little harder" is very vague.

text/0000-env-root-not-optional.md
```diff
@@ -0,0 +1,96 @@
...
+
+# Alternatives
+
+Just accept the way it is now. This only affects short examples in a bad way, not big code bases.
```
SeanTAllen: I disagree with the characterization of "in a bad way".

SeanTAllen:
"How we teach this" needs to be updated to identify everything that needs to be updated. 

"How we test this" is blank and needs content.

SeanTAllen:
Small nit. Env.root isn't in my mind optional. It might not provide `AmbientAuth` but that is different than optional. I think before we accept this we need to give it a more clear name as the name will appear in release notes, Last Week in Pony, etc.

text/0000-env-root-not-optional.md
ORBAT: How _would_ you create eg. a testing `Env` that doesn't need `AmbientAuth` under this proposal? As far as I can tell that'd be impossible, and I'm not exacty convinced we'd even want every test `Env` to have `AmbientAuth` when they don't need it

mfelsche: An `AmbienAuth` would be required to create an artifical `Env` with this RFC. The `root` argument to `Env.create` will be changed from `(AmbienthAuth | None)` to `AmbientAuth`.

Currently in every ponytest, one has the process `Env` available as `TestHelper.env`. From the current layout of things, one can create artifical `Env` instances in tests.

If the current ponytest design should be changed, this is a completely different discussion. Which i am totally happy to have (in another context though).

jemc:
We agreed in today's sync to mark it as final comment period.

PR #161, opened 2020-04-21
Add maybe to itertools

mfelsche:
[Rendered](https://github.com/ponylang/rfcs/blob/maybe-stdlib/text/0000-iter-maybe.md)


Theodus:
I suggest the alternative of adding similar functionality to `itertools`:
```pony
class Iter[A] is Iterator[A]
  let _iter: Iterator[A]

  // ...

  new maybe(value: (A | None)) =>
    _iter =
      object is Iterator[A]
        var _value: (A | None) = consume value
        fun has_next(): Bool => _value isnt None
        fun ref next(): A ? => (_value = None) as A
      end

  fun ref next_or(default: A): A =>
    if has_next() then
      try next()? else default end
    else
      default
    end
```
Example of usage:
```pony
let x: (U64 | None) = 21
env.out.print(
  Iter[U64].maybe(x).map[U64]({(n) => n * 2 }).next_or(0).string()) // ==> 42
```
This also gives access to all existing operations on `Iter`, such as `flat_map`, `filter`, `next`, etc.

SeanTAllen:
@Theodus I'm not super familiar with the Itertools package. A quick glance makes it look like there are a lot of allocations that pony-maybe doesn't have. Is that a reasonable first impression?

Theodus:
Yes, I think I will open a PR on ponyc to reduce that a bit this evening

Theodus:
Never mind, it seems that the performance improvements I was planning are impossible with current ponyc. I would need to be able to do something like the following:
```pony
class Iter[A, I: Iterator[A] ref = Iterator[A] ref] is Iterator[A]
```
which works, but at the callsite the `A` on the RHS of the `=` for type parameter default is not replaced with the `A` given. e.g. `Iter[U8]([0].values())` gives the following errpr:
```
type argument is outside its constraint
    Iter[U8]([0].values())
        ^
    Info:
    argument: Iterator[A #any] ref
    class Iter[A, I: Iterator[A] ref = Iterator[A] ref] is Iterator[A]
                                       ^
    constraint: Iterator[U8 val] ref
    class Iter[A, I: Iterator[A] ref = Iterator[A] ref] is Iterator[A]
                  ^
    A #any is not a subtype of U8 val: the subtype has no constraint
    class Iter[A, I: Iterator[A] ref = Iterator[A] ref] is Iterator[A]
```

mfelsche:
@Theodus I think we should track this as a new ponyc issue. Thoughts?

damon-kwok:
@mfelsche 
The union of pony types already has strong expressive power, It seems unnecessary.
````pony
let x: (String | U64 | None) = "pony"
````
Note that if maybe is added, users may do this:
````pony
Maybe[(String | U64 | None)]
````
This is obviously a bit awkward

mfelsche:
@damon-kwok the Maybe would be used like this:

`Maybe[(String | U64)]` which is equivalent to `((String | U64) | None)` which is equivalent to `(String | U64 | None)`. So, when using maybe, you can omit the `| None`.

Theodus:
> @Theodus I think we should track this as a new ponyc issue. Thoughts?

Done: https://github.com/ponylang/ponyc/issues/3540

Theodus:
I will do some further investigations into the performance comparison between the proposed `Maybe` library and the itertools equivalent.

Theodus:
I have a simple comparison here: https://github.com/Theodus/maybe-perf. It may be worth having someone more familiar with reading LLVM IR investigate.

Theodus:
The consensus from the sync call is that the Pony optimizer (LLVM) handles `Iter` much better than expected. We would like to see this functionality added to `itertools` rather than added as another package in the standard library. See https://github.com/Theodus/maybe-perf (main.pony) for some of the implementation details for the `maybe` constructor and `get_or` (feel free to rename if you desire).

mfelsche:
So, we close this RFC? I suppose the additions to `itertools` don't need to go through the RFC process. Lets track those changes with a ponyc issue then. :)

Theodus:
RFC updated to reflect the proposed extension of `itertools.Iter`

SeanTAllen:
This has been accepted however, I am not merging yet so that I won't be made to look like a liar for what I told @Theodus about automating RFC updates before anything gets merged.

rhagenson:
As was mentioned in Sync today, I previously voted for in favor of this and am still in favor of it.

PR #174, opened 2020-05-25
Parameter assignment syntax

jemc:
This PR adds my RFC for assign param syntax.

### DIFF
--- a/text/0000-assign-param-syntax.md
+++ b/text/0000-assign-param-syntax.md
@@ -55,10 +55,10 @@ class val UseFFIDecl is (AST & UseDecl)
   let partial: (Question | None)
   let guard: (IfDefCond | None)
 
-  new val create(this.name, this.return_type, this.params, this.partial, this.guard, this.attachments)
+  new val create(this.name, this.return_type, this.params, this.partial, this.guard, this.attachments) => None
 ```
 
-Note that there is now no syntactical body of the method, but the compiler will add a body that does the appropriate assignments.
+Note that there is now no syntactical body of the method apart from the obligatory `None` expression, and the compiler will add a body that does the appropriate assignments.
 
 Of course, "assign parameters" can be used alongside normal parameters, and not every field need be associated to an "assign parameter". Assign parameters can also have a "default value" as normal parameters can, which specifies the value to use if the caller does not supply an argument for that position. The following refactored example demonstrates all of those concepts:
 
--- a/text/0000-assign-param-syntax.md
+++ b/text/0000-assign-param-syntax.md
@@ -9,8 +9,6 @@ The syntax change would allow a field to be referenced in the position of a para
 
 # Motivation
 
-*Note: The syntax examples and text of this RFC assume that the RFC for "explicit self member access" with the `@` character have already been accepted and included in the language. If you haven't read the other RFC yet, please review that first, then read this one with that context.*
-
 In current Pony syntax, there can be significant code redundancy required to build simple data classes that take the arguments of their constructors and simply assign them to the field of the same name.
 
 One must declare the field, with its name and type, then declare the parameter with the same name and type in a particular position, then write an assignment statement in the body of the method.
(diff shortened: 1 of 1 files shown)


aturley:
Discuss more next week.

text/0000-assign-param-syntax.md
```diff
@@ -0,0 +1,115 @@
...
+
+# How We Test This
+
+New compiler tests that demonstrate this case, including the edge cases mentioned in the implementation details above.
+
+# Drawbacks
+
+* New patterns to learn (though no codebases will be required to use these patterns, you have to learn the new patterns if you want to write or read code that uses them).
+* Added compiler complexity.
+
```
jasoncarr0, line 106: Symbols/glyphs are hard to understand and search for, so this makes code less explicit and thus possibly harder to learn, or more intimidating

jemc: I personally don't really buy this argument in this case, because if we used the `@` symbol for `this.` (and if `this.` were mandatory), as proposed in the other RFC, this symbol would be used in nearly every function of every file of every Pony program.

Sure, somebody who was learning the language for the first time won't immediately know what it means, but they'd learn what it means within the first 5-10 minutes of learning, and they wouldn't forget because it would be so ubiquitous. Searchability of the symbol in search engines is not really relevant, because anyone encountering it for the first time would also be learning the language for the first time, so they'd already be looking at the tutorial.

I actually think that your argument is a much more compelling argument *in favor* of RFC #173 instead of against it - that is, it is a compelling argument for us to *not* use `@` for FFI or bare functions (which are rare/arcane), and to use it for the much more ubiquitous `this.` instead. FFI code is much more obscure, and a user could be using Pony for years before encountering their first reading of FFI code. Hence, if and when they do encounter the `@` symbol for FFI or a bare function they are likely to be confused, and find it hard to search for.

Note that in the Savi language I followed this design principle - it uses the `@` symbol for self member access, but FFI usage and declaration has no special symbol - FFI looks like a normal type declaration at the declaration site, and a normal call at the use site - it just happens to be a type declaration that is marked as being `:ffi` rather than some other kind of type. In this regime, users can quickly adapt to understanding `@`, but when they get to the more obscure concepts like FFI there are no new symbols to learn / search fo.

SeanTAllen:
@jemc how would you feel about moving forward with `this.` instead of `@`? I'd like to move this and #173 forward but we have some syntax issues to resolve. I'm in favor of "this" because it is the least breaking option.

jemc:
I'd agree that `this.` is "better than nothing" for the purposes of this RFC, so I'd be good with amending this RFC to use that syntax. It's still less boilerplate than the current status quo.

For #173 I would not be in favor of making that change with `this.` - I think it's too much friction added for something so ubiquitous.

SeanTAllen:
We appear to have an impasse on #173, but it would be good to move forward with this RFC so let's do that and continue to discuss #173.

text/0000-assign-param-syntax.md
SeanTAllen: This now has `this.` instead of `@` so this is resolved for now.

text/0000-assign-param-syntax.md
```diff
@@ -0,0 +1,113 @@
...
+  let name: (Id | LitString)
+  let return_type: TypeArgs
+  let params: Params
+  let partial: (Question | None)
+  let guard: (IfDefCond | None)
+
+  new val create(name: String, this.return_type, this.params, this.attachments = None) =>
+    this.name = Id(name)
+    this.partial = None
+    this.guard = None
```
jasoncarr0, line 77: What happens with this PR if the body of the constructor is now empty? Do we still need to include `=> None`?

SeanTAllen: My expectation would be yes, you need to. I think it is worth explicitly noting that in this RFC.

text/0000-assign-param-syntax.md
```diff
@@ -0,0 +1,113 @@
...
+```pony
+class val UseFFIDecl is (AST & UseDecl)
+  let attachments: (Attachments | None)
+  let name: (Id | LitString)
+  let return_type: TypeArgs
+  let params: Params
+  let partial: (Question | None)
+  let guard: (IfDefCond | None)
+
+  new val create(this.name, this.return_type, this.params, this.partial, this.guard, this.attachments)
```
SeanTAllen: this should have `=> None` yes?

SeanTAllen:
Is there an implicit consume of `iso` parameters?

text/0000-assign-param-syntax.md
jemc: I've updated it to explicitly mention that requirement.

text/0000-assign-param-syntax.md
```diff
@@ -55,10 +55,10 @@ class val UseFFIDecl is (AST & UseDecl)
...
   let guard: (IfDefCond | None)
 
-  new val create(this.name, this.return_type, this.params, this.partial, this.guard, this.attachments)
+  new val create(this.name, this.return_type, this.params, this.partial, this.guard, this.attachments) => None
 ```
 
-Note that there is now no syntactical body of the method, but the compiler will add a body that does the appropriate assignments.
+Note that there is now no syntactical body of the method apart from the obligatory `None` expression, and the compiler will add a body that does the appropriate assignments.
 
 Of course, "assign parameters" can be used alongside normal parameters, and not every field need be associated to an "assign parameter". Assign parameters can also have a "default value" as normal parameters can, which specifies the value to use if the caller does not supply an argument for that position. The following refactored example demonstrates all of those concepts:
```
jasoncarr0, line 63: One more note possibly worth adding: are any names supported for keyword argument syntax?

jasoncarr0:
One more note possibly worth adding: are any names supported for keyword argument syntax? Does that create semver issues with the name requirements?

SeanTAllen:
It was noted on the sync call that @jemc will note that how this interacts with named arguments in unresolved questions.

jasoncarr0 (approved):

SeanTAllen:
This was accepted during sync. I will do the merge and what not later.

PR #175, opened 2020-06-07
Add log_level argument to LogFormatter.apply()

rkallos:
[Rendered](https://github.com/rkallos/rfcs/blob/add-log-level-to-logformatter/text/0000-add-log-level-argument-to-logformatter.md)

### DIFF
--- a/text/0000-add-log-level-argument-to-logformatter.md
+++ b/text/0000-add-log-level-argument-to-logformatter.md
@@ -1,0 +1,78 @@
+- Feature Name: add-log-level-argument-to-logformatter
+- Start Date: 2020-06-07
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add a `log_level` argument to logger.LogFormatter.
+
+# Motivation
+
+In an attempt to replace Corral's custom logging facilities with the standard
+library logger, it was discovered that prepending the log level to log messages
+would be much easier if the log level was exposed in LogFormatter.apply().
+
+# Detailed design
+
+- Add a `log_level` argument to `LogFormatter.apply()` with type `LogLevel`.
+- Change `Logger.log` to add `_level` to `_formatter.apply()`
+- Change `DefaultLogFormatter.apply` to accept (and ignore) the new `log_level`
+  argument.
+
+# How We Teach This
+
+This RFC could be announced as a note in last week in Pony. If the RFC is
+accepted and subsequently implemented, the addition of the new argument to
+`LogFormatter.apply()` could be announced as a breaking change in the release
+notes of a new release of ponyc.
+
+Upon upgrading to the new release of ponyc, users will be faced with
+compile-time errors if their code contains implementations of the old
+`LogFormatter` interface.
+
+# How We Test This
+
+- CI should pass for the ponyc repo after the changes are implemented.
+- The test suite for the logger package has its own implementation of
+LogFormatter that will have to change as the LogFormatter interface changes.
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I'm in favor of this, however, I'd like to see if expanded so that each log level implements Stringable so that it will be easy for the common use case for this to go:

log_level.string()

to get the string representation like "INFO", "ERROR" etc.

adri326:
We could make this non-breaking by placing the log_level argument in second place and giving it a default value (INFO for instance). I'm not sure how others feel about default values in the standard library though.

SeanTAllen:
I think it is worth being non-breaking. I don't think a default value there is a good idea. That's a good way to hide a user bug.

rkallos:
The change could become non-breaking and avoid taking a stance on what the default value should be if we have `log_level: LogLevel | None = None` in LogFormatter.apply().

SeanTAllen:
@rkallos also not in favor of that. I'm not worried about the breaking change nature. We are pre-1.0 and its an easy breaking change to adjust to.

SeanTAllen:
We discussed at sync. 

@jemc, @sylvanc, @Theodus and @SeanTAllen are all for this if the log level is the first argument and its a breaking change. 

@rkallos can you update accordingly?

damon-kwok:
> Let’s talk about logging
https://dave.cheney.net/2015/11/05/lets-talk-about-logging

What’s so bad about the stdlib’s log package?
https://forum.golangbridge.org/t/whats-so-bad-about-the-stdlibs-log-package/1435

what.Happens - a debug logging package for developers only
https://appliedgo.net/what/

SeanTAllen:
@damon-kwok I think changing the log levels in the logger is out of scope of this RFC.

damon-kwok:
@SeanTAllen Yes, I think...If this proposal is valuable, then the expansion of the apply method of log is meaningless.
About log levels, We can go to zulip discussion: https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/About.20LogLevel.20(PR-175)

SeanTAllen:
Please copy over any relevant comments from the Zulip here. Not everyone who will interact with this RFC is guaranteed to have access to the Zulip.

This RFC solves an immediate problem with the usability of existing code.

Your linked articles are "I think there should be fewer log levels". Fine, ok, do a library that has fewer.
There's nothing meaningless about this RFC. It solves an actual problem with the existing logger library that was encountered in Corral. Nothing you are proposing is required to solve that problem.

Changing the number of log levels is irrelevant to exposing what the log level is to the formatter.

damon-kwok:
I see. I will submit a department PR in the future.

SeanTAllen:
Sorry @damon-kwok, what did you mean by "department PR". I think this is the rare situation where what said isn't what you intended. Did you mean "separate PR" as in, "PR unrelated to this one"?

SeanTAllen:
@rkallos I'm in favor of this RFC as it is now written. I'll request it be moved to final comment period at the next sync if everyone else on the call is also in favor.

damon-kwok:
@SeanTAllen  I mean: I will try implement a new log system, and then give it to everyone to use, and then submit an RFC about log level.

SeanTAllen:
@rkallos are you up for doing the final updates to this?

rkallos:
I'd be delighted to! What needs to be done?

SeanTAllen:
@rkallos Ah, you already made the changes. Nevermind. Apparently I don't read so good these days. Can I blame the pandemic?

SeanTAllen:
We approved this at sync today. Thank you for writing this up @rkallos.

PR #176, opened 2020-06-10
Add new collection type without generic constraints

damon-kwok:
Add new collection type without generic constraints:
- SortBy : https://github.com/damon-kwok/pony-shoe/blob/master/sort_by.pony
- AnyMap : https://github.com/damon-kwok/pony-shoe/blob/master/any_map.pony
- AnySet : https://github.com/damon-kwok/pony-shoe/blob/master/any_set.pony

### DIFF
--- a/text/0000-new-collection-type-without-generic-constraints.md
+++ b/text/0000-new-collection-type-without-generic-constraints.md
@@ -1,0 +1,64 @@
+- Feature Name: new-collection-type-without-generic-constraints
+- Start Date: 2020-04-27
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add `SortBy` `AnyMap` and `AnySet` to `collections` package.
+
+# Motivation
+
+`SortBy` primitive like `Sort`, but it use a `lambda` replaced `interface Comparable`. it can use the any type for Key.
+`AnyMap` like `Map`, but it use a `lambda` for hash calculation. it can use the any type for Key.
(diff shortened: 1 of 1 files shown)


SeanTAllen:
My criteria for adding something to the standard library is that at least one of the following needs to be met:

- Functionality is existing library in widespread usage that has become a defacto standard and including it in the standard library solves a pain point
- The functionality addresses a common beginners pain point
- Using the functionality as a 3rd party library is painful, but that pain could be resolved if it is part of the standard library

SortBy isn't a new collection type so I think it is out of place in this RFC and should be moved to its own. It's a general helper primitive for doing a specific operations on Arrays. As such, I think it would be better as part of a primitive that provides different array helper methods such as copying, sorting etc that might come up for beginners. A well done starter primitive is something I can see (particularly with copying and sorting) coming up often from beginner.

The AnyMap and AnySet, I don't see what pain points they address. Without more explanation for the problem that is solved (this should be in motivation section), I'm not in favor of them.

At best, they seem like yet another way to do something in the standard library that could lead to confusion. Without more motivation, I'm inclined to say they should 1 or more 3rd party libraries.
They've seen little to no community usage so far and might need to undergo changes as they are used. Being part of the standard library and being subject to the RFC process would slow down their development towards a widely used library.

This is a key part of my "- Functionality is existing library in widespread usage that has become a defacto standard and including it in the standard library solves a pain point" metric. If a library is a defacto standard, its API has probably matured and as such, the potentially slower pace of development that being in the standard library entails would not be detrimental.

damon-kwok:
Ok. I will separate SortBy separately.

PR #184, opened 2021-02-19
Add mandatory FFI declarations proposal

ergl:
[Rendered](https://github.com/ergl/rfcs/blob/caff9b3554e7afa9a2e90c6c6c614255e5313b59/text/0000-force-declaration-of-ffi-functions.md)

### DIFF
--- a/text/0000-force-declaration-of-ffi-functions.md
+++ b/text/0000-force-declaration-of-ffi-functions.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-This proposal aims to make mandatory the declaration of all C-FFI (Foreign Function Interface) functions through `use` statements, such that the compiler is able to distinguish variadic functions from regular functions.
+This proposal aims to make mandatory the declaration of all C-FFI (Foreign Function Interface) functions through `use` statements, such that the compiler is able to distinguish variadic functions from regular functions. In addition, it also aims to remove from the language the ability to specify the return type of an FFI function at the call site, as this is made redundant in the face of mandatory declarations.
 
 # Motivation
 
@@ -24,6 +24,8 @@ In the absence of an explicit declaration, Pony will assume the function is vari
 Making declarations mandatory only for variadic FFI functions would make it hard for the Pony compiler to detect undeclared variadic C functions, given that it is impossible to distinguish a regular function from a variadic function that is always called with the same number of arguments. This limitation would restrict the ability of the Pony compiler to give meaningful errors to users. Making declarations mandatory only for variadic FFI functions is also confusing for users, as the choice of making some declarations mandatory while others are kept optional can feel arbitrary unless they have explicit knowledge of calling convention differences.
 
 Given that forgetting to write `use` declarations for variadic functions could cause Pony programs to crash at runtime only when run on specific platforms, this proposal aims to make `use` declarations mandatory for _all_ FFI functions, variadic or not. In this way, the Pony compiler will be able to give meaningful error messages when a FFI function is used, but not declared via a `use` statement.
+
+With mandatory declarations in place, specifying the return type for an FFI function at the call site is made redundant, so it should be removed from the language to minimise syntax noise.
 
 # Detailed design
 
@@ -45,11 +47,35 @@ This RFC proposes to remove option (3) above. The default behaviour by the compi
 
 Since we expect the compiler to catch any possible errors that can arise at runtime, it is expected that the compiler enforces mandatory `use` declarations, at least for variadic functions. However, from the point of view of the compiler, it is impossible to distinguish regular functions from variadic ones that happen to always use the same number of arguments. Merely changing (3) above so that the compiler generates FFI functions as if they had a fixed number of parameters by default, and relying on heuristics to know if a function is variadic, will not work. It is for this reason that this RFC aims to enforce declarations for _all_ FFI functions.
 
+## Removing explicit return type declarations at call sites
+
+Currently, one can specify the return type of an FFI function at the call site, as such:
+
+```pony
+let foo = @printf[I32]("some words here".cpointer())
+```
+
+If a declaration for an FFI function is not provided, the compiler enforces that an explicit return type is specified, since otherwise it wouldn't have enough information to generate the correct function signature. If the programmer provides an explicit `use` declaration, specifying the return type at the call site is still allowed, with the compiler attempting to type-check the types against the original declaration.
+
+However, if `use` declarations are made mandatory for all FFI calls, the use of the return type at the call site is made redundant. As such, if this RFC is accepted, the option to specify the return type of an FFI call will be removed, and the compiler should treat any explicit return type declarations as a syntax error. Keeping the option of specify return types at the call site could be confusing to the users: specifying the correct return type would not accomplish anything, so it can only become a source of unintentional type-check errors.
+
 ## Compiler changes
+
+Changes related to enforcing FFI declarations:
 
 1. In the `expr` pass: the `expr_ffi` function in `ffi.c` should ensure that an explicit declaration exists for all FFI calls. If a declaration is not found, it should emit an error, stopping the compilation process. Intrinsic and internal functions should also require declarations.
 
-2. Code generation: the `gen_ffi` function in `gencall.c` won't need to generate FFI declarations on the fly, and as such it can be simplified, since it will always have enough information to generate the correct function for LLVM. Intrinsic and internal functions will no longer be generated as regular functions by default, the compiler will use the provided declarations.
+2. In the `reach` pass: the `reachable_ffi` function in `reach.c` won't need to defer reification of an FFI call due to having an undeclared FFI declaration.
+
+3. Code generation: the `gen_ffi` function in `gencall.c` won't need to generate FFI declarations on the fly, and as such it can be simplified, since it will always have enough information to generate the correct function for LLVM. Intrinsic and internal functions will no longer be generated as regular functions by default, the compiler will use the provided declarations.
+
+Changes related to removing explicit return types on FFI calls:
+
+1. In the `name` pass: the `ffidecl` rule in `treecheckdef.h` should change to remove the return type.
+
+2. In the `reach` pass: the `reachable_ffi` function in `gencall.c` should remove any references to the return type of the FFI call.
+
+3. Code generation: the `declared_ffi` function in `ffi.c` should change to remove any references to the return type of the FFI call. This function can be simplified to remove any type-checking related to the return type of the function.
 
 ## Standard Library, example, test changes
 
@@ -151,7 +177,9 @@ Downloaded all repos under the ponylang org, ran grep, then removed some false p
 
     In particular, the ["Calling C from Pony" section](https://tutorial.ponylang.io/c-ffi/calling-c.html) should note that a declaration is needed in order to call a FFI function, as well as introduce the `...` syntax for declaring variadic functions. Although we could expect that users that attempt to use the C FFI will be familiar with the concept of variadic functions, it wouldn't hurt to mention them, even if we don't use the "variadic" term.
 
-- All the examples that show FFI code, be it in the `examples/` folder of the compiler or on the tutorial should be changed to include mandatory declarations. These examples are located in the C FFI and [Serialisation](https://tutorial.ponylang.io/appendices/serialisation.html) chapters, and in the [examples appendix](https://tutorial.ponylang.io/appendices/examples.html).
+    The chapter should also remove any explicit type declarations from FFI calls.
+
+- All the examples that show FFI code, be it in the `examples/` folder of the compiler or on the tutorial should be changed to include mandatory declarations, and remove return types at the call site. These examples are located in the C FFI and [Serialisation](https://tutorial.ponylang.io/appendices/serialisation.html) chapters, and in the [examples appendix](https://tutorial.ponylang.io/appendices/examples.html).
 
 In general, the use of the C-FFI should be considered an advanced feature, as its usage can violate many of the safety properties enforced by the language. As such, not much would change for regular Pony users.
 
@@ -164,6 +192,8 @@ In general, the use of the C-FFI should be considered an advanced feature, as it
 * Compiler tests can/should be written to demonstrate that programs that lack explicit FFI declarations fail to compile.
 
 * Apple ARM64 should be added as a target platform for CI when our current provider (Cirrus CI) offers instances of this type. In the meantime, the Pony team owns two machines with Apple Silicon chips that should serve as testing infrastructure.
+
+* We should ensure that `use` scoping rules allow us to remove explicit return types on FFI calls.
 
 # Drawbacks
 
@@ -182,6 +212,3 @@ In general, the use of the C-FFI should be considered an advanced feature, as it
 # Unresolved questions
 
 Exact implementation details are still to be determined.
-
-If declarations are mandatory, specifying the return type at FFI call site is made redundant, and maybe should be removed from the language.
-
--- a/text/0000-force-declaration-of-ffi-functions.md
+++ b/text/0000-force-declaration-of-ffi-functions.md
@@ -1,0 +1,96 @@
+- Feature Name: Force declaration of FFI functions
+- Start Date: 2021-02-17
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+This proposal aims to make mandatory the declaration of all C-FFI (Foreign Function Interface) functions through `use` statements, such that the compiler is able to distinguish variadic functions from regular functions.
+
+# Motivation
+
+Currently, the Pony compiler treats all undeclared C-FFI functions as variadic, even if the function has a fixed number of arguments (a regular function). This doesn't cause a problem in any of the currently supported platforms, since their calling conventions allow to treat regular functions as variadic, and vice versa.
+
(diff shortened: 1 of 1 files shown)


text/0000-force-declaration-of-ffi-functions.md
```diff
@@ -0,0 +1,96 @@
...
+
+* Use an alternative syntax to let the compiler know about variadic functions at the call site, so that a declaration is not needed, such as (for example) `@printf[I32](...)(arg1, arg2)`.
+
+* Leave the compiler as it is, and do not support platforms with different calling conventions.
+
+# Unresolved questions
+
+Exact implementation details are still to be determined.
+
+If declarations are mandatory, specifying the return type at FFI call site is made redundant, and maybe should be removed from the language.
```
jemc, line 92: Agreed. :+1:

jemc: Agreed :+1:

text/0000-force-declaration-of-ffi-functions.md
```diff
@@ -0,0 +1,96 @@
...
+
+Should C functions provided by the runtime (`@pony*`) be declared explicitly?
```
jemc: For consistency, especially to avoid confusion arising from inconsistency, I would say yes.

text/0000-force-declaration-of-ffi-functions.md
```diff
@@ -0,0 +1,96 @@
...
+
+Should LLVM (`@"llvm.*"`) and internal (`@"internal.*"`) functions be declared explicitly?
```
jemc: I'm less sure about this one, but again I'd lean toward consistency (requiring mandatory `use` declarations).

The convenience of omitting declarations for these intrinsics is not important for the common user because I don't know of any user code that uses these intrinsics directly - only the standard library, for which the convenience of omitting a few declarations can be considered less of a concern.

jemc (commented):
Thanks, a very well-written RFC! :clap:

I've provided my opinion on the unresolved questions below.

text/0000-force-declaration-of-ffi-functions.md
SeanTAllen: Also agreed.

text/0000-force-declaration-of-ffi-functions.md
SeanTAllen: Agreed. Yes.

text/0000-force-declaration-of-ffi-functions.md
SeanTAllen: I think for consistency, this should be required. It should make all FFI handling code easier at the cost of a few declarations in the standard library. Not doing this seems like a potential source of bugs.

SeanTAllen:
This is a great RFC. I'm in favor. 

I'd like to see a couple bits of bookkeeping added...

- listing of all files in the standard library that use FFI functions and will need to be updated. partially because this will be valuable to the implementer and partially because it might expose an issue we haven't considered
- same listing for all pony libraries maintained by the `ponylang` organization

by including in the RFC (particularly the latter set), the RFC wouldn't be considered completed until all needed updates were done.

text/0000-force-declaration-of-ffi-functions.md
ergl: Great. Do you think this should be included in the RFC itself?

text/0000-force-declaration-of-ffi-functions.md
ergl: Ok. Including these as well as the runtime functions is good for consistency. It would save us if in the future we end up introducing a variadic function in the runtime, or if LLVM itself introduces one.

ergl:
Thanks for the encouragement everyone, I've left some replies inline. 

> * listing of all files in the standard library that use FFI functions and will need to be updated. partially because this will be valuable to the implementer and partially because it might expose an issue we haven't considered
> * same listing for all pony libraries maintained by the ponylang organization

@SeanTAllen I agree with that, I will see if I can include that in the RFC itself.

text/0000-force-declaration-of-ffi-functions.md
ergl: Also, thinking about this a bit, we should ensure that the current scope rules for `use` statements are enough to be able to remove explicit return types, or if something is needed to change to support it.

SeanTAllen: I think it makes sense to be part of this RFC because each is a breaking change, best to do them return type at call-site removal at the same time.

SeanTAllen: > Also, thinking about this a bit, we should ensure that the current scope rules for use statements are enough to be able to remove explicit return types, or if something is needed to change to support it.

Agreed.

text/0000-force-declaration-of-ffi-functions.md
SeanTAllen: I hope they don't and we don't but... yes, excellent point.

text/0000-force-declaration-of-ffi-functions.md
ergl: Added to the RFC

text/0000-force-declaration-of-ffi-functions.md
ergl: Added to the RFC

text/0000-force-declaration-of-ffi-functions.md
ergl: Included in the RFC

ergl:
I've included removing return type declarations at call site, and included some tentative pointers to all the files in the ponyc repo and other pony libraries in the ponylang org that would need to be updated. 

Someone else that is more familiar with the compiler could probably add more details to the "Compiler changes" section, I have taken a look at all the places that this RFC would need to change, but it's quite likely that I might have missed some things.

ergl:
After investigating a bit, I've found that we might have to think more about if/how to write FFI declarations for intrinsic functions. We're currently overloading several intrinsics in the standard library, mainly to force LLVM to return either signed or unsigned integers, for example:

```pony
// unsigned.pony
primitive U8
  fun bit_reverse(): U8 => @"llvm.bitreverse.i8"[U8](this)

// signed.pony
primitive I8
    fun bit_reverse(): I8 => @"llvm.bitreverse.i8"[I8](this)
```

This gets trickier with 32/64 bit types, where an intrinsic function might be used not only to distinguish the sign, but also the type of integer when returning `U32/U64/USize/ULong` (same applies for `I32/I64/ISize/ILong`).

```pony
primitive U32
  fun bit_reverse(): U32 => @"llvm.bitreverse.i32"[U32](this)

primitive ULong
  fun bit_reverse(): ULong =>
    ifdef ilp32 or llp64 then
      @"llvm.bitreverse.i32"[ULong](this)
    else
      @"llvm.bitreverse.i64"[ULong](this)
    end

primitive USize
  fun bit_reverse(): USize =>
    ifdef ilp32 then
      @"llvm.bitreverse.i32"[USize](this)
    else
      @"llvm.bitreverse.i64"[USize](this)
    end
```

The Pony compiler doesn't allow you to specify different return types for different calls of the different FFI functions, unless LLVM can ensure that the different types:

1. [Have the same ABI size](https://github.com/ponylang/ponyc/blob/08e82205917a6e9174496e4a656813ff886a739c/src/libponyc/codegen/gencall.c#L1109-L1113), or

2. [Can be safely casted to each other](https://github.com/ponylang/ponyc/blob/08e82205917a6e9174496e4a656813ff886a739c/src/libponyc/codegen/gencall.c#L1116-L1132)

However, these checks only occur during the code generation phase. If the user supplies a FFI declaration, the check against the returned type is done [via `is_eqtype`](https://github.com/ponylang/ponyc/blob/08e82205917a6e9174496e4a656813ff886a739c/src/libponyc/expr/ffi.c#L132-L141), which only checks equality via subtyping rules.

---

Now, one could get around this via casting on the Pony side, like in the examples below. I don't know if this would impose any kind of overhead (since conversions are done via `compile_intrinsic`), and opens us up to mistakes if one is not careful (although this applies to the current situation, where we're forcing LLVM to cast types for us).

```pony
use @"llvm.bitreverse.i8"[U8](src: U8)
use @"llvm.bitreverse.i32"[U32](src: U32)
use @"llvm.bitreverse.i64"[U64](src: U64)

primitive I8
    fun bit_reverse(): I8 => @"llvm.bitreverse.i8"(this.u8()).i8()

primitive U8
  fun bit_reverse(): U8 => @"llvm.bitreverse.i8"(this)

primitive ULong
  fun bit_reverse(): ULong =>
    ifdef (ilp32 or llp64) then
      @"llvm.bitreverse.i32"(n.u32()).ulong()
    else
      @"llvm.bitreverse.i64"(n.u64()).ulong()
    end  
```

text/0000-force-declaration-of-ffi-functions.md
```diff
@@ -0,0 +1,214 @@
...
+* Imposes overhead on users to declare the type of FFI functions in advance.
+
+# Alternatives
+
+* Make the compiler treat FFI functions as regular functions by default, and require explicit declaration only for variadic FFI functions.
+
+* Use an alternative syntax to let the compiler know about variadic functions at the call site, so that a declaration is not needed, such as (for example) `@printf[I32](...)(arg1, arg2)`.
+
+* Leave the compiler as it is, and do not support platforms with different calling conventions.
+
```
jasoncarr0, line 231: One other alternative is to be able to handle header files from C in some way for this use case. Would keep it convenient for most use cases

SeanTAllen: I think that's a good add on feature for FFI. I'm not sure I would call it an alternative given what a large change it would be from the existing FFI mechanisms.

jasoncarr0: My thought is that it could be used to give enough information to continue handling FFI without "use declarations"

jemc:
@ergl Regarding the issue with LLVM intrisincs, here's an idea:

We could possibly resolve this use case and also open up some new use cases by adding a feature to have an FFI "alias" declaration.

For example, you could declare the `llvm.bitreverse.i8` function with the `I8` based signature, then you could also declare an alias called `llvm.bitreverse.u8` (an alias of `llvm.bitreverse.i8`), which has the `U8` based signature.

This could also be useful as a convenience for cases where you are say, wrapping a C library that has a consistent prefix on all of its C functions - you could define aliases which omit the prefix, making all of your calling code slightly less verbose at expense of making the declarations more verbose.

What would syntax look like for this? I'm not sure, honestly.

text/0000-force-declaration-of-ffi-functions.md
SeanTAllen: What's really nice about the current is that I can compile for any supported platform and it will work and type check without needing any additional headers etc.

I've never installed windows headers on my machine but the declarations give me type checking.

I love that if a pony program type checks on one platform, it type checks on all.

I don't want to lose that.

The FFI declarations are "trust me" for that type checking but I still LOVE what it gives me and do not want to give it up.

redvers: The selfish part of me would love to be able to take C header files in lieu of FFI declaration proposals because I spend a non-trivial amount of time converting header files into these types of declarations.  It's non-trivial both because of the number of functions I'm dealing with, but also the number of macros and other cpp tomfoolery that's found in many C libraries.

With 'self-authored' FFI declarations it's type-checking my function call against the what I understand the C function prototype to be.

If it's generated from the header-file, then it removes human error from another sensitive part of the process which is a good thing for software reliability - especially when juggling knives with C-FFI (by definition) at runtime.

In order to not lose the platform independence functionality that you value so much @SeanTAllen - would/could/should an independent tool which converts C header files into use statements exist?

Then potentially we could have the best of both worlds.

redvers: (How one would go about writing such a thing, (and if it's even feasible) is left as an exercise for the reader?)

SeanTAllen: An independent tool could do that. I think it should be optional.

I shouldn't be required to have headers installed to type check.

I don't think there is anything preventing folks from developing such a tool now.

redvers: (arguably, this is the same hard problem that I've been struggling with for the last month or so)

SeanTAllen: Another thing we currently have is you can support multiple API versions of the same library like OpenSSL and type check them all in one go.

That's built on the same foundation as the "type check all platforms" functionality I mentioned earlier.

jemc: The thing about the approach of scanning C headers / type signatures, in addition to the extra compiler complexity and the header dependency problem @SeanTAllen mentioned, is that it does not necessarily give enough information to wrap the calls for use in Pony. There are various cases of ambiguity that need an extra information channel (such as human-interpreted documentation, or something like the GIntrospection format that @redvers has been using) to resolve to the most appropriate Pony FFI use patterns.

ergl: I agree that this would be a useful feature for Pony, but I have to side with Sean, I don't think it should be a requirement to have headers installed to use the FFI. Right now I can write some windows-dependent code on unix* platforms by reading the reference on msdn, and the compiler will check the types. It's not a guarantee that the code will work, but I'd like to keep that option open.

ergl:
@jemc I'm not sure how aliases would work from the typing point of view. An alias would need to be compatible with the old declaration, either using subtyping rules or using the ABI size / casting check that the compiler does now for FFI calls.

jemc:
Yes, the ABI size check is what I was expecting. This would let you have different Pony typings declared for the same ABI function, as in the case of the LLVM intrinsics, as well as some other cases where you might want to vary things on the Pony typings side based on, say, opaque struct pointer patterns vs non-opaque struct pointer patterns.

jemc:
So from the Pony type system standpoint they would be distinct functions with different type signatures, but at the FFI level we understand them to be referencing the same FFI function and enforce the compatible ABI types.

ergl:
I see, so right now we have the following rules:

1. Two declarations for the same FFI function have to use the same types (exactly the same, the compiler checks that the both declarations are the same `decl` pointer)

2. A declaration and a call might differ in the type if the types are equal (using `is_eqtype`)

We remove the second one, and change it to:

2. A FFI declaration might alias another FFI declaration as long as the types are compatible:

    - Either they are the same (using `is_eqtype`)
    - ABI size check
    - Can be safely casted to each other (?)

I'm not so sure about the last one, but the compiler already allows casting ints to pointers (and vice versa) or between different `struct` types.

jemc:
I'm not sure how "safely casted" is being used in this context. Other than that, it sounds right.

jemc:
We discussed on the sync call today.

We are wary of adding the "alias" idea into scope of the RFC.

@ergl brought up again his suggestion to use explicit numeric conversions in the standard library, like in the last example in this comment of his above (https://github.com/ponylang/rfcs/pull/184#issuecomment-785063230).

After looking at this closer, I agree that this is probably the best approach, and should be zero runtime cost (All of the mentioned numeric conversions will be using the same LLVM type internally, so there will be no runtime cost or change in behavior from adding those conversion).

ergl:
I double-checked all the FFI code in the compiler and in the ponylang org to verify that all cases of FFI calls with multuple return types were due to numeric types. There were some examples of non-numeric types, described below, but are easily fixed. So I think it's safe to go with casting on the Pony side.

Only the compiler contained mismatched types for FFI calls, most related to LLVM intrinsics, but there were other cases in the standard library:

1. Most variants are signed / unsigned casts of the same type: `I8` to `U8` or viceversa. 

2. Integer width on 32/64 bit plaftforms: cast between `I32`/`I64` and `ISize`/`ILong`. (This also happens for tuple returns like casting `(U64, Bool)` to `(ULong, Bool)`.

3. Both of the above: `I32` to `ULong` / `USize`.

4. Casting pointer types: `Pointer[U8]` to `Pointer[None]` and viceversa. (cases: `@pony_os_stderr` and `@pony_os_stdout`).

5. Other: `USize` to `U8`, `USize` to `Pointer[None]`. These are found in Windows-related code, mainly modeling the `HANDLE` type as either [`USize`](https://github.com/ponylang/ponyc/blob/dbe4bfbc3caf0ef5337cda0923d6900a5fec09ef/examples/fan-in/main.pony#L208), [`U8`](https://github.com/ponylang/ponyc/blob/main/packages/files/file_path.pony#L306) or [`Pointer[None]`](https://github.com/ponylang/ponyc/blob/dbe4bfbc3caf0ef5337cda0923d6900a5fec09ef/packages/files/file.pony#L11). These cases could be easily fixed by picking any of the above types, as they are always treated as an opaque type from the Pony point of view.

ergl:
I've added the type-casting discussion to the RFC. I think this RFC would be ready to be voted on if no one else has any comments on the latest version I pushed.

SeanTAllen:
I feel like this RFC is in a good place and should be moved to final comment period and then voted on at the next sync.

I'm in favor of the RFC as it currently stands.

aturley:
The vote was taken in the 2021-03-30 sync and it passed.

SeanTAllen:
@Theodus do you have time to do the RFC shuffle?

Theodus:
> @Theodus do you have time to do the RFC shuffle?

Yup, I'll get to it later today

PR #192, opened 2021-11-23
WIP RFC for Unix domain socket support

sgebbie:
Here we provide an RFC related to introducing Unix domain socket support into the `net` package.

### DIFF
--- a/text/0000-unix-domain-sockets.md
+++ b/text/0000-unix-domain-sockets.md
@@ -1,0 +1,64 @@
+- Feature Name: Unix Domain Socket Support
+- Start Date: 2021-11-22
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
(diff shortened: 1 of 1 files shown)


adri326:
[Rendered (latest)](https://github.com/sgebbie/ponylang-rfcs/blob/unix-domain-socket-support/text/0000-unix-domain-sockets.md)

sgebbie:
Following the [Pony sync 2021-11-23](https://ponylang.zulipchat.com/#narrow/stream/190591-sync/topic/2021-11-23) the general view is that there is support for including Unix domains into the `net` package. However, the before moving forward with the RFC, it would be good to have some proof-of-concept code showing how this might be implemented.

That is, it might be possible to use more FFI and the existing ASIO rather than making changes to libponyrt. Or, for maintainability it might be better to follow the existing code style.

PR #193, opened 2022-01-09
Make Array iterator classes private to hide implementation from client

pmetras:
Proposal to make `Array` iterator classes `ArrayKeys`, `ArrayValues` and `ArrayPairs` private. Instead `Array` function `keys`, `values` and `pairs` will return general `Iterator` interface like other stdlib classes that implement theses functions. Hiding the implementation of these iterators follow good design principles.

A new interface `Rewindable` has been added for when an iterator can be rewinded.

These changes have been implemented in my repository of the Pony compiler and all tests pass.

**EDIT**: Rendered RFC: https://github.com/pmetras/rfcs/blob/rfc-array/text/0000-array-private-classes.md


SeanTAllen:
If this is discussed at a sync that I'm not present at before I have a chance to comment more fully, I would like it noted that I am not in favor of this change. 

I want to note this note as I probably won't have time to put together a full write up within the next few days.

text/0000-array-private-classes
```diff
@@ -0,0 +1,105 @@
...
+`Array` functions `keys`, `values` and `pairs` are changed to return `Iterator` and the classes that implement these iterators are make private. Here are the full implementation of these functions.
+
+```pony
+  fun keys(): Iterator[USize]^ =>
+    """
+    Return an iterator over the indices in the array.
+    """
+    _ArrayKeys[A, this->Array[A]](this)
+
+  fun values(): (Iterator[this->A] & Rewindable[this->A]) =>
```
jemc: Rather than seeing `Rewindable` as a separate interface, I'd prefer to collapse the two together into an interface that is a subtype of `Iterator`, called `IteratorRewindable`:
```suggestion
  fun values(): IteratorRewindable[this->A] =>
```

My reasons are:
1. Most (all?) known use cases are for an iterator that is rewindable, not some non-iterator object that has a `rewind` method that returns an iterator
2. I think it's important to be really careful with the type names we put in `builtin`, because it means that no one else can ever declare a type with that name. I think `IteratorRewindable` is less likely to collide as compared to the other possible uses one can imagine for the name `Rewindable`.

In fact, one of the reasons I favor this change is because it would reduce the number of public type names defined in `builtin`, and thus reduce the collision surface for type names with those that users may want to define in their programs. So I personally see `IteratorRewindable` as being a better option for keeping that collision surface minimized, while also keeping it convenient to define an iterator that is rewindable (the suggested change reduces the total amount of characters to type/read, while keeping the intent very clear):

jemc:
Apart from the naming suggestion I gave above, I am in favor of this change :+1:

text/0000-array-private-classes
pmetras: In fact, the reason why I did not create a `RewindableIterator` interface but used `Rewindable` instead, is that this interface is not to dedicated to iterators. The interface defines a property that classes that implement `Rewindable` must implement a function named `rewind`. The semantic of that property is limited to the `rewind` name; it does not define a contract that `rewind` must implement, for instance that `has_next` must be true if it was true in a previous iteration and that `next` must give again the first element of the iterator.

Because of that, I thought that even if the risk of naming collision exists (one developer could already have a `Rewindable` type; the name is not used in Pony code on Github), the risk of semantic collision was null as `Rewindable` is an interface. But thinking while typing this, I now understand that the compiler will complain if there are two types with the same name but different definitions...

A `RewindableIterator` interface overlaps `Iterator` definition, something like:
```pony
interface RewindableIterator[A]
  fun has_next(): Bool
  fun next(): A ?
  fun ref rewind()
```

As Pony is using composition instead of inheritance to define new types, perhaps does it limit the ability to build new types with this interface? While one can add
```pony
type RewindableIterator[A] is (Iterator[A] & Rewindable[A])
```
and get the same result with more flexibility...
**EDIT: corrected the syntax of the type alias**

On the other hand, while Pony's stdlib is still evolving, defining shorter builtin types and names while Pony's ecosystem is still developing is the right time to take such decisions because they can't be changed later.

In the end, as we both agree on naming of builtins is very important for future code, I can change `Rewindable` to whatever name will be decided. This only concern the type returned by the `values` function, as `keys` and `pairs` could also probably return such iterator type.

pmetras:
A bit of context for this RFC.

I'm presently learning Pony, reading all documentation and particularly the one of the stdlib. Looking at the `Array` class, I was surprised that the API was unveiling array iterators implementations, while traditional API design promotes encapsulation. Looking at other classes in the stdlib, I saw that there were other classes that make public their internal implementation while others did not.

```
 $ find . -type f -name '*.pony' -exec grep "fun values" {} \; -print
  fun values(): Iterator[this->ByteSeq box]
./packages/builtin/std_stream.pony
  fun values(): (Iterator[this->A] & Rewindable[this->A]) =>
./packages/builtin/array.pony
  fun values(): Iterator[this->A]^
./packages/builtin/read_seq.pony
  fun values(): StringBytes^ =>
./packages/builtin/string.pony
  fun values(): Iterator[this->A]^
./packages/builtin/seq.pony
  fun values(): ListValues[A, this->ListNode[A]]^ =>
./packages/collections/list.pony
  fun values(): MapValues[K, V, H, this->HashMap[K, V, H]]^ =>
./packages/collections/map.pony
  fun values(): Iterator[val->A]^ =>
  fun values(): Iterator[val->A]^ =>
./packages/collections/persistent/list.pony
  fun values(): Iterator[A]^ =>
./packages/collections/persistent/set.pony
  fun values(): SetValues[A, H, this->HashSet[A, H]]^ =>
./packages/collections/set.pony
  fun values(): Iterator[this->A]^ =>
./packages/collections/heap.pony
  fun values(): Iterator[this->String]^ => strings.values()
./packages/cli/command_spec.pony
```

I couldn't find the logic but the fact that some results concern interfaces and some concrete classes. When designing classes, exposing internals is often considered bad design. But it could be related to some Pony *constraints* that I don't know about yet. So I decided to [ask the question on Zulip regarding the class `Array`](https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/Make.20Array.20iterators.20private). Sean T Allen listed his arguments but I did not understood them as they were neither in favour or against such API design decision. And he decided that he wouldn't answer my questions on the subject but if I open an RFC.

A few days passed and I was still thinking about that API decision, and because I think that there must be some rational in any decision, I thought that it could be because of a technical constraint imposed by the language. So I decided to test it by myself to see if there was something in Pony that prevented it. I did the changes in the `Array` class in the compiler and ran all the tests, and they passed without problem. So that was probably about something else I don't understand...

Here is the link to my branch of Pony compiler that implement the changes: https://github.com/pmetras/ponyc/tree/rfc-array

Following Sean T Allen's recommendation that the discussion should continue with a RFC, that's what I did.

One of Sean's argument is that naming things is important: "*I personally find naming the thing that implements the interface to be a good thing. I have a thing I can refer to. A name that can be reused.*"

Thinking about this argument, from a type or object perspectives is not important, in my opinion, when client code call `values` or `keys` functions. The client only needs to know that she is getting an opaque object of an unknown type to her (that's the private class), that respect the `Iterator*` return type of the functions...
I thought that this could be different in the situation of using composition, instead of inheritance, to extend this opaque object. But first, that's the `Array` data structure that a client generally wants to extend and not `ArrayValues`, and secondly composition extends through redefinition of functions in the new types. So hidden classes still remain hidden. And like I've shown in the RFC example of code using `ArrayValues`, client have no reasons to use these types.

All my attempts to find a rational failed. I must say I don't understand Sean's argument. I agree that naming thing for public types are important, but I don't see why you need to name things that clients of your class don't need to know (private implementation classes). That's against information hiding/encapsulation. I don't think that Pony syntax or stdlib promotes not enforcing this design principle.

pmetras:
And I forgot to say that `Array` is only one situation where this occurs. If one would want to *simplify* the stdlib API, this change could be applied in other places like `List`, `Map`, `Set`, etc.

text/0000-array-private-classes
jemc: > In fact, the reason why I did not create a `RewindableIterator` interface but used `Rewindable` instead, is that this interface is not to dedicated to iterators. The interface defines a property that classes that implement `Rewindable` must implement a function named `rewind`.

Well, your interface definition (as given in your document here) is not only that it must implement a function named `rewind`, but also that this function must return an `Iterator[A]`. Hence, your interface is only relevant to the use case of iterators and not to any arbitrary use case that calls for rewinding.

That's why I think it needs a more specific name - it is specifically about rewinding to produce an iterator - not about a fully abstract concept of rewinding anything.

jemc: Remember that while there exist workarounds for preventing namespace collisions in libraries you import, the `builtin` library (the one where `Array` is defined) is special in that it is always present and its implicit import cannot be modified, limited, or otherwise controlled. This is why I place such emphasis on taking care not to pollute the `builtin` library with public type names that could collide with useful types that the user may want to define differently with that same name - `builtin` is the only library that will make defining such a type in user code literally impossible.

pmetras: Yes, you're right. I focused on the function name and I forgot about the return type.
Should the `RewindableIterator` interface define the 3 functions `has_next`, `next` and `rewind`, or only the `rewind` one to be used by composition? I think the former makes the code simpler by defining a new type of iterator.

jemc: Yes, I'd prefer the former - with a name like `RewindableIterator`, I'd expect it to be a subtype of `Iterator`, so I believe it should have all the same methods as `Iterator`, in addition to the `rewind` method.

ergl:
I don't have a strong opinion on this RFC, but if `Array` already returned an `Iterator` interface, I'd probably vote against an RFC that wanted to change them to return concrete types as it does now.

My only concern about this RFC is adding the overhead of virtual calls for these iterators. I assume iterating over arrays is very common, so these specific classes are used a lot, even if users don't name them.

jemc:
@ergl - I had the same concern about virtual calls, though I think in practice most iterator-using use cases are direct enough (e.g. passing straight into a `for` loop) that LLVM "should" be able to inline enough of the code to devirtualize the calls by seeing the concrete type descriptor being written in the same context where the otherwise-virtual calls are being made.

This is a hypothesis we should probably test prior to accepting the RFC. I can do some testing of that if it looks like the RFC has enough support to be accepted.

jemc:
@SeanTAllen mentioned in the sync call that he'd like to see one of two things happen with this RFC:

- :one: Update the RFC to apply to state that it will apply to all collection iterators in the standard library
- :two: Update the reasoning provided in the RFC, to explain why it will apply only to `builtin` and not set a precedent that should apply to other packages in the standard library

He wants to hone in on one of those two directions so that he can better focus his energy in stating his objections, either to the general case (:one:) or to the `builtin`-specific reasoning (:two:)

I think this a reasonable request, and agree.

pmetras:
My initial concern was only to get an answer to the question in the title of this RFP that I could understand. It could have been *"It's that way because I decided that it is"*, without any later discussion... But I was intrigued by the fact that the design decision is against the Sanctus encapsulation/information hiding principle of computer science. And I decided to open this RFC to get a definite answer.

Now, if I had to select between direction 1️⃣ or 2️⃣, I think that applying a design principle only on part of a library would be a bad decision by creating an exception (the less exceptions, the better the language). The remaining paths are either correct everything ( 2️⃣) or do nothing (cancel and close the RFC).

Personally, I think that 2️⃣ is the best move, but I don't know yet what could be the impact on existing code or other consequences of such change. But I'm ready to investigate the impacts and update the text of the RFC in that direction if others think it's worth my effort.

jemc:
@pmetras - I'm confused by your use of numbering - can you please check the numbers in the above comment I gave to make sure they match the use of your numbers - I think you meant :one: when you said :two: - can you confirm?

We discussed this on today's sync call, and Sean suggested you go ahead and try to update the RFC to make your arguments in favor of the approach for either :one: or :two: - that's not a guarantee the RFC will be accepted but it's to ensure he (and others) fully understand the points being argued so that he (and others) can voice any relevant objections clearly.

pmetras:
> I'm confused by your use of numbering - can you please check the numbers in the above comment I gave to make sure they match the use of your numbers - I think you meant  when you said  - can you confirm?

Effectively, I mixed the numbers. I meant [one], that is applying the design change to all stdlib.

SeanTAllen:
This is very hard to read in GitHub due to lack of line breaks. Can you add hard line breaks around column 80 or so to make it easier to read in the GitHub web UI?

SeanTAllen:
I think this RFC doesn't consider the cost of breaking user code enough. Even if I was to be convinced of the value of this change, I think we need to give more weight to considering that this will break people's code and a good justification beyond "design principles" needs to be given for breaking code.

Regardless of how I feel about this specific RFC, I would make the same point for any RFC. I would like to see the breaking change taking more seriously and consideration given to justifying that breaking change. 

RFCs set precedent for future changes and the breaking change in here feels somewhat cavalierly presented.

text/0000-array-private-classes
```diff
@@ -0,0 +1,148 @@
...
+    """
+    _ArrayPairs[A, this->Array[A]](this)
+```
+
+As the function `values` of class `Array` uses an iterator with a `rewind` function that is not part of the `Iterator` interface, a new interface `RewindableIterator` is added to enable creation of rewindable iterators.
+
+Note: To remain consistent with `Array` behaviour, functions `keys` and `pairs` should return a `RewindableIterator` too but we limited the API change to minimum as we did not understood why it was not already the case.
+
+```pony
+interface RewindableIterator[A] is Iterator[A]
```
SeanTAllen: There's no need for `is Iterator[A]` as far as I can tell here. Is this an attempt to make sure in the future that RewindableIterator doesn't diverge at all from Iterator?

text/0000-array-private-classes
```diff
@@ -0,0 +1,148 @@
...
+- Collections' functions `keys`, `values` and `pairs` definitions are made more general. Iterators implementation details are not public. Internal classes used by implementation like `*Keys`, `*Values` and `*Pairs` are now [opaque data types](https://en.wikipedia.org/wiki/Opaque_data_type). Generally, when using these collection classes, clients are not interested by the iterators implementation, but by the types these iterators return and that is provided by the generic parameters.
+- The generic return signature of these 3 iterating functions is simpler to understand for clients of collection classes.
+- Makes the standard library more simple by hiding 18 specialised classes from stdlib of which 3 are from `builtin`.
+- The interface `RewindableIterator` is added to create rewindable iterators (can be re-start from first value).
+
+This change remains compatible with the existing code base but for client code that is directly using the classes `*Keys`, `*Values` and `*Pairs`. A search on Github shows that the impact is very limited.
+
+# Detailed design
```
SeanTAllen: I'd like to see a full enumeration of all the existing method signatures that would change and what they would change to.

text/0000-array-private-classes
```diff
@@ -0,0 +1,148 @@
...
+- Collections' functions `keys`, `values` and `pairs` definitions are made more general. Iterators implementation details are not public. Internal classes used by implementation like `*Keys`, `*Values` and `*Pairs` are now [opaque data types](https://en.wikipedia.org/wiki/Opaque_data_type). Generally, when using these collection classes, clients are not interested by the iterators implementation, but by the types these iterators return and that is provided by the generic parameters.
+- The generic return signature of these 3 iterating functions is simpler to understand for clients of collection classes.
+- Makes the standard library more simple by hiding 18 specialised classes from stdlib of which 3 are from `builtin`.
+- The interface `RewindableIterator` is added to create rewindable iterators (can be re-start from first value).
+
+This change remains compatible with the existing code base but for client code that is directly using the classes `*Keys`, `*Values` and `*Pairs`. A search on Github shows that the impact is very limited.
```
SeanTAllen: There seems to be a thought missing from:

`
This change remains compatible with the existing code base but for client code that is directly using the classes `*Keys`, `*Values` and `*Pairs`.
`

What did you mean to say here? That this is a breaking change for anyone who names any of the concrete implementations in their code?

If yes, that isn't really a motivation. That would be a drawback. I don't believe you intend to say that breaking people's existing code is a motivation for this change.

SeanTAllen:
@pmetras thank you for the linebreaks. its much easier to read now.

text/0000-array-private-classes
pmetras: That's true for `interface`but I just kept with the coding style of the stdlib where numerous classes (i.e. `Array`, `String`, `List`, `Range`, `Reverse`, `HashSet`, etc.)  explicitly name the interface they implement.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+functions discussed in this RFC, iteration).
+- Collections' functions `keys`, `values` and `pairs` definitions are made more
+general. Iterators implementation details are not public. Internal classes used
+by implementation like `*Keys`, `*Values` and `*Pairs` are now
+[opaque data types](https://en.wikipedia.org/wiki/Opaque_data_type). Generally,
+when using these collection classes, clients are not interested by the iterators
+implementation, but by the types these iterators return and that is provided by
+the generic parameters.
+- The generic return signature of these 3 iterating functions is simpler to
+understand for clients of collection classes.
```
SeanTAllen: Can you provide justification for this? I think this needs a supporting argument.

text/0000-array-private-classes
pmetras: You can see them on [my branch](https://github.com/pmetras/ponyc/tree/rfc-array).

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+- Makes the standard library more simple by hiding 18 specialised classes from
+stdlib of which 3 are from `builtin`.
```
SeanTAllen: What's the definition of simpler here? The end result is the same number of total classes plus 1 new interface. Is simpler, "fewer non-private classes"?  The idea here seems to be that fewer means easier to use, which I don't think is a good argument. I doubt that is what you mean. Can you expand on the reasoning here?

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+- The interface `RewindableIterator` is added to create rewindable iterators
+(can be re-start from first value).
+
+This change remains compatible with the existing code base but for client code
+that is directly using the classes `*Keys`, `*Values` and `*Pairs`. A search on
+Github shows that the impact is very limited.
```
SeanTAllen: This doesn't seem to be motivation and should be worked into design as these are notes and statements about a some specific design points.

text/0000-array-private-classes
pmetras: I don't know how to translate **but for client** when you want to say that only that minority of users will be impacted by this change...

Show me actual code that will break.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+    """
+    Return an iterator over the values in the array.
+    """
+    _ArrayValues[A, this->Array[A]](this)
+
+  fun pairs(): Iterator[(USize, this->A)]^ =>
+    """
+    Return an iterator over the (index, value) pairs in the array.
+    """
+    _ArrayPairs[A, this->Array[A]](this)
```
SeanTAllen: It strikes me as very weird that keys and pairs aren't currently rewindable. It seems like they should be rewindable. I think that rather than making this RFC a straight translation of what currently exists, that detailing changed like "making ArrayKeys and ArrayPairs" rewindable would be a good addition. 

The lack of rewindability of the existing keys and pairs (that probably happens elsewhere as well) seems more like an accident of history than a design choice.

Looking for oddities like this is why I'd like an enumeration of all the changes.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+```
+
+As the function `values` of class `Array` uses an iterator with a `rewind` function
+that is not part of the `Iterator` interface, a new interface `RewindableIterator`
+is added to enable creation of rewindable iterators.
```
SeanTAllen: RewindableIterator is mentioned in the code before this, I think you should move this text before RewindableIterator is introduced in the code.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+
+Note: To remain consistent with `Array` behaviour, functions `keys` and `pairs`
+should return a `RewindableIterator` too but we limited the API change to minimum
+as we did not understood why it was not already the case.
```
SeanTAllen: As noted above, I think you should drop this point.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+    """
+    Return the next item of the iteration or an error in case there are no other
+    items. A previous call to `has_next` check if we can continue iteration.
+    """
+
+  fun ref rewind(): Iterator[A]^
+    """
+    Get a new iterator that can be used to start the iteration again from the
+    first item.
+    """
```
SeanTAllen: The existing rewind on ArrayKeys doesn't create a new iterator. It resets where the current iterator is to the beginning. Is this an intentional change? If yes, it should be noted and documented as to why. This change could possibly break existing code.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+```
+
+The code of the standard library is adapted to remove use of these now private
+classes, mainly in tests. Here are the files that must be changed:
```
SeanTAllen: An example of where they are used in tests and why the concrete type in tests doesn't matter would be good.

text/0000-array-private-classes
pmetras: In `Array`:
```pony
-  fun keys(): ArrayKeys[A, this->Array[A]]^ =>
+  fun keys(): Iterator[USize]^ =>
```
function signature seems to be simpler to me for the client user that knows now that she gets an iterator over `USize`.

Idem for
```pony
-  fun values(): ArrayValues[A, this->Array[A]]^ =>
+  fun values(): RewindableIterator[this->A]^ =>

-  fun pairs(): ArrayPairs[A, this->Array[A]]^ =>
+  fun pairs(): Iterator[(USize, this->A)]^ =>
```

Let's take something more cryptic like `Map`:
```pony
-  fun keys(): MapKeys[K, V, H, this->HashMap[K, V, H]]^ =>
+  fun keys(): Iterator[this->K]^ =>
-  fun values(): MapValues[K, V, H, this->HashMap[K, V, H]]^ =>
+  fun values(): Iterator[this->V]^ =>
```
We see that we iterate over keys (`K` type) or values (`V`) and not over a complex thing...

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+* `packages/collections/heap.pony` in function `values`
+* `packages/collection/builtin/_test.pony` in class `_TestArrayValuesRewind`
+* `packages/collections/list.pony`
+* `packages/collections/map.pony`
+* `packages/collections/persistent/map.pony`
+* `packages/collections/persistent/vec.pony`
+* `packages/collections/set.pony`
+* `test/libponyc/util.cc` to change the name of the class to `_ArrayValues`
+
+# How We Teach This
```
SeanTAllen: Nothing in this section seems to be about teaching people anything going forward about the pattern nor explaining to people impacted by the breaking change what they need to be aware of. This section should be devoted to what if anything should be done with Pony Patterns, the Pony tutorial, standard library documentation, and release notes to adapt to the new world this RFC would bring about.

text/0000-array-private-classes
pmetras: *Simpler* means less public classes (lower number), when these classes are not used in real code. Just show collections client code on Github where using instances of the `*Keys`, `*Values` or `*Pairs` classes is justified. We'll then compare with the number of calls of the `keys`, `values` and `pairs` functions that are simpler when to specifying these now private types.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+but that the `values` function must return an `RewindableIterator` over `Xml2node`.
+The new version makes the code simpler to understand.
+
+This change in `array.pony` and other collections will break such code but it
+can be easily adapted to use the new API. And it will make the standard library
+easier to learn by reducing the number of public types.
+
+# How We Test This
+
+Pony tests must continue to pass.
```
SeanTAllen: I assume this means that existing standard library tests are updated and that no new tests are to be added? Are there non standard library tests that this impacts on as well?

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+
+# Drawbacks
+
+As said, some client's code must me adapted when using these classes. As these
+classes are just concreted implementations of `Iterator` by collection classes,
+their use in client code is limited and the code can very easily be changed.
```
SeanTAllen: I don't believe this is entirely correct. I'd rather see this as

- Will break any existing code that uses any of the classes that are currently public and will be made private by this RFC.

It states clearly what the expected impact would be and leaves editorializing out.

text/0000-array-private-classes
pmetras: Documentation comments are mine as they are absent from original source.

text/0000-array-private-classes
```diff
@@ -0,0 +1,198 @@
...
+
+# Alternatives
+
+Stay as is. Continue the
+[discussion on Zulip](https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/Make.20Array.20iterators.20private).
```
SeanTAllen: If you update the RFC to also include normalizing `rewind` usage, then I think an alternative would be to keep the concrete classes but augment where it makes sense to include `rewind`. If you do include the `rewind` additions, then I think you should add the alternative here.

I believe that you also discussed previously a different way of implementing the RewindableIterface- I think it is worth adding that as an alternative here.

text/0000-array-private-classes
pmetras: The goal of this RFC is to follow good API design and to remain compatible as far as possible. No new tests were required and the existing ones still pass.

I don't have access to a larger test base than the stdlib to test for compatibility.

text/0000-array-private-classes
SeanTAllen: So in the case of those, it made sure that if the Iterator interface was changed that all those concrete classes would fail compilation, is that the intention here? If yes, I think there is a larger discussion about the idea of duplicating the methods from Iterator in RewindableIterator and then using `is` to enforce the Iterator interface. 

It feels odd. It might be worth it to avoid having an interface that is only `rewind` and the returning the two interfaces, but, it does feel odd and I think the reasoning is worth discussion.

I know you had some conversation around this with @jemc. I think you should discuss with him some supporting justification for this. Is this a one off? Is this establishing a pattern that we want to use in the standard library? Is it a pattern that is already being used in the standard library?

text/0000-array-private-classes
SeanTAllen: Please include them in the RFC.

text/0000-array-private-classes
SeanTAllen: I think this there's nothing here that is a motivation and you should either cut the text in question or move it summary.

Having a summary that gives a quick overview and says that it is a breaking change but you expect from a GitHub search that it should impact only a small amount of code would be reasonable.

text/0000-array-private-classes
SeanTAllen: So simpler to understand is that the type signature is smaller? Am I correct that is your meaning? If yes, you should replace "simpler" with "smaller type signature" and something about how that should be easier to understand.

text/0000-array-private-classes
SeanTAllen: Rather than simpler please change to "less public classes".

text/0000-array-private-classes
SeanTAllen: I don't think you understand my statement. Your documentation changes the behavior of rewind. If that is intentional you need to include in the design. If that's unintentional, the text should be something like "Start the iterator over again from the beginning".

text/0000-array-private-classes
SeanTAllen: Its unclear what the comment in the RFC means, can you be more explicit. Something like:

"No additional tests are need as after review the existing coverage in Pony standard library tests is sufficient"

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: Done

text/0000-array-private-classes
pmetras: > I think an alternative would be to keep the concrete classes

You didn't show an example where having these classes is justified outside the collections code.

This alternative reminds me of Sun, ten years ago before being bought by Oracle, keeping some buggy classes in the Java stdlib because some of its customers were using them and having built software upon non-documented (understand buggy) behaviour. But Sun had tens of millions of Java users, many of them paying for a Java license or support, and Java has a [deprecation procedure](https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html) to allow API transition between language versions (since version 1.1). None of these conditions applies to Pony yet... Is it worth adding an alternative not supported by concrete proofs?

SeanTAllen: I'm not asking for a change that will break some code. You are. I don't need to justify anything. I presenting you with items that will make your RFC that I am not at the moment in favor of stronger. You do not have to make your RFC stronger, that is up to you. Despite my not being in favor of this RFC, I've been spending my free time helping you make it more convincing, I don't have to do that.

text/0000-array-private-classes
jemc: I agree that those standard library iterators that *can* be made efficiently rewindable should all be made such.

jemc: And I think the fact that you were able to notice that so quickly in the code actually speaks to one of the strengths of this approach - when I see the type name `ArrayValues` or `ArrayKeys` in the return signature I don't immediately know what features it exposes (I wouldn't even necessarily know it was an iterator at all just by the name). Whereas with the approach advocated by this RFC, as long as I know what `Iterator` and `RewindableIterator` are, I can immediately understand what it is and what features it exposes (as long as I already know about `Iterator` and/or `RewindableIterator`).

And for us, making this change made it easy to notice features that were missing (rewinding in some iterators).

text/0000-array-private-classes
jemc: For the record, I 100% agree that it's simpler to understand. Let me know if you need help coming up with ideas on how to express that in the RFC.

text/0000-array-private-classes
jemc: My suggestion would be to keep the `is Iterator[A]` part and remove the explicit `next` and `has_next` definitions.

For an `interface`, that roughly reads as:
- "To implement `RewindableIterator[A]`, you need to implement everything `Iterator[A]` requires (which doesn't need to be repeated here), but you additionally need to implement this `rewind` method."

Or from the perspective of the interface user:
- "If you get a `RewindableIterator[A]`, it can do everything `Iterator[A]` can do (which doesn't need to be repeated here), but you additionally can use it to `rewind`"

text/0000-array-private-classes
jemc: I agree that those standard library iterators that *can* be made efficiently rewindable should all be made such.

To me, that's the best definition of "consistent" rather than keeping consistent with the current (artificially limited) behavior.

text/0000-array-private-classes
jemc: @pmetras - "How We Teach This" could include a post on the "Pony Patterns" site that demonstrates how to use a public interface to hide the implementation of a private class, rather than making that class public.

Like you, I see this as a desirable design pattern, so we could have a "Pony Patterns" example showing how we did this in the Pony standard library with `Iterator`s (assuming we can get the RFC approved).

text/0000-array-private-classes
```diff
@@ -0,0 +1,285 @@
...
+structure that would implement a rewindable property. This alternative was
+[put aside](https://github.com/ponylang/rfcs/pull/193#discussion_r780793165) to
+prevent name colisions in `builtin` with user-named types.
+
+# Unresolved questions
+
+Pony language definition is not precise about
+[type variance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science))
+and its impact in collections classes, particularly for covariant return types
+of functions (in this RFC, `keys`, `values` and `pairs`).
```
jemc: Standard notions of covariance and contravariance do apply as normal in Pony, including covariant return types.

Do you still have open/unresolved questions on this topic? If so, I can probably help resolve them. Feel free to expand your questions here or hit me up in the Zulip chat to discuss further.

Once you're comfortable with the answers we can remove this from the "unresolved questions" section, since I don't think there are any Pony surprises related to this particular topic that are relevant to this RFC.

jemc:
@SeanTAllen asked me to weigh in on this RFC to talk a bit about my reasoning, since he knows I support this RFC.

I believe that *minimizing* API surface is an important component in an overall strategy of *stabilizing* an API surface.

To be clear about my terms, and to further explain:

When I talk about "stabilizing an API surface", I mean the process of carefully designing an API surface in such a way that the need for future breaking changes is prevented to the greatest extent that is reasonable. That design process includes being clear and explicit about which API guarantees you are making to the API consumers.

It also includes carefully considering which API guarantees are actually *helpful* to the API consumers, so that guarantees which are not helpful to API consumers can be excluded. And that's what I mean when I talk about "minimizing an API surface". An ideal stable API will minimize the set of guarantees it makes to include only the most *helpful* set (above some particular threshold of "helpfulness"), because every guarantee made by a stable API is a fetter that may weigh you down during the process of introducing future improvements.

To quote Antoine de Saint-Exupéry:
> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.

And that's the simplest reason why I am in favor of this RFC. Currently the API makes a guarantee that the return value of each of these functions is a specific concrete class with a specific public name. That particular guarantee does not actually serve any practical needs of users (except, arguably, the virtual calls caveat [that was mentioned above](https://github.com/ponylang/rfcs/pull/193#issuecomment-1010186664)). If it is not helpful to users in practice, it can be removed. And if it *can* be removed, then it *should* be removed, in the name of taking another step toward stabilizing the Pony standard library's API.

That's my basic argument.

It's worth mentioning that if this change was being debated in reverse (the current standard library had minimal interfaces for the return values, and somebody proposed to make them public concrete classes instead, because they found some tangible benefit in doing so), then it would not be a breaking change at all, because the new proposed type would be a subtype of the existing return types.

It'll always be easier to get more specific with these return types than more general. So we should default to offering the most general return type that is still useful to users, and ratchet down to more specific subtypes as needed (as tangible use cases for more specific return types are discovered). Ratcheting down will always be possible without a breaking change, but ratcheting up will be a breaking change each time. It's best to make such "ratchet up" breaking changes like this RFC before we reach Pony 1.0.0 as part of a general effort to make Pony and its standard library stable.

SeanTAllen:
@jemc and I discussed this at length on Friday and came up with a version of this RFC that we can both agree upon. One of us will be adding information about what agreed upon as that we could both support.

SeanTAllen:
Things Joe and I discussed and agreed upon (@jemc please add any additional points I missed adding):

- We discussed why we consider this a change to ONLY collections iterators and why it wouldn't set a precedent outside of the case we considered for collections. The reason for this is that looking at the existing collections and collections we forsee coming, we believe that the iterators for all of the collections should always remain in lockstep for the interface they provide. For example, `rewind` doesn't exist on all the current iterators for collections but really, we think it should. A shared interface being returned as with this RFC would assure they stay in sync as intended.

- Further, we believe that all collection iterators should have `rewind as a method as part of this RFC.

- We also believe from the above that the collections iterator shouldn't be called "RewindableIterator" or use "rewind" in the name as we would add new collection iterator methods to said interface in the future and then it would have a poor name. That is, the interface in question is an interface that describes a collections iterator.

- Based on how structural typing works in Pony, we do not need more general interfaces like `rewind` or `RewindableIterator` that exist in the standard library as users can add them on their own and in fact should. If a user only needs the next() and has_next() methods from an interface that includes next(), has_next(), and rewind(), they should create their own interface that is limited to those two methods IF they want to be protected from any breaking change that might come from a change in the standard library.

- We doubt that this change will break much user code, however, we don't really have much evidence either for or against that position.

- We doubt that this change will require a lot of work for someone to update to fix.

- We are most worried about breakage in terms this breaking 3rd party libraries that might still work prior to this change but are otherwise not being maintained. Given our dearth of 3rd party libraries, we consider this an important point, but again, not particularly likely.

- We don't think that this change will result a large immediate win for user.

- We do think that this will make the standard library easier to maintain and will make it easier to present users with a consistent interface to collections iterators.

Given all of the above points, we are in favor of this RFC if it was to be rewritten to take all of the above points and justifications into account thereby codifying for future decision making, why this RFC was accepted and what impact if any it has on precedence for making future decisions.

SeanTAllen:
The only thought I've had so far for the name of the iterator that collections iterators would conform to is `CollectionIterator`.

jemc:
I have not yet thought of a better name than that.

pmetras:
Sorry for the delay in responding, I'm quite busy these days. I'll try to complete the open points before next week-end.

I've seen that you want to limit the changes to the `collections` module, and that's the place where the `Iterator` interface is used. And the code changes are effectively limited to that module.

But I'm questioning about the use of interfaces in Pony when compared with other languages (where I'm more accustomed to call them *traits*). For me, a Pony interface is used to present a *type view* (I don't have a better term) of an object to a function that expects that this objects implements some functions. The receiving function can be assured, from that parameter object that implements the *type view*, that it'll be able to call the functions defined by the interface. That's a contract between the receiving function and the parameter. The receiving function is not concerned by other *type views* (other Pony interfaces) that the parameter would implement, nor is it concerned if the return types from the interface functions are **subtypes** of the types as defined in the interface. These are minimal requirements for the receiving function when using interfaces.

So I was surprised when I saw that Pony developers wanted to make the return types explicit, as they were subtypes of the minimum types, as this was particularly flagrant with the `Array` iterators functions. For this reason, I want to audit the whole stdlib source and check if this development pattern is used elsewhere with other interfaces. This is not really part of this RFC and won't be included, but for trying to understand what could justify it. Why would one need to unveil implementation classes, as it it still a question without explanations in my mind?

pmetras:
A way to solve the `CollectionIterator` vs `RewindableIterator` naming issue is to reconsider what I suggested in a previous discussion with @jemc but that we put aside: using functional traits composition and create `interface Rewindable`. But at that time I did an error in the way I wrote it, as it was dependant on `Iterator`.

Instead, write:
```pony
interface Iterator[A]
  """Something we can iterate on"""
  fun ref has_next(): Bool
  fun ref next(): A ?

interface Rewindable[A]
  """Something that can be rewound"""
  fun ref rewind(): A
```

Now you can create rewindable iterators for `Array` and non-rewindable ones for other collections with something like the following untested code:
```pony
type RewindableIterator[A] is (Iterator[A] & Rewindable[Iterator[A]])

class _ArrayValues[A: A, B: Array[A] #read] is RewindableIterator[B->A] ref
```

That way, one can really build functional interfaces/traits from composition of basic interfaces/traits.

SeanTAllen:
We dont want to introduce a number of not needed interfaces into builtin @pmetras. We want a single interface for collectins iterators because we expect them to change in lockstep.

pmetras:
Yes, but you can't have a single interface to describe two different functional behaviours: rewindable vs non-rewindable iterators.

The way I explained is the way to limit the number of specialized interfaces. Imagine that later there's a new kind of iterator where you can ask for the number of remaining elements in the iteration. The way to solve it is to create a `RemainingCountable` interface an use composition to match with the different types, not creating all the variations `RemainingCountableIterator`, `RemainingCountableRewindableIterator`, `RewindableIterator`, etc.

SeanTAllen:
We want 1 interface for all collections because we want them to move in lockstep. All collections iterators should be rewindable.

SeanTAllen:
When discussed during the sync call on Tuesday, consensus amongst attendees was that agreement with the points and suggestions put forth in my summation comment of the ["Sean and Joe believe comment"](https://github.com/ponylang/rfcs/pull/193#issuecomment-1021509909).

Attendees were @sgebbie, @kulibali, @redvers, @ergl, @jemc, and @SeanTAllen. I don't believe all attendees voiced an opinion but those who did supported the plan as laid out in the [aforementioned comment](https://github.com/ponylang/rfcs/pull/193#issuecomment-1021509909).

Audio of the call for context is available [here](https://sync-recordings.ponylang.io/r/2022_02_01.m4a).

pmetras:
Sorry, I've not much time to work on it in the coming days. I'm not sure I'll be able to complete it. Though here are some thoughts I had while browsing the stdlib source. I've numbered them for easier reference, but the sort order is not important.

1. First, here is how I understand the **change in lockstep** thing as the given explanations are quite terse. In a `collection` data structure, you want that all iterator functions that exist (usually `keys`, `values` or `pairs` but some collections have more of them or less) return a `CollectionIterator` and not an `Iterator`. If a new function is added later in `CollectionIterator`, for instance `infinite` that returns if the iterator is bound or infinite, the goal is for all the collections to benefit from this addition.
Like [I wrote](https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/Add.20rewind.20to.20iterator.20interface), changing the contract of an interface by adding new function will break client code (i.e. all clients using `CollectionIterator`). So usually this is something that is not done nor desired by stdlib authors. I don't understand what justifies this *change in lockstep*.

2. What would be the rule for a developer to decide using `Iterator` vs `CollectionIterator`? Because she will see signatures for `values`, for instance, that are sometimes `CollectionIterator` (in `collections` or `builtin`) and other times `Iterator` (elsewhere in stdlib). It'll be more complex both for the API stdlib user but also the stdlib author to decide which type to use without a simple rule.

3. Though `Range` and `Reverse` are part of `collections`, are they `Iterator` or `CollectionIterator`? If they are `CollectionIterator`, their `rewind` function signature must be changed. What will be the impact on existing client code?

4. `Array` is not part of `collections`, but does it need to be considered as a collection? It means that `builtin` will also knows about `CollectionIterator`, and probably Pony compiler...

5. Do we need to consider the builtin interfaces `Seq` and `ReadSeq` as `Iterator` or `CollectionIterator`? If you consider that a sequence is a collection of items, like `Array`, how do you consider `StringBytes` or `StringRunes` because one would want to be able to rewind iteration on these types?
And what about `ByteSeqIter` as used in streams, where we can't rewind when the stream has been flushed?

6. Many iterators in collection are built with the following destructive pattern:
    * `create`: Create a data structure, frequently `Array` where all the collection items are put.
    * `has_next`: Test if the data structure is empty.
    * `next`: Remove an element from the data structure and return it

To include `rewind` in these iterators, as far as I've understood from a previous comment from Sean, you don't want to rebuild the data structure but use existing content, meaning that the content will have to stay in the data structure. It requires to rewrite the whole iterator code. What is the impact on data structure elements capabilities? Do we need to declare them `tag` or `val`?

Even if some of you have a definite opinion on the best solution to implement this RFC, I'm still thinking on the various options, balancing impact on existing and future code.

SeanTAllen:
> Like [I wrote](https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/Add.20rewind.20to.20iterator.20interface), changing the contract of an interface by adding new function will break client code (i.e. all clients using CollectionIterator). So usually this is something that is not done nor desired by stdlib authors. I don't understand what justifies this change in lockstep.

Adding a method to an interface will only break implementers and the intention is that the only implementers are in the standard library. Any user code that accepts a `CollectionIterator` will not break.

SeanTAllen:
> Array is not part of collections, but does it need to be considered as a collection? It means that builtin will also knows about CollectionIterator, and probably Pony compiler...

For the sake of this conversation when Joe and I say "all collections", that includes `Array`. A full enumeration of exactly what uses CollectionIterator and what doesn't should be included in the RFC. Anything that is unclear should be in unresolved to highlight that it would need to be addressed.

I might not have stated explicitly and I apologize for that `CollectionIterator` would be part of the builtin package.

SeanTAllen:
> Many iterators in collection are built with the following destructive pattern:
>
> create: Create a data structure, frequently Array where all the collection items are put.
> has_next: Test if the data structure is empty.
> next: Remove an element from the data structure and return it

Can you provide an example of this pattern? I didn't see it upon a quick review.

SeanTAllen:
> Do we need to consider the builtin interfaces Seq and ReadSeq as Iterator or CollectionIterator?

Seq and ReadSeq are interesting and limited cases as they only provide access to `values()` as it is the bare minimum interface. I would imagine they to be `Iterator`. They are interfaces not concrete collections like we would be applying `CollectionIterator` to. `ReadSeq` and `Seq` are used outside of the standard library and the intent of `CollectionIterator` is that it is "standard library only"

SeanTAllen:
> Though Range and Reverse are part of collections, are they Iterator or CollectionIterator? If they are CollectionIterator, their rewind function signature must be changed. What will be the impact on existing client code?

I believe those are CollectionIterator but that would require more thought.

The only change I see to the signature would be "should this return a `this` like ArrayValues does. Given that can do method chaining with ".>", it seems like `rewind` on ArrayValues is using the pre-method chaining approach to returning `this` which wouldn't be needed. I think the key here is to decide on the appropriate signature for rewind and implementing accordingly.

If some existing rewinds changed to have a return type, that isn't going to break code, but if we remove the return value (say from ArrayValues) then it would be a breaking change.

Looking at ArrayValues, that looks like we missed changing the return value when we introduced method chaining. I personally would be in favor of changing the ArrayValues return type and accepting the breakage from that given that we already have accepted breakage to make this change.

pmetras:
> 1. Adding a method to an interface will only break implementers and the intention is that the only implementers are in the standard library. Any user code that accepts a CollectionIterator will not break.

Any existing client code that presently accept a `*Key` or `*Value` or `*Pairs` iterator will have to be recompiled to accept an `CollectionIterator` instead, but this can be done usually without code change.

But if I implemented a `BitSet` data structure or other type of collection, that is not part of the stdlib, with a `values` function, I'll have to adapt my code if I want to offer a *collection*-compatible interface to clients, meaning that I'll have to implement the new `rewind` function.

text/0000-array-private-classes
rhagenson: I will also voice my support that these type signatures are easier to understand for me. I primarily see them as simpler due to the generic parameters not being repeated. For example, in `MapKeys` although it is not difficult to know that K, V, H in `MapKeys[K, V, H, this->HashMap[K, V, H]]^` are the same K, V, H as in `MapKeys[K, V, H: HashFunction[K] val, M: HashMap[K, V, H] #read]` having the return here shortened to `Iterator[this->K]^` avoid irrelevant repetition of V and H. I read this more clearly as "an iterator over the K values as viewed by this".

text/0000-array-private-classes
```diff
@@ -0,0 +1,285 @@
...
+classes allows changing the implementation. Usually, one must return the most
+general type that fullfils the contract of the function (in the case of the
+functions discussed in this RFC, iteration).
+- Collections' functions `keys`, `values` and `pairs` definitions are made more
+general. Iterators implementation details are not public. Internal classes used
+by implementation like `*Keys`, `*Values` and `*Pairs` are now
+[opaque data types](https://en.wikipedia.org/wiki/Opaque_data_type). Generally,
+when using these collection classes, clients are not interested by the iterators
+implementation, but by the types these iterators return and that is provided by
+the generic parameters.
```
rhagenson: Motivation can be used a precedence for future RFCs so I want this clarified to state that we are not setting a precedence to generally avoid returning concrete types in the stdlib -- which would be a mistake in my opinion. This reasoning to use opaque types due to client disinterest does not extend beyond this RFC -- specifically we are stating that we have those expectations of disinterest for iterator implementations.

text/0000-array-private-classes
rhagenson: As noted below by me, I prefer the alternative design of not having `RewindableIterator[A]` overlap with `Iterator[A]` and instead have `Iterator[A] & Rewindable[A]` define `RewindableIterator[A]`.

text/0000-array-private-classes
```diff
@@ -0,0 +1,285 @@
...
+  fun ref rewind(): Iterator[A]^
+    """
+    Start the iterator over again from the beginning.
+    """
+```
+
+Then one can create a rewindable iterator by composing these two interfaces:
+
+```pony
+type RewindableIterator[A] is (Iterator[A] & Rewindable[A])
```
rhagenson: As noted above, I prefer this approach. I think whenever possible we should have non-overlapping interfaces in the stdlib which are then joined into commonly used type alias such as this.

pmetras:
> 6. Can you provide an example of this pattern? I didn't see it upon a quick review.

Here are examples I've found tonight doing a search on `shift`:
* https://github.com/ponylang/ponyc/blob/8ebf87119cf958de07577c0c8b71e2ce63c5986e/packages/collections/persistent/vec.pony#L280
* https://github.com/ponylang/ponyc/blob/8ebf87119cf958de07577c0c8b71e2ce63c5986e/packages/collections/persistent/_map_node.pony#L73

text/0000-array-private-classes
```diff
@@ -0,0 +1,285 @@
...
+  fun values(): RewindableIterator[Xml2node]^ ? =>
+    if (allocated) then
+      nodearray.values()
+    else
+      error
+    end
+```
+
+In this sample, the developer was not really concerned by the type of the iterator
+but that the `values` function must return an `RewindableIterator` over `Xml2node`.
```
rhagenson: I do not think you can speak on behalf Red's concerns here; unless I am missing some prior conversation where Red stated this directly.

pmetras:
No one explained the advantages or disadvantages of creating a new type `CollectionIterator` instead of updating the existing interface `Iterator`. If `CollectionIterator` were private to the `collection` package, I could understand that its definition evolves like the classes it operates on and that changes to that type of iterator would have limited impact (I think that's what Sean T. Allen calls lockstep), but it does not seem to be what's desired. I don't see a major advantage in creating a new type of iterator for collections only, when both iterators are builtin. `Iterator` could be updated to define `rewind` with a default implementation, and neither client's nor implementer's code won't be impacted at all...

Please explain why adding `CollectionIterator` is a better solution than updating `Iterator` so these arguments can be added to the RFC.

rhagenson (commented):
I am neither against nor for this RFC in its current form. I think we have remaining issues to address before this can be accepted, although I like the direction and some of the precedent it is setting. My main concerns right now are:

+ Decide on how `RewindableIterator[A]` will be defined -- either by overlap with `Iterator[A]` (not my preference) or by type alias of `Iterator[A] & Rewindable[A]` (my preference)
+ Ensure that the precedent of "simpler types" due to an expectation of client disinterest in implementation does not extend beyond this RFC
+ Remove presumptions of the internal concerns of others -- it is possible to justify this change without speaking for others

text/0000-array-private-classes
SeanTAllen: This will lead to an explosion of types in the standard library and in this case, they will need to be in builtin unless Array doesn't implement the same iterator interface as things like Map etc in collections which strikes me as weird.

A key goal is to keep builtin as small as possible. If this wasn't going to start us towards either

- lots of new interfaces in the standard library
- having array be a special case

then I might be in favor of this. As is, I am not.

text/0000-array-private-classes
SeanTAllen: +1

pmetras:
Also, the iterator contract does not specify if a collection can be updated while iterated. This is a frequent source of bugs in other languages, even sequential ones. I haven't thought how to deal with this situation with Pony and if capabilities can help there. But suppose that we allow some collections to be updated by other actors that we had a new function `updated` to notify an iterator when the iterated collection has been changed (new elements added, removed or updated) and let the iterator implementer decide what to do. In which iterator interface would you add this `updated` function?

If it were added by combining different types, for instance like `(Iterator[A] & Updatable[A])`, implementors would have the flexibility to define precise types, compared with the suggested `CollectionIterator` interface.

Here, `Updatable` is something like
```pony
interface Updatable[A]
  fun updated(): A
    """Notifies that the type `A` has been updated and returns the new content."""
```

Please explain why adding `CollectionIterator` is a better solution than combining types so these arguments can be added to the RFC.

text/0000-array-private-classes
pmetras: DONE

text/0000-array-private-classes
pmetras: Checking that point, the only change in tests I can find is in `test/libponyc/util.cc` and that's a change of names.

pmetras: DONE.

text/0000-array-private-classes
pmetras: DONE

text/0000-array-private-classes
pmetras: DONE

text/0000-array-private-classes
pmetras: No, it was only an example I found while surveying how existing code were using the concrete class on Github and that I used to demonstrate the type of impact in the RFC. I saw that in the few examples I found, these concrete class were not used for their implementation, but they were used as iterators only. In the example, I showed how an existing piece of code could be rewritten using the new functions signatures while still giving the same results. And as a consequence, the code will be simpler to understand.

text/0000-array-private-classes
pmetras: The objective of the RFC is not about concrete type vs virtual ones, but more about applying [Liskov principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle). And it's difficult to say that a principle applies only on a special RFC... My initial questioning was about only about `Array` iterators and now it extends to all collection iterators, where we have to define what collection means (because collection data structures are not limited to the collection package).

SeanTAllen: I would vote against the RFC on those grounds. If you want to include it in the motivation you can, but I won't be in support in that case.

text/0000-array-private-classes
pmetras: I don't understand what is finally desired here. What would be the `RewindableIterator` (or `CollectionIterator`) definition so that I can include it in the RFC text?

ergl:
@pmetras This part:

> I don't see a major advantage in creating a new type of iterator for collections only, when both iterators are builtin. `Iterator` could be updated to define `rewind` with a default implementation, and neither client's nor implementer's code won't be impacted at all...

would not work for iterators that define a streaming sequence. You could, potentially, define an iterator over an infinite stream of events that can't possibly be rewinded, unless you kept all previous elements in memory. We don't want to constraint future uses of the `Iterator` interface so that _all_ implementations have to be rewindable.

pmetras:
@ergl , perhaps I did not express correctly my concern. The important word is **for collection only**. I have [previously suggested naming it `RewindableIterator` or better, in my opinion, having a `Rewindable` interface](https://github.com/ponylang/rfcs/pull/193#issuecomment-1023256406).

If `Iterator` is defined as an interface with a default implementation of `rewind` that raises an error or does nothing or create a new iterator (behaviour to be defined later), this type could be used even by streams or infinite sequenses that can't be rewound. For instance:

```pony
interface Iterator[A}
  fun has_next(): Bool
  fun next(): A ?
  fun rewind() ? => error
    """By default, iterators can't be reset"""
```

I don't think that such interface is the best solution, but at least it prevents having two types with one that is difficult to limit to collections usage.

SeanTAllen:
@pmetras This RFC should not modify Iterator[A]. No one is suggesting anything that would result in a change to `Iterator`.

pmetras:
I've added the rendered RFC at https://pmetras.github.io/rfcs/ and updated the text.

Now, I've just to find some time to see how to implement it and measure the impact.

rhagenson:
@pmetras I can only speak for myself, but I am not going to accept an RFC which breaks from our RFC procedure. There is no need to add HTML commits as GitHub will render the RFC's Markdown directly. You can link to that rendering like I did in #190

pmetras:
@rhagenson , please explain how to have it rendered and how to get a rendered URL. I wasn't able to find it automatically rendered as I wanted to be able to read it and I would have save a few hours today. So I used pandoc and followed Github documentation...

rhagenson:
@pmetras Thank you! Will you update your first comment with the link to the GitHub rendered Markdown?

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+This design principle is applied to the other collection classes that expose
+internals too like:
+
+* `List`
+* `Map`
+* persistent `Map`
+* `Vec`
+* persistent `Vec`
+* `Set`
+* `Itertools`
```
rhagenson, line 31: `itertools` seems out of place to me here. I see no internals of `itertools` being exposed. `Iter` is the only public type in `itertools`.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+(can be re-start from first value).
+
+This change remains compatible with the existing code base but for client code
+that is directly using the classes `*Keys`, `*Values` and `*Pairs`. A search on
+Github shows that the impact is very limited.
+
+To quote Antoine de Saint-Exupéry:
+
+> Perfection is achieved, not when there is nothing more to add, but when there
+> is nothing left to take away.
```
rhagenson, line 69: To quote from the [Pony Philosophy](https://www.ponylang.io/discover/#the-pony-philosophy):

> ## Performance
> Runtime speed is more important than everything except correctness. If performance must be sacrificed for correctness, try to come up with a new way to do things. The faster the program can get stuff done, the better. This is more important than anything except a correct result.
>
> ## Simplicity
> Simplicity can be sacrificed for performance. It is more important for the interface to be simple than the implementation. The faster the programmer can get stuff done, the better. It’s ok to make things a bit harder on the programmer to improve performance, but it’s more important to make things easier on the programmer than it is to make things easier on the language/runtime.

If making this change to abstract types has a performance impact than it is not inline with our driving philosophy which places performance at a higher priority than simplicity.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+Currently the API makes a guarantee that the return value of each of these functions
+is a specific concrete class with a specific public name. That particular guarantee
+does not actually serve any practical needs of users (except, arguably, the virtual
+calls caveat that was mentioned on [Github discussion](https://github.com/ponylang/rfcs/pull/193#issuecomment-1009733064)). If it is not helpful to users in practice,
+it can be removed. And if it can be removed, then it should be removed, in the
+name of taking another step toward stabilizing the Pony standard library's API.
```
rhagenson, line 76: This linked issue comment is in line with my previous review comment concerning performance taking priority over simplicity in the Pony Philosophy. I would like to see some evidence that the compiler as it is today devirtualize the calls and as such we get both performance (no worse than today) and simplicity (improved by this RFC).

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+types.
+
+It'll always be easier to get more specific with these return types than more
+general. So we should default to offering the most general return type that is
+still useful to users, and ratchet down to more specific subtypes as needed (as
+tangible use cases for more specific return types are discovered). Ratcheting down
+will always be possible without a breaking change, but ratcheting up will be a
+breaking change each time. It's best to make such "ratchet up" breaking changes
+like this RFC before we reach Pony 1.0.0 as part of a general effort to make Pony
+and its standard library stable.
```
rhagenson, line 92: I do not agree with this as motivation. It is always easier to start with many concrete implementations/returns which are then used to derive a generic implementation/return and trying to be generic before concrete presents more room for instability.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+# Detailed design
+
+Iterating functions in collections `keys`, `values` and `pairs` are changed to
```
rhagenson, line 96: In your example they return a `CollectionIterator`, not an `Iterator`.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+  fun has_next(): Bool
+    """
+    Return `true` when function `next` can be called to get next iteration item.
+    """
+
+  fun ref next(): A ?
+    """
+    Return the next item of the iteration or an error in case there are no other
+    items. A previous call to `has_next` check if we can continue iteration.
+    """
```
rhagenson, line 144: It may be my personal preference or lack of knowledge on the topic, but is there a reason to repeat these methods which are found on `Iterator`?

If these methods must be repeated I would prefer we add tests which check that the properties of `CollectionIterator` in the stdlib match `Iterator`.  That is, the elements from [Implementing Iterators](https://stdlib.ponylang.io/builtin-Iterator/#implementing-iterators) hold for "implementing CollectionIterators" just as well.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+  fun ref rewind(): Iterator[A]^
+    """
+    Start the iterator over again from the beginning.
+    """
+```
+
+The code of the standard library is adapted to remove use of these now private
+classes, mainly in tests. Here are the files that must be changed:
```
rhagenson, line 153: What private classes? Isn't this RFC about making now public classes into private classes?

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+* `packages/builtin/array.pony` as shown above
+* `packages/itertools/iter.pony` in function `cycle`
```
rhagenson, line 156: I see that in source `cycle` has a type annotation to `ArrayValues[A!, Array[A!]]`. This would indeed need to change, but does clarify by previous question/concern of where `itertools` needs to change.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+-  fun keys(): MapKeys[K, V, H, this->HashMap[K, V, H]]^ =>
++  fun keys(): Iterator[this->K]^ =>
+-  fun values(): MapValues[K, V, H, this->HashMap[K, V, H]]^ =>
++  fun values(): Iterator[this->V]^ =>
+```
+
+we see that we iterate over keys (`K` type) or values (`V`). The initial signature
+gave complex generic classes that the user has to find in the documentation to
+understand that they are iterators, and then dive deeper to understand on what
+they iterate.
```
rhagenson, line 245: I do think these new signatures are more readily understood.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+# How We Teach This
+
+This change keeps the code compatible in the vast majority of cases. When client
+classes are defining objects of these now private types, the reason is usually
```
rhagenson, line 250: This RFC would be making them private types, but as of now they are now public types.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+to get access to the function `rewind` that was not defined in `Iterator`. By
+adding the interface `CollectionIterator`, client code can easily be adapted,
+replacing `ArrayValues[A]` by `CollectionIterator[A]`.
+
+Also, client code generally uses these functions to iterate on the returned types
+and does not try to access the iterator directly but is interested by the iterated
+items. When client code refers to the iterator type, that's generally useless and
+the code can be rewritten to be made shorter and more future proof.
```
rhagenson, line 258: How is referring to a type useless? Code that is rewritten to update a referenced type is neither more nor less "useless" than before.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+  fun values(): CollectionIterator[Xml2node]^ ? =>
+    if (allocated) then
+      nodearray.values()
+    else
+      error
+    end
+```
+
+In this sample, the developer was not really concerned by the type of the iterator
+but that the `values` function must return an `CollectionIterator` over `Xml2node`.
```
rhagenson, line 288: Again you cannot speak on behalf of the mindset of the developer about whether or not Red was "not really concerned" about an aspect of his code.

You splice your changes into your explanation here in a way that suggests Red was only concerned that `values` returns a yet unavailable `CollectionIterator` type. Which gets at my comment just above that you cannot speak on behalf of the mindset of the developer here.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+The new version makes the code simpler to understand.
+
+This change in `array.pony` and other collections will break such code but it
+can be easily adapted to use the new API. And it will make the standard library
+easier to learn by reducing the number of public types.
```
rhagenson, line 293: I do agree that reducing the number of public types makes it initially easier to learn Pony. However, it the changes suggested here result in a performance decrease I cannot support these changes. My reason being that it takes only slightly longer to learn what `ArrayValues`, `ArrayKeys` and `ArrayPairs` return then have that knowledge "banked" for future use of `Array`.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+# How We Test This
+
+Pony tests must continue to pass. No additional tests are need as after review
+the existing coverage in Pony standard library tests is sufficient
```
rhagenson, line 298: I would like to see tests added that guarantee all `CollectorIterator`s follow the rules for implementing `Iterator`s. They "should" but as part of duplicating the methods of `Iterators` test to ensure the absence of future regressions need to be added.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+Will break any existing code that uses any of the classes that are currently
+public and will be made private by this RFC.
+
+# Alternatives
+
+1. Stay as is. Continue the
+[discussion on Zulip](https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/Make.20Array.20iterators.20private).
+
+2. Update the existing concrete classes to include `rewind` where needed.
```
rhagenson, line 310: How is this an alternative?

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+    Rewind the type `A`.
+    """
+```
+
+Then one can create a rewindable iterator by combining these two interfaces into
+a intersection type:
+
+```pony
+type CollectionIterator[A] is (Iterator[A] & Rewindable[Iterator[A]])
+```
```
rhagenson, line 328: This is my preference, aka how I would write my code, but I have had conversations which swayed me against this. I do not, as I sit here now, recall the reasoning that swayed me, but I am leaving a review comment here to voice my position on this alternative.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+the existing coverage in Pony standard library tests is sufficient
+
+# Drawbacks
+
+Will break any existing code that uses any of the classes that are currently
+public and will be made private by this RFC.
+
+# Alternatives
+
+1. Stay as is. Continue the
```
rhagenson, line 307: Why would we continue the conversation of keeping this as is? If no change is needed then no conversation is needed? Are you suggesting other changes which should be pulled into this RFC from Zulip?

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+4. Create intersection type by defining a `Rewindable` interface and combine
+with `Iterator`.
+5. Change `Iterator` to add `rewind`.
+
+| Justification        /      Alternative # |  1  |  2  |  3  |  4  |  5  |
+|:------------------------------------------|:---:|:---:|:---:|:---:|:---:|
+| a. Simpler stdlib API                     | --- | --- |  +  | +++ | +++ |
+| b. Simpler function signatures            | --- | --- |  +  | +++ | +++ |
+| c. Stay compatible with existing stdlib   | +++ | +++ |  +  | +++ |  +  |
+| d. Evolutive                              | --- |  +  | +++ | +++ |  +  |
```
rhagenson, line 353: Changes for the sake of changes is not evolution -- there must be positive or negative selection which lead microevolution into macroevolution. Put another way, if all the suggested changes have negative effects than it is still "evolution" to do nothing as we do not bring those negative effects forward with us.

Side note: I am a career "full stack biologist" who has done work ranging from cancer to conservation to COVID-19 -- Pony is the project that I volunteer my time with when **not** working on the "evolving nature" of biological problems.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+| e. No impact on compiler                  | +++ | +++ |  -  | +++ |  -  |
+| f. No impact on performance               | +++ | +++ | +++ | +++ | +++ |
```
rhagenson, line 355: Have you benchmarked these changes to ensure no performance impact?

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+| g. Limit stdlib pollution with interfaces | +++ | +++ |  +  |  +  | +++ |
+| g. Limit stdlib pollution with classes    | --- | --- | +++ | ++  | +++ |
```
rhagenson, line 357: Why is "g" repeated?

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+# Unresolved questions
+
+- Must analyze how `Range` and `Reverse` would be impacted if defined as
+`CollectionIterator`. Particularly, impact on existing client code.
+- Possible candidates to be analyzed: `StringBytes`, `StringRunes`, `ByteSeqIter`
+and probably others will be found in stdlib code.
+- Understand why the intersection type proposal is not considered as it seems to
+provide a better flexibility and is easier to understand.
```
rhagenson, line 366: None of these are questions.

rhagenson (commented):
I have left a large number of comments. My overall impression is still that I am in favor of the simpler return types, but there are issues that need to be addressed with respect to performance impact and ensuring the motivation for these changes are not setting a precedent for hiding all implementation details based on a notion of strict information hiding.

text/0000-array-private-classes.md
rhagenson: I addressed this below. There are references to `ArrayValues` within `itertools` which would need to change. Having `itertools` in this list is still out of place though as this list is for "other collection classes that expose internals too" which `itertools` is neither "[an]other collection class" nor does it "expose internals too".

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+# Summary
+
+Collection classes should not expose internal classes through iterators functions.
+
+Following information hiding design principle, the builtin classes of collection
+data structures must not be made visible through iterator functions like `ArrayKeys`,
+`ArrayValues` and `ArrayPairs`. These classes can be made private as they are
+only used as return types for `Array` functions `keys`, `values` and `pairs`.
+The return values for these functions are changed to the more general interface
+`Iterator`.
```
SeanTAllen, line 17: without "Following information hiding design principle"

being removed. I can't support this RFC. There's no "design principle" that I want to codify into Pony via RFC as a precedent for making changes.

text/0000-array-private-classes.md
```diff
@@ -0,0 +1,366 @@
...
+
+A new interface `CollectionIterator` is defined to allow for rewindable iterators,
+like it is the case for `Array` `values`.
+
+This design principle is applied to the other collection classes that expose
+internals too like:
```
SeanTAllen, line 23: Again, no, I am not in favor of anything in here that I am in favor of based on an abstract design principle. I can't support this language being included.

SeanTAllen:
I've repeated numerous times that I am against setting precedent based on a design principle. I didn't read the entire new version of this RFC. I read up to the 3rd mention of this being done because of a design principle and stopped. 

I'm against this RFC and at this point, I'm not going to invest more time in it. I can't support the motivation given, do not support it being used to set precedent going forward and have stated that more than once. I have many other Pony related things to do and so I am voicing my opposition to this RFC. I would welcome an RFC that meets the reasoning stated here: https://github.com/ponylang/rfcs/pull/193#issuecomment-1021509909.

As is, it seems clear to me this RFC is not going to be written taking that motivation into account.

pmetras:
I won't update this text anymore. It seems my English vocabulary and wording are not adequate to describe the objective of the RFC. Anybody who wants to support this proposal can take over and update this text.

rhagenson:
@pmetras Thank you for the work you have put into this RFC.

I am in support of the core of the RFC so I am still hoping to see it through to the end. However, I am already working on another RFC (#200) which I expect will draw most of my attention. Is there anybody else in the Pony community with the bandwidth needed to take over this RFC?

PR #194, opened 2022-01-13
Add RFC to expose runtime scheduler information

SeanTAllen:


jemc (commented):
Didn't we discuss in Zulip that we want to require an object capability to make use of these functions?

If you've decided against that, can you at least add that option to the "Alternatives" section of the RFC, so the choice will be an explicit one?

SeanTAllen:
I had forgotten that. I'll add that.

SeanTAllen:
@jemc added.

text/0000-expose-scheduler-thread-info-in-stdlib.md
```diff
@@ -0,0 +1,86 @@
...
+
+- pony_schedulers
+- pony_active_schedulers
+
+which will be marked as `PONY_API`.
+
+Additionally, we will define a single auth required to access runtime information. The auth will be defined in a file called `auth.pony`:
+
+```pony
+type RuntimeInfoAuth is (AmbientAuth | SchedulerInfoAuth)
```
jemc, line 56: The naming here is weird to me - but maybe I guess I can just wait until we resolve the Auth conventions in another way, as we've been discussing elsewhere, because the union aspect is causing most of the weirdness.

SeanTAllen: This follows a pattern that currently exists. I figured it was good to do that now and be consistent given my expectation that this will go in before any object capability change.

jemc: Aside from the baseline union weirdness, It's mainly weird to me that:
- we are treating the name fragments `RuntimeInfo` and `SchedulerInfo` as if they were synonymous in meaning, so there is a redundancy here (they are not as similar in meaning as say `TCPListen` and `TCPListener`).
- we have the name fragment `RuntimeInfo` corresponding to the name of the package (`runtime_info`), but we have the name fragment `SchedulerInfo` corresponding to the primitive `Scheduler` - so where did the `Info` part come from?

SeanTAllen: the package is runtime_info, with a primitive for called "Scheduler" to me that means because this is "runtime info" that "Scheduler" here means it provides "runtime scheduler info", so I called it "SchedulerInfo".

SeanTAllen: The auth could be called "AccessSchedulerInfoAuth".

SeanTAllen: @jemc given that the two of us have agreed to the general shape of an RFC to remove the type unions like this, I'm not real concerned with the naming. I think when we write up the object capability RFC we should change to remove the `RuntimeInfoAuth` and leave only `SchedulerInfoAuth`.

jemc: What I'd prefer most with the naming is full consistency:
- the library being called `runtime_info`
- the primitive being called `RuntimeInfo`
- the auth primitive being called `RuntimeInfoAuth`

SeanTAllen: I think we should have primitives for different parts of the runtime, not a single RuntimeInfo primitive. I thought that is what we agreed upon in our Zulip conversation.

jemc: Eh, I don't really like the idea of splitting them up into separate primitives if they are all just primitives that have functions to return information about the runtime.

But if you want to have one that is only about the scheduler, I'd suggest calling it `SchedulerInfo` instead of only `Scheduler` - the latter name would seem more appropriate for an API that has an active component where you can use it for actually scheduling things. The latter is also more likely to collide with a user-code type name that wants to use the `runtime_info` package to intellligently scheduler work onto pools of actors.

At any rate, I'm not going to vote against the RFC on naming alone - I'm just trying to give suggestions to try to make the library less annoying to users like me who get uptight about naming conventions.

jemc: This was discussed on the sync call today. As I mentioned there, I'm still not a fan of this naming within Pony, but it does match the way types are named in many other Pony libraries, including the standard library. And it drives me crazy in all of those other places too, but my complaint is more with the way Pony `use` statements work than the naming conventions itself.

So any of those complaints I have are definitely outside the scope of this RFC.

SeanTAllen:
Accepted. @Theodus do you have the time to do the RFC song and dance with this?

SeanTAllen:
@Theodus I did the tracking.

PR #195, opened 2022-01-15
Change FFI to be disallowed by default

redvers:
As discussed here: https://ponylang.zulipchat.com/#narrow/stream/189959-RFCs/topic/FFI.20change

### DIFF
--- a/text/0000-ffi-safe-default.md
+++ b/text/0000-ffi-safe-default.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-The current default is to allow FFI calls from any package. The optional `--safe` flag takes an argument of a colon delemeted list of packages and limits FFI calls to those and `builtin`.
+The current default is to allow FFI calls from any package. The optional `--safe` flag takes an argument of a colon delimited list of packages and limits FFI calls to those and `builtin`.
 
 The purpose of this RFC is three-fold:
 
--- a/text/0000-ffi-safe-default.md
+++ b/text/0000-ffi-safe-default.md
@@ -1,0 +1,77 @@
+- Feature Name: builtin-ffi-safe-only-as-default
+- Start Date: 2022-01-15
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The current default is to allow FFI calls from any package. The optional `--safe` flag takes an argument of a colon delemeted list of packages and limits FFI calls to those and `builtin`.
+
+The purpose of this RFC is three-fold:
+
+* Change default from "allow all" to "deny all except stdlib"
+* Rename the `--safe` flag to the more explicit `--allow-ffi`
+* Provide an error message should a non-existent package be in the allow-list
+
+
+# Motivation
+
+Object Capabilities, Referencial Capabilities, and Type-Safety are how the pony compiler provides its guarantees. FFI calls bypass the pony compiler's checks allowing third party packages be able to make FFI calls. This is an unsafe default.
+
+If a package requires FFI then the user should be explicitly aware that this is the case.  By mandating the user acknowledge this with the `--allow-ffi` flag, we have confirmed they are aware and have had the opportunity to do their due diligence or forgo its use completely.
+
+An error should be raised if a non-existant package is in the pre-compiled "stdlib allow-list" or provided to `--allow-ffi` to ensure that if in the future we remove an approved library from stdlib and forget to remove it from the allow-list then it does not create a namespace that can be exploited.
+
+
+# Detailed design
+
+stdlib is not builtin so the packages that are in stdlib will need to be enumerated in the allow-list.
+
+* assert
+* backpressure
+* builtin
+* capsicum
+* collections
+* debug
+* files
+* format
+* net
+* ponybench
+* ponytest
+* process
+* term
+* time
+* serialise
+* signals
+
+We should not add to the allow-list packages not in stdlib but still maintained by ponylang as there is no guarantee at compile-time that the "regex" package the user has in their tree is the one from ponylang.
+
+
+# How We Teach This
+
+The packages section should mention that packages may request FFI access and a top-level description of what that means.
+
+The FFI section should mention the flag requirement if your FFI calls are in a package (and a reminder to document it)
+
+
+# How We Test This
+
+Our existing tests should cover if we break any existing functionality.
+
+Some of our existing tests use FFI so will need their build flags modified to allow them to compile.
+
+
+# Drawbacks
+
+* Breaks existing code which uses non-stdlib packages that use FFI.
+
+
+# Alternatives
+
+Keep things as they are with the default that any package can make FFI calls.
+
+
+# Unresolved questions
+
+Should there be an option for `--allow-ffi=*`?
+


text/0000-ffi-safe-default.md
```diff
@@ -0,0 +1,77 @@
+- Feature Name: builtin-ffi-safe-only-as-default
+- Start Date: 2022-01-15
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The current default is to allow FFI calls from any package. The optional `--safe` flag takes an argument of a colon delemeted list of packages and limits FFI calls to those and `builtin`.
```
SeanTAllen:
```suggestion
The current default is to allow FFI calls from any package. The optional `--safe` flag takes an argument of a colon delimited list of packages and limits FFI calls to those and `builtin`.
```

text/0000-ffi-safe-default.md
```diff
@@ -0,0 +1,77 @@
...
+* time
+* serialise
+* signals
+
+We should not add to the allow-list packages not in stdlib but still maintained by ponylang as there is no guarantee at compile-time that the "regex" package the user has in their tree is the one from ponylang.
+
+
+# How We Teach This
+
+The packages section should mention that packages may request FFI access and a top-level description of what that means.
```
SeanTAllen, line 52: packages section?

text/0000-ffi-safe-default.md
```diff
@@ -0,0 +1,77 @@
...
+
+The FFI section should mention the flag requirement if your FFI calls are in a package (and a reminder to document it)
```
SeanTAllen, line 54: ffi section?

SeanTAllen:
So I think from "How to test" there's no mention of how we know in the future that we haven't broken the new functionality. We need a testing strategy and it should be detailed in the RFC as the implementation of said testing will be part of the acceptance criteria for the change.

SeanTAllen:
There's no details in here on how we go about maintaining the list of packages that are auto-whitelisted. That's a very important design decision that needs to be covered in detailed design, alternatives should be mentioned as well in the appropriate alternatives section. 

Is it hardcoded? If yes, that has a drawback of being fragile as we need to remember to update it any time a new package is added to the standard library.

Is it autogenerated? I kind of like this idea. We could have a github action that on each PR builds an alphabetical list of packages in the standard library and compares them to a text file that is included into the compiler as part of the build (perhaps a c header, perhaps something else - needs discussion) it could then either automatically update said file or fail the build. I lean towards failing the build, I don't like things that add to PRs without human input.

Drawbacks for each approach should be considered and documented and the alternatives should be enumerated.

SeanTAllen:
Drawbacks should list that this is additional work for folks to use 3rd party libraries that they don't currently need to do. It complicates the ponyc command line for those packages. Makes adding FFI to a library that didn't previously have a breaking change. There are probably other drawbacks as well, but those are one that I can think of right now.

SeanTAllen:
I think this RFC has a lot of unresolved questions that aren't yet detailed. All questions that aren't answered yet by the RFC (which is a lot of them) should be enumerated. We should have resolved them all before we accept the RFC but we can discuss it and work on it with questions remain unresolved.

text/0000-ffi-safe-default.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+# Alternatives
+
+Keep things as they are with the default that any package can make FFI calls.
+
+
+# Unresolved questions
+
+Should there be an option for `--allow-ffi=*`?
```
SeanTAllen, line 76: Sorry, what does this mean? Allow anything? I'm not sure what this is denoting as an unresolved question.

text/0000-ffi-safe-default.md
redvers: This section:  https://tutorial.ponylang.io/packages/index.html#the-package-structure

text/0000-ffi-safe-default.md
redvers: This section: https://tutorial.ponylang.io/c-ffi/index.html

(I'm guessing that you're looking for me to be more specific?)

SeanTAllen: I had no idea what FFI section. So yes more specific but not just want section, but where in the section, how should be in taught?

text/0000-ffi-safe-default.md
```diff
@@ -0,0 +1,77 @@
...
+- Pony Issue:
+
+# Summary
+
+The current default is to allow FFI calls from any package. The optional `--safe` flag takes an argument of a colon delemeted list of packages and limits FFI calls to those and `builtin`.
+
+The purpose of this RFC is three-fold:
+
+* Change default from "allow all" to "deny all except stdlib"
+* Rename the `--safe` flag to the more explicit `--allow-ffi`
```
jasoncarr0, line 13: What about --trusted? --trusted-packages?

SeanTAllen: I prefer the explicit "allow-ffi", I don't think "trusted" as is explicit.

text/0000-ffi-safe-default.md
```diff
@@ -0,0 +1,77 @@
...
+* net
+* ponybench
+* ponytest
+* process
+* term
+* time
+* serialise
+* signals
+
+We should not add to the allow-list packages not in stdlib but still maintained by ponylang as there is no guarantee at compile-time that the "regex" package the user has in their tree is the one from ponylang.
```
ergl, line 47: This problem also affects other packages in the standard library: how do we know that "assert" refers to the standard library package?

Package names are not unique, so the current way of enumerating them would not be enough to prevent a 3rd party package to have the same name as a standard library one. This would allow them to gain FFI privileges due to a name clash.

redvers: That's resolved by the paths ordering:

https://github.com/ponylang/ponyc/issues/3779

ergl: Edit: Simplified example.

@redvers If I'm not mistaken, I think that PR only applies to `builtin`, not to the entire standard library. But I don't think it would fix the problem. Imagine the following scenario:

```pony
// Directory structure:
// .
// ├── assert
// │   └── lib.pony
// └── main.pony

// assert/lib.pony
use @printf[I32](fmt: Pointer[U8] tag, ...)

primitive Assert
  fun apply(test: Bool, msg: String = "") ? =>
      @printf("Surprise!\n".cpointer())
      if not test then
        error
      else
        None
      end

// main.pony
use "assert" // Refers to "assert" above

actor Main
  new create(env: Env) =>
    try Assert(true, "foo")? end
```

The code above doesn't depend on the `assert` package in the standard library, and yet compiling this code with `ponyc --safe=assert` would cause the resulting program to print "Surprise!" to the screen.

If ponylang/ponyc#3779 applied to all packages in the standard library, the above would also be the case, as long as no package directly imports "assert" from the standard library. Only when importing "assert" from the standard library would the code above fail to compile.

SeanTAllen: I dont believe this should be an issue. Here is how the safe works.

on package system init, we do the following (see package.c for more)

- add package paths AFTER the standard library path
- convert the "safe list" to fully qualified paths starting with standard library

the code however is tricky and could be changed. I think the most reasonable thing based on how "find path" is to make sure the included paths are absolute paths to the standard library packages. This would mean they need to be generated at the time the compiler starts up.

ergl: @SeanTAllen Good point about absolute paths. For some reason I though `safe` only worked on package names, hence my worry about external libraries using the same name.

SeanTAllen: @ergl as the code currently stands, only package name would still be fine because of how we turn names into "the package on disk", that said, absolute paths protects against a change in how we do that having an unintended consequence with the items that are automatically in the "allow list" so I think it is a very good idea. Thanks for bringing up the topic.

ergl: @SeanTAllen We can expand on the sync call, since I don't know enough about how packages work in the compiler. However, I don't think we're putting _all_ the packages from the standard library before 3rd party packages. We only do that for `builtin`.

Consider the code I pasted above (and the repo [here](https://github.com/ergl/pony_test_allow_ffi)):

- When we do `ponyc --safe=assert`, that "assert" should probably talk about the standard library package, but ends up being treated as a 3rd party package.

- If we uncomment `use a = "assert"` [from this line](https://github.com/ergl/pony_test_allow_ffi/blob/f134b2766e0bb3d87fcfb4b50c42259c7f98e39d/assert/lib.pony#L3) in our fake "assert" package, the compiler will complain about the standard library, not about our own code:

```
# ponyc --safe=assert . -o build -b main
Building builtin -> /usr/local/lib/pony/0.46.0-eea94aed/packages/builtin
Building . -> /Users/ryan/Downloads/test_safe
Building assert -> /Users/ryan/Downloads/test_safe/assert
Building assert -> /usr/local/lib/pony/0.46.0-eea94aed/packages/assert
Error:
/usr/local/lib/pony/0.46.0-eea94aed/packages/assert/assert.pony:30:18: this package isn't allowed to do C FFI
        @fprintf(@pony_os_stderr(), "%s\n".cstring(), msg.cstring())
                 ^
Error:
/usr/local/lib/pony/0.46.0-eea94aed/packages/assert/assert.pony:30:9: this package isn't allowed to do C FFI
        @fprintf(@pony_os_stderr(), "%s\n".cstring(), msg.cstring())
        ^
```

SeanTAllen: That's a bit of a different scenario and I don't think it is problematic. You the user have explicitly overriden it with your own code. That said, absolute paths are best and avoid this in general.

So what you did was you said "I am going to override assert from the standard library with this assert here. You can override any part of the standard library by placing an override package at the base of your application. 

Note here that by putting assert in the base directory of your application, it is not by definition "a 3rd party package". It is your own package that you are overriding the standard library with.

SeanTAllen: A note for clarity, if the path in your use is reachable from your base then it isn't considered "3rd party",

so if you had something like this:

app/main.pony
lib/assert/assert.pony

and in app/main.pony you did:

```pony
use "../lib/assert"
```
then that is also not considered "3rd party" much like if you put --safe ../lib/assert. 

I think what "is and isn't in a trust boundary" in terms of overriding the standard library and if we make this FFI change should be in the packages section of the tutorial now and the ramifications for FFI if we make this change should also be noted in the FFI section.

rhagenson:
I do not see it in the Zulip conversation or in the RFC here, but I think it need to be part of our conversation: Is this change transitive? Should `--allow-ffi=package-i-trust` where `package-i-trust` uses `package-i-dont-trust` imply I trust `package-i-trust` to have sufficiently audited `package-i-dont-trust` and remained safe, or should I be required to use `--allow-ffi=package-i-trust,package-i-dont-trust` in order to compile?

SeanTAllen:
It is not transitive currently and it would remain the same. Explicitly noting that it isnt transitive would be a very good thing.

rhagenson:
I foresee that causing some user-level aggravation, but better than unsafe FFI being "trusted" by default -- which is the underlying purpose of this RFC to begin with.

SeanTAllen:
@rhagenson I look at it this way...

This saves me aggravation, I can use library foo and not worry if it currently uses FFI or has a new FFI package dependency later as the compiler will alert me.

Next step would be to add tooling that can check what FFI is used and store a delta so if that FFI changes... then weee! I get alerted. This is something we discussed a while ago. Mostly @jemc, @aturley, and I. 

The general topic was "allowing FFI, ok, but I don't want to have to audit for changes all the time for calls, I'd like to be able to get alerted to C call changes", now how that might work in practice is a giant handwave. But for me, I consider this "a good thing".

rhagenson:
I agree with it not being transitive for the same reasons you state. (My aggravation comment was more tongue-in-cheek about knowing no matter which decision is made there will be complaints that it was not the other decision.) Having it be non-transitive fits with the purpose here of requiring explicitly allowing FFI.

stefandd:
As a motivation, the proposal states that FFI calls are (potentially) unsafe and allowing them is an unsafe default. It then goes on to propose the stdlib to be on some kind of pre-compiled allow-list including mechanisms for removal from that list for those packages that once were but are no longer part of the stdlib.

I am a bit puzzled how motivation and proposed changes can go together smoothly. Are you proposing that when such unsafe FFI calls are made by contributors to the stdlib then they are *safer* than the ones made by 3rd party libary authors? Even if that were so, this would also ignore the possibility of the code into which the stdlib makes FFI calls (the LLVM runtime functions etc.) changing in a way that is unsafe. How is the stdlib any more secure from such effects than others unless we know all side effects to be caught in tests? Are these much better than the tests of external libraries?

rhagenson:
If I recall from discussion, the rationale for having stdlib be trusted by default is that Pony programs rely on stdlib being safe and by extension the FFI in it to be safe. It is not "safer" by virtue of being stdlib, but rather the base level of trust placed into any Pony program extends to its stdlib and the FFI within it.

stefandd:
> If I recall from discussion, the rationale for having stdlib be trusted by default is that Pony programs rely on stdlib being safe and by extension the FFI in it to be safe. It is not "safer" by virtue of being stdlib, but rather the base level of trust placed into any Pony program extends to its stdlib and the FFI within it.

Ok, but one could ask: on what grounds?

The way this is proposed partitions code authors into safe and unsafe in a bit wobbly way. Also, the compiler flag with the to-be-allowed-list passes all control the users anyhow. Is that any safer than it is now? Someone who wants to link a certain library because they need a functionality the stdlib is not providing, will want to use that library no matter what message the compiler puts on the terminal.

rhagenson:
On what grounds should we trust a Pony program to be safe? What point are you trying to make?

I understand a general sentiment that FFI is inherently unsafe, but not extending our distrust of it so far as to distrust every Pony program in existence.

stefandd:
> On what grounds should we trust a Pony program to be safe? What point are you trying to make?
> 
> I understand a general sentiment that FFI is inherently unsafe, but not extending our distrust of it so far as to distrust every Pony program in existence.

@rhagenson Sorry, in my unedited comments I mistook you for one of the proposers.

The point is that the stdlib is one of the biggest FFI users and I am sure that there is a lot of unsafe code in there. I don't understand why it should land on some kind of white list.
My actual point is on the other hand that I don't think it is productive to say: FFI is unsafe. We (Pony stdlib) write unsafe code, other lib authors write unsafe code. However, ours is to be trusted, external one is not since we didn't write. Therefore,  if you don't just use our code, you must put it an extra effort to get it to compile. It hollows out the premise of the RFC which was safety. It feels like:  we must trust stdlib so all the unsafe code in there is ok, but only that code. Additional code that calls to FFI is to be flagged as dangerous by the compiler.

Is that argument sound enough to make the change that this RFC is proposing and abolish the status quo which is that all FFI-calling packages and codes, stdlib or my own code, are seen as equally (un)safe by the compiler?

rhagenson:
No worries about the mistake. I take no offense; especially given the good debate of priority going on.

> The point is that the stdlib is one of the biggest FFI users and I am sure that there is a lot of unsafe code in there.

I absolutely agree on the part of "the biggest FFI users" as even if it is not by LOC, it is definitely by number of programs using it. As for "there is a lot of unsafe code there" I do not disagree, but I want to add to the conversation that *if* the FFI in stdlib is unsafe then that constitute a different problem than any 3rd party FFI being unsafe. We have to be able to trust the stdlib to be safe and if it is unsafe then any Pony program is unsafe and therefore it is the stdlib that must change, not our level of trust in it.

We seem to be approaching a point where we could easy begin talking past one another so I want to say that in my perspective I am ignoring the technical aspect of FFI which allows arbitrary code and therefore tosses out all the guarantees of Pony in favor of a guiding principle that the stdlib **must** adhere to (and therefore be just as safe as) the guarantees that Pony makes. A breach of safety in the stdlib (via FFI or in general) means the stdlib must change -- FFI does not change that. I agree with a move to FFI being disallowed by default because FFI throws out the guarantees of Pony. It is only the stdlib where we, as authors/owners/committers of the codebase, are taking on the full responsibility that the FFI within is safe.

ergl:
@stefandd There are two arguments for not trusting FFI by default:

1. To avoid potentially unsafe code or code that violates certain runtime invariants.
2. To avoid external code from bypassing the capability "sandbox" (quotes because it's not a real sandbox, just enforced by the compiler) and potentially performing all kinds of nasty things on your computer.

In this sense, we're trusting the standard library by default because it is written by the same group that writes the compiler and the runtime. Think of it like this: if you don't trust the maintainers of the stdlib, then it follows that you don't trust the compiler or the runtime.

As such, disallowing the standard library from doing FFI doesn't matter because the compiler or the runtime could already be performing malicious things. The compiler could even ignore your request to disallow the standard library from doing FFI.

stefandd:
@ergl I see your point.

I do however also worry that the "tooling" (https://github.com/ponylang/rfcs/pull/195#issuecomment-1030372227) that would somewhat slipstream this or alleviate the impact of the proposed changes will not anytime soon see the light of day.
In the past drastic changes such as completely cutting the FFI tuple call ability (because some users might have used unpacked C structs [or had some other ABI-related issues) were made swiftly but at the time promised to be temporary. That is many years ago and I still now am forced to write lots of shim code to call C functions that return or take struct values. Pretty much every change made to the FFI after 2016 meant a reduction in FFI functionality for the sake of safety and to protect users from evil actors (pun intended). I just worry a bit that the laudable progress towards safety will force users that want to write or use FFI wrappers for C libraries to eventually abandon the use Pony when they will run out of ways to use 3rd party libs in a somewhat painfree way.

SeanTAllen:
This RFC is going to be replaced by a much more extensive one from me. @redvers how do you feel about closing this one?

redvers:
@stefandd - From the RFC:

```
Object Capabilities, Referencial Capabilities, and Type-Safety
are how the pony compiler provides its guarantees. **FFI calls
bypass the pony compiler's checks allowing third party packages
be able to make FFI calls.** This is an unsafe default.

If a package requires FFI then the **user should be explicitly
aware** that this is the case. By **mandating the user acknowledge**
this with the --allow-ffi flag, we have confirmed they are aware
and have had the **opportunity to do their due diligence** or
forgo its use completely.
```

Let's take a hypothetical example.  Say, a library that left-pads text.

There's no need for a capabilty.  There's no need for FFI.

It should be completely safe to include in a program without in-depth review.

If a future version gains "_undocumented functionality_"™ such as a remote shell, that's the kind of thing that would need an FFI call and that this RFC was explicitly designed to address.

If you don't white-list stdlib then you're forcing the user to either explicitly enumerate all 16 stdlib packages that use FFI on the command line or, disable FFI checking completely.

@SeanTAllen  - of course, looking forward to seeing what you got.

redvers:
@stefandd - you wrote: " That is many years ago and I still now am forced to write lots of shim code to call C functions that return or take struct values."

If you're a heavy FFI user I'd love to compare notes.  The FFI subsystem is near and dear to me so please - let's chat.

PR #196, opened 2022-02-02
Change the standard library pattern for object capabilities

### LINKED ISSUE #177: Changes to Auth implementation
Sylvan shared an example of an alternative POC approach that he came up with that would allow env.root to be something besides just AmbientAuth | None: https://playground.ponylang.io/?gist=f0dca3abd61d28d2c9ad806393d3c025

SeanTAllen:
Closes #177

### DIFF
--- a/text/0000-change-stdlib-object-capabilities-pattern.md
+++ b/text/0000-change-stdlib-object-capabilities-pattern.md
@@ -1,0 +1,135 @@
+- Feature Name: Change standard library object capabilities pattern
+- Start Date: 2022-02-02
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Change the pattern for how object capabilities are used in the standard library.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


jemc (commented):
The `Backpressure` example used in this RFC is only one level "deep", and thus it doesn't demonstrate one other key detail that we need to solidify in this RFC:
- What is the convention for the parameter type of the constructor of an auth type that is more than one level "deep" in the conceptual hierarchy?

That is, does such a constructor accept :a: the union of all more powerful auths, or :b: only the nearest more powerful auth?

I believe that :a: is what we discussed doing, but I think that this RFC needs to explicitly reflect that.

SeanTAllen:
I felt that was covered as it remains as is and the RFC is written in terms of what changes. I can made that more explicit.

SeanTAllen:
@jemc the lack of change in object capability constructors has been explicitly noted.

jemc:
Looks good :+1:

SeanTAllen:
Accepted.

PR #198, opened 2022-02-04
Remove logger package from standard library

### LINKED ISSUE #181: Expand or extract logging package from stdlib
The baseline logging in Pony is incredibly simple, but leans perhaps too much so. In order to support RFCs like RFC 67 (#175), we may need to expand the interface to be more flexible in passing information, besides the simple string computations of the current logger design.

Pony's standard library is not meant to be "batteries-included", in implementing all features. Nevertheless, a standard interface for logging will allow more code to interact in the future, and ensure critically important consistency in output. There are many existing, mature logging facades in other languages, which we may use for reference. Some such mature facades include [SLF4J](https://en.wikipedia.org/wiki/SLF4J), [python's logging module](https://docs.python.org/3/library/logging.html), as well as competing standards such as [Flogger](https://github.com/google/flogger)

SeanTAllen:
Closes #181

### DIFF
--- a/text/0000-template.md
+++ b/text/0000-template.md
@@ -1,0 +1,60 @@
+- Feature Name: Remove logger package from standard library
+- Start Date: 2022-02-04
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Remove the logger package from the standard library and move to its own library under the ponylang org.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
.

jemc:
:+1:

SeanTAllen:
Accepted.

PR #199, opened 2022-02-15
Add ansi escape codes

csos95:


text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
+- Feature Name: add_ansi_escape_codes
+- Start Date: 2022-02-14
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the addition of more escape code functions to the ANSI primitive of the term package and the consolodation of the thirty-two existing color functions into four functions.
```
SeanTAllen:
```suggestion
This RFC proposes the addition of more escape code functions to the ANSI primitive of the term package and the consolidation of the thirty-two existing color functions into four functions.
```

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+# Motivation
+
+The purpose of these additional methods is to cover more of the standard ANSI escape codes within the stdlib.
+
+# Detailed design
+
+- Add escape codes for 3-bit, 4-bit, 8-bit, and 24-bit colors for the foreground and background.  
+The naming scheme of the color functions would be `{fg,bg}_{3,4,8,24}bit`.  
+The 3-bit, 4-bit, and 8-bit versions would take a U8 specifying the color.  
+The 24-bit versions would take three U8s specifying the r, g, and b values.
```
SeanTAllen: This is really hard to read rather formatting like this.

I suggest sections that are `##` for things to add, things to remove, things to change

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+# Detailed design
+
+- Add escape codes for 3-bit, 4-bit, 8-bit, and 24-bit colors for the foreground and background.  
+The naming scheme of the color functions would be `{fg,bg}_{3,4,8,24}bit`.  
+The 3-bit, 4-bit, and 8-bit versions would take a U8 specifying the color.  
+The 24-bit versions would take three U8s specifying the r, g, and b values.  
+- Remove the thirty-two named color functions in favor of the 3-bit and 4-bit functions.
+- Add escape codes for enter and leave alternate screen buffer.
+- Add escape code for the bell alert.
+- Change `clear` and `erase` functions to take a U8 specifying what part of the screen/line to clear.
```
SeanTAllen: I think that any breaking changes need justification. And time should be spent on them. If there is a breaking change section you can spend more time on that.

text/0000-add-ansi-escape-codes.md
SeanTAllen: Also newline whitespace would make this easier to read. I'm having a hard time taking it in.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+The naming scheme of the color functions would be `{fg,bg}_{3,4,8,24}bit`.  
+The 3-bit, 4-bit, and 8-bit versions would take a U8 specifying the color.  
+The 24-bit versions would take three U8s specifying the r, g, and b values.  
+- Remove the thirty-two named color functions in favor of the 3-bit and 4-bit functions.
+- Add escape codes for enter and leave alternate screen buffer.
+- Add escape code for the bell alert.
+- Change `clear` and `erase` functions to take a U8 specifying what part of the screen/line to clear.  
+The default for the escape codes if a value isn't given is zero.  
+The action these functions currently take is three for `clear` and one for `erase`.  
+I would like to have the default value in these functions be zero.
```
SeanTAllen: I don't understand what any of these means. More detail/explanation would be helpful.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+# How We Teach This
+
+The added functions would be documented as the existing functions on the `ANSI` primitive are and an example of usage would be added to `ANSI`.
+
+# How We Test This
+
+I could not find existing testing for the `ANSI` primitive.  
+If there are existing tests for these functions, additional tests in the same format would be added to ensure that the new functions return the correct escape codes.  
+Most of these functions are just directly returning a string of the escape code without any parameters so they're pretty straightforward.
```
SeanTAllen: "Most" suggests some aren't. Do the ones that aren't most need testing? Which aren't like this?

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Drawbacks
+
+The removal of the existing color functions and the change to the `clear` and `erase` functions (if the default of zero is used) are breaking changes.
```
SeanTAllen: All of these changes need justification. If you want, I would consider splitting this RFC into two, one with breaking changes only, one with the less "controversial" non-breaking changes.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Alternatives
+
+## Alternative to the rfc as a whole
+These escape codes can simply be printed independently of whatever is supported by functions on `ANSI`.
```
SeanTAllen:
```suggestion
## Alternative to the rfc as a whole

These escape codes can simply be printed independently of whatever is supported by functions on `ANSI`.  
```

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Drawbacks
+
+The removal of the existing color functions and the change to the `clear` and `erase` functions (if the default of zero is used) are breaking changes.
+
+# Alternatives
+
+## Alternative to the rfc as a whole
+These escape codes can simply be printed independently of whatever is supported by functions on `ANSI`.  
+Having them available on `ANSI` simply makes it so that users do not need to look the codes up themselves.
```
SeanTAllen: I don't understand what you mean by this.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Alternative to the breaking changes
+The existing color functions could remain and the 3-bit and 4-bit color functions could exist alongside them or not be implemented.
```
SeanTAllen:
```suggestion
## Alternative to the breaking changes

The existing color functions could remain and the 3-bit and 4-bit color functions could exist alongside them or not be implemented.  
```

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+These escape codes can simply be printed independently of whatever is supported by functions on `ANSI`.  
+Having them available on `ANSI` simply makes it so that users do not need to look the codes up themselves.
+
+## Alternative to the breaking changes
+The existing color functions could remain and the 3-bit and 4-bit color functions could exist alongside them or not be implemented.  
+The `clear` and `erase` functions could have default values that match their existing behavior.
+
+# Unresolved questions
+
+Should the breaking changes be made to simplify the interface?
```
SeanTAllen: This question really makes me think any breaking changes should be split into 1 or more different RFCs.

text/0000-add-ansi-escape-codes.md
csos95: I'll update this RFC to only include the additions and make another with the breaking changes.

SeanTAllen: Sounds good.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+I would like to have the default value in these functions be zero.
+- Add escape codes for scroll up/down.
+- Add escape codes for next/previous line.
+- Add escape code for cursor horizontal position (moves the cursor to column n in the current line).
+- Add escape codes for hide/show cursor.
+- Add escape codes for save/restore cursor position.
+- Add escape code for get cursor position (it is sent to stdin).
+- Add escape codes for various text styles: italic, strike, conceal, faint.
+
+If a value passed to any of the functions is invalid, such as 200 being passed to `fg_3bit`, an empty string would be returned.
```
rhagenson: Will you explain to me why you want to make this return a default value as opposed to producing an error?

Alternatively, why not use the type system to prevent such errors in the first place?

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
+- Feature Name: add_ansi_escape_codes
+- Start Date: 2022-02-14
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the addition of more escape code functions to the ANSI primitive of the term package and the consolidation of the thirty-two existing color functions into four functions.
```
rhagenson: Is this RFC adding and consolidating, or is the addition meant to result in consolidation?

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Motivation
+
+The purpose of these additional methods is to cover more of the standard ANSI escape codes within the stdlib.
+
+# Detailed design
+
+- Add escape codes for 3-bit, 4-bit, 8-bit, and 24-bit colors for the foreground and background.  
+The naming scheme of the color functions would be `{fg,bg}_{3,4,8,24}bit`.
```
rhagenson: In my quick review of ANSI escape codes, splitting the `{fg,bg}` does not make sense to me. It appears that different `U8` values either set foreground or background. I am using the table for [Colors](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors) in making this comment. 

What is the plan to handle, for example, `fg_4bit(40)` which should be invalid while `bg_4bit(40)` is valid?

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+
+The purpose of these additional methods is to cover more of the standard ANSI escape codes within the stdlib.
+
+# Detailed design
+
+- Add escape codes for 3-bit, 4-bit, 8-bit, and 24-bit colors for the foreground and background.  
+The naming scheme of the color functions would be `{fg,bg}_{3,4,8,24}bit`.  
+The 3-bit, 4-bit, and 8-bit versions would take a U8 specifying the color.  
+The 24-bit versions would take three U8s specifying the r, g, and b values.  
+- Remove the thirty-two named color functions in favor of the 3-bit and 4-bit functions.
```
rhagenson: I am not a fan of removing these functions entirely. They do overlap with the new 3-bit and 4-bit functions, but only partially. Might I propose creating `FG` and `BG` primitives or simple a `Color` primitive to move this functionality to? The usage could then appear as something like: `ASNI.fg4bit(Color.red())` or `ASNI.fg4bit(FG.red())`

Better yet (in my opinion) is leverage types a bit more to make it impossible to do something like `ASNI.fg4bit(Color.red_bg())` where an `fg` function is being used with a `bg` color.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+- Add escape codes for enter and leave alternate screen buffer.
+- Add escape code for the bell alert.
+- Change `clear` and `erase` functions to take a U8 specifying what part of the screen/line to clear.  
+The default for the escape codes if a value isn't given is zero.
```
rhagenson: I presume you mean that `fun fg4bit(code: U8 = 0)` and not that the default return is a "zero value" which with `String` would be the empty string. Correct?

text/0000-add-ansi-escape-codes.md
rhagenson: I am also confused by your intent here.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+The default for the escape codes if a value isn't given is zero.  
+The action these functions currently take is three for `clear` and one for `erase`.  
+I would like to have the default value in these functions be zero.
+- Add escape codes for scroll up/down.
+- Add escape codes for next/previous line.
+- Add escape code for cursor horizontal position (moves the cursor to column n in the current line).
+- Add escape codes for hide/show cursor.
+- Add escape codes for save/restore cursor position.
+- Add escape code for get cursor position (it is sent to stdin).
+- Add escape codes for various text styles: italic, strike, conceal, faint.
```
rhagenson: Would you do me a favor and add the codes you mean here? I understand some from looking them up, but others I am not seeing in reference tables.

text/0000-add-ansi-escape-codes.md
rhagenson: I would like to also say that I am not against having a default return of an empty string, but would prefer making new functionality error avoidant if possible.

text/0000-add-ansi-escape-codes.md
rhagenson: I do not understand this under the "Alternatives" section.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,63 @@
...
+The removal of the existing color functions and the change to the `clear` and `erase` functions (if the default of zero is used) are breaking changes.
+
+# Alternatives
+
+## Alternative to the rfc as a whole
+These escape codes can simply be printed independently of whatever is supported by functions on `ANSI`.  
+Having them available on `ANSI` simply makes it so that users do not need to look the codes up themselves.
+
+## Alternative to the breaking changes
+The existing color functions could remain and the 3-bit and 4-bit color functions could exist alongside them or not be implemented.
```
rhagenson: This is my current preference, but would like more modularity with my aforementioned `Color` or  `FG`/`BG` primitives. I do not want to have to look up escape codes just to change the color of something when that reduction of friction is already supported right now.

rhagenson (commented):
I left some comments mostly about use of default values, additional functionality, preventing errors, and my opinion on the direction I would prefer.

text/0000-add-ansi-escape-codes.md
csos95: I'm going with Sean's recommendation and rewriting this RFC to only be additions and making another later for breaking changes.

text/0000-add-ansi-escape-codes.md
csos95: My reason for wanting to have them split is so that they match the functions for 8-bit and 24-bit colors.  
The plan right now is to return an empty string on invalid values which gives the same effect (of doing nothing) as if those values were printed to the console.

text/0000-add-ansi-escape-codes.md
csos95: Do you think it would be better to have them listed after each code is mentioned like this:

- Add escape code for hide cursor `\x1B[?25l` and show cursor `\x1B[?25h`.

Or just have the table row from the escape codes page like this:

Code | Abbr | Name | Effect
-- | -- | -- | --
CSI ? 25 h | DECTCEM |   | Shows the cursor, from the VT220.
CSI ? 25 l | DECTCEM |   | Hides the cursor.

text/0000-add-ansi-escape-codes.md
rhagenson: Okay! I saw that conversation and thank you for clarifying. I presume this decision means that you do not intend to remove the existing color functions in this RFC (which I was against anyway, but will voice that on the new RFC when it is ready).

text/0000-add-ansi-escape-codes.md
rhagenson: I am against such use of default returns -- or "silent failure" -- when it is possible to make the failure loud and thereby avoid errors reaching runtime. I understand the desire to keep the functions consistent though and appreciate that.

text/0000-add-ansi-escape-codes.md
csos95: There's three general parts to why I wanted to have it return an empty string when invalid values passed as opposed to making the functions partial or using the type system.

1. All of the current ansi functions return a string and are straight forward to combine to make an output string. Making the functions partial would add extra code at the call site.
2. I'm not aware of any way to model types smaller than 8 bits in pony besides doing something like having primitives for each value and then combine them in a type union and that would also add more code at the call site (also, my understanding is that large type unions have a large impact on compile times right now).
3. Having invalid parameters in escape codes isn't really an error that's going to cause issues other than it's not going to do anything, so an empty string would have the same effect.

text/0000-add-ansi-escape-codes.md
rhagenson: I just realized that my statement there contradicts another statement I made here where I say I am **not** against default returns. I want to clarify that I am against silent failure when it is possible to make that failure loud, but not against a non-functional default return such as an empty string. If we can make these additions help the user caught mistakes prior to runtime then I want that, if that is going to break existing functionality and an empty return maintains current stability then I am okay with such empty returns. Does that make sense, or have I talked past myself in my attempts to clarify where I stand on this topic?

text/0000-add-ansi-escape-codes.md
rhagenson: Good question. I would list them inline, personally, as I do not think it breaks up the flow of the bullet points.

text/0000-add-ansi-escape-codes.md
rhagenson: I agree with all three points and each is reasonable. I would then ask that the docstring for these methods list the valid values for each function. If we cannot have the compiler tell the user the value is invalid, then we need to have the valid values available to a user (who does not want to mine through the source to find out why their text is still white/black/gray/whatever).

text/0000-add-ansi-escape-codes.md
csos95: Given the discussion here and the splitting of this RFC into two, I'm thinking that in this one I could leave out the 3-bit and 4-bit color functions and then that leaves the only two functions that could have invalid values passed being the ones for erase in display and erase in line.

The number of valid values for those are small (four for the first and three for the second) and they line up, so making primitives and type unions wouldn't add a ton of types/be a large union like doing one for every color would.

```pony
primitive EraseAfter
primitive EraseBefore
primitive EraseAll
primitive EraseBuffer

type EraseDisplay is (EraseAfter | EraseBefore | EraseAll | EraseBuffer)
type EraseLine is (EraseAfter | EraseBefore | EraseAll)
```

How does that sound?

rhagenson: I am not sure I understand. Are you suggesting dropping `{fg,bg}_{3,4}bit` entirely here (which are currently covered by the existing color functions) which would leave only the erase functions with possible invalid inputs? If so, then I can support that. It is a considerable change to make this RFC about adding the 8 and 24 bit colors, but it does free up the design for this initial additions RFC considerably. We could further debate how we want to handle the overlap of breaking changes with existing functionality in the second RFC.

csos95: Yes, you've got it right.

rhagenson: Perfect. Then as I said, I support that. I have been feeling a little bad that there are additions (which I have no objection to) tied in with existing overlap (which I have been commenting on) here, but my approval is all or nothing. Thank you for splitting this RFC into separate components. Feel free to take as much or as little of my comments here for the RFC's "Alternative Designs" section.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,88 @@
...
+- device status report `\x1B[6n` (sends cursor position to stdin)
+- faint `\x1B[2m`, italic `\x1B[3m`, conceal `\x1B[8m`, and strike `\x1B[9m` text
+- erase in display `\x1B[nJ`
+- erase in line  `\x1B[nK`
+
+
+## Erase In Display/Line
+
+There are currently two similar functions, but they have a couple of issues:
+- Neither of them expose the parameter that the escape codes take.
```
SeanTAllen:
```suggestion
There are currently two similar functions, but they have a couple of issues:

- Neither of them expose the parameter that the escape codes take.  
```

jemc:
As discussed in the sync call, I'd like to request that the RFC include the method signatures for the new methods.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,88 @@
...
+- erase in display `\x1B[nJ`
+- erase in line  `\x1B[nK`
+
+
+## Erase In Display/Line
+
+There are currently two similar functions, but they have a couple of issues:
+- Neither of them expose the parameter that the escape codes take.  
+- The `clear` function moves the cursor to the top left and then clears the screen.  
+This may have been done to create the same result that this escape code gave on DOS.
```
SeanTAllen, line 41:
```suggestion
- The `clear` function moves the cursor to the top left and then clears the screen.  

This may have been done to create the same result that this escape code gave on DOS.
```

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,88 @@
...
+
+Unlike the other added escape codes, these two only have a few valid values
+for the parameters.  
+To enforce the safety of these functions using the type system, four primitives
```
SeanTAllen, line 45:
```suggestion
for the parameters.  

To enforce the safety of these functions using the type system, four primitives
```

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,88 @@
...
+primitive EraseAfter
+primitive EraseBefore
+primitive EraseAll
+primitive EraseBuffer
+
+type EraseDisplay is (EraseAfter | EraseBefore | EraseAll | EraseBuffer)
+type EraseLine is (EraseAfter | EraseBefore | EraseAll)
+```
+
+
```
SeanTAllen:
```suggestion

```

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,88 @@
...
+The added functions would be documented as the existing functions on the `ANSI`
+primitive are and an example of usage would be added to `ANSI`.
+
+# How We Test This
+
+The functions that have parameters would have tests added to ensure that they
+return the correct escape code for the given parameters.
+
+The functions that would have parameters and would be tested are:
+- 8-bit and 24-bit colors for the foreground and background
```
SeanTAllen:
```suggestion
The functions that would have parameters and would be tested are:

- 8-bit and 24-bit colors for the foreground and background
```

SeanTAllen:
During sync we discussed that we'd like to see more detailed information about function signnatures for new methods.

text/0000-add-ansi-escape-codes.md
```diff
@@ -0,0 +1,88 @@
...
+# Motivation
+
+The purpose of these additional methods is to cover more of the standard ANSI
+escape codes within the stdlib.
+
+# Detailed design
+
+## Added Escape Codes
+
+- 8-bit forground `\x1B[38;5;nm` and background `\x1B[48;5;nm` colors
```
ergl:
```suggestion
- 8-bit foreground `\x1B[38;5;nm` and background `\x1B[48;5;nm` colors
```

text/0000-add-ansi-escape-codes.md
csos95: If that extra newline is added, the next line would be split out into a separate paragraph instead of being attached to the bullet point about the functionality of the current clear function.

rhagenson (commented):
Thank you for splitting the modifications into a separate RFC! As well, thank you for adding example codes inline within the added escape codes list -- it is far easier to cross-reference the suggestions here with escape code tables now.

As this RFC stands I see no reason against these additions, although (as discussed on sync by Joe) I would like to see the API being suggested.

Issue #168, opened 2018-04-30
Implement BigDecimal and BigInteger


sblessing:
Extend the standard library with BigDecimal and BigInteger. I have started to working on this, and I would like to submit this as a pull request soon.

igotfr:
> Extend the standard library with BigDecimal and BigInteger. I have started to working on this, and I would like to submit this as a pull request soon.

BIgDecimal is basically F128 and BigInteger U128 and I128
I prefer F128, U128, I128

ergl:
@cindRoberta BigInteger and BigDecimal are arbitrary precision integers, so they can be as big as you want as long as you have the memory to spare.

igotfr:
> @cindRoberta BigInteger and BigDecimal are arbitrary precision integers, so they can be as big as you want as long as you have the memory to spare.

@ergl it looks the arbitrary bit-width integers from Zig: https://ziglang.org/documentation/0.8.0/#Primitive-Types

it's a good proposal, but no with the names BigDecimal and BigInteger

PR #200, opened 2022-02-28
Introduce new numerics

rhagenson:
[Rendered](https://github.com/rhagenson/rfcs/blob/expand-math/text/0000-expand-math.md)

Partially related to:

+ https://github.com/ponylang/rfcs/issues/168
+ https://ponylang.zulipchat.com/#narrow/stream/192795-contribute-to.20Pony/topic/math.20lib
+ https://ponylang.zulipchat.com/#narrow/stream/192795-contribute-to.20Pony/topic/How.20OO.20to.20go.3F.20.28was.20math.20lib.29
+ https://ponylang.zulipchat.com/#narrow/stream/189985-beginner-help/topic/.E2.9C.94.20BigDecimal

### DIFF
--- a/text/0000-expand-math.md
+++ b/text/0000-expand-math.md
@@ -1,26 +1,23 @@
-- Feature Name: Expand `math`
+- Feature Name: Introduce new numerics -- `Rational`, `BigInt`, `BigFloat`, `Complex`
 - Start Date: 2022-02-28
 - RFC PR:
 - Pony Issue:
 
 # Summary
 
-This RFC proposes expanding the standard library's `math` package and the shape it might take to leverage the full extent of Pony. There was prior discussion on [Zulip](https://ponylang.zulipchat.com/#narrow/stream/192795-contribute-to.20Pony/topic/math.20lib) years ago, but I expect some opinions to have changed.
+This RFC proposes the introduction of new numeric types; in particular the addition of a type representing a fractional number (`Rational`), arbitrary precision integer (`BigInt`), arbitrary precision float (`BigFloat`), and complex number (`Complex`).
 
 # Motivation
 
-Currently, `math` includes limited functionality. This should be expanded to include math types, constants, and present a structure with further expansion in mind. An expanded `math` library will allow a unified mathematics among Pony developers. As was discussed during our [2022-03-01 Sync](https://sync-recordings.ponylang.io/r/2022_03_01.m4a), one core principle for including these types in the stdlib is to have a single canonical implementation of them which allow interoperability of numeric types across the Pony ecosystem. Any numeric types introduced by this RFC **must** existing within the current numerical type hierarchy by being compliant with existing numeric traits.
+The primary motivation for adding these types to the stdlib is to have a single canonical implementation of them which allow interoperability of numeric types across the Pony ecosystem.
 
 # Detailed design
 
-The primary goals of this initial expansion are:
-
-1. restructure the `math` package into distinct subpackages; allowing for separation of concerns over continuing to build a monolithic `math` package
-2. provide common `math` data types; for example, `BigInt`, `Rational`, and `Complex`
+I propose we add the aforementioned numeric types into `builtin` so they exist alongside the other standard numeric types. These introduced numeric types **must** existing within the current numeric type hierarchy by being compliant with existing numeric traits.
 
 ## Numeric Hierarchy
 
-Current the Pony numerics hierarchy is as follows.
+Current the Pony numeric type hierarchy is as follows:
 
 ```mermaid
 classDiagram
@@ -63,7 +60,7 @@ SignedInteger <-- ISize
 SignedInteger <-- ILong
 ```
 
-This RFC introduces a few more numeric types: `Rational`, `Complex`, and arbitrary precision `BigInt`. This fit into the hiearchy in the following manner.
+This RFC introduces four more numeric types: `Rational`, `BigInt`, `BigFloat`, and `Complex`. These fit into the numeric type hierarchy in the following manner:
 
 ```mermaid
 classDiagram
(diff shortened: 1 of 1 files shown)


text/0000-expand-math.md
```diff
@@ -0,0 +1,103 @@
...
+1. restructure the `math` library into distinct sub-packages
+2. provide common `math` data types
+
+## Structure
+
+I propose a structure of distinct subpackages including the following:
+
++ `math/big`: Arbitrary precision numbers
++ `math/series`: Mathematical series
++ `math/constant`: Mathematical constants
```
jemc: Small naming suggestion - `constants` feels more natural to me than `constant` as the name, in part because it matches the existing precedent of the `collections` package.

```suggestion
+ `math/constants`: Mathematical constants
```

SeanTAllen:
Is there a reason to tie these different packages together? Would there be something we lose by discussing each individually?

text/0000-expand-math.md
```diff
@@ -0,0 +1,103 @@
...
+
+Example series include `Fibonacci` (already exists), `Pascal` (nCk), `Triangular` ({n+1}C{2}), `Square` (n^2), `Pentagonal` ({2n * (2n - 1)} / 2), etc.
+
+Current `math/Fibonacci` would go into this series package.
+
+### `math/constant`
+
+Initial values to include are those with underlying LLVM representations from the [numbers namespace](https://llvm.org/doxygen/namespacellvm_1_1numbers.html).
+
+Once these values exist in `math/constant`, they should be removed from where they are now, which is on `F32` and `F64` of [`Float`](https://github.com/ponylang/ponyc/blob/master/packages/builtin/float.pony).
```
jemc: What's the rationale/motivation for moving them out of the `F32`/`F64` types and into a new package?

Currently the "Motivation" section of this RFC only talks about enhancing/expanding math capabilities, but this section includes taking existing functionality and moving it to a new place, so I think that warrants its own discussion of motivation. It wasn't immediately clear to me what the benefit of this is, so it probably should be stated explicitly.

SeanTAllen: It feels weird to me that "basic datatypes" are in math rather than operations on datatypes. 

A question would also be... is it possible to move F32 etc from builtins or are they required there.

text/0000-expand-math.md
```diff
@@ -0,0 +1,103 @@
...
+# Motivation
+
+Currently, `math` includes limited functionality. This should be expanded to include math types, constants, and present a structure with further expansion in mind. An expanded `math` library will allow a unified mathematics among Pony developers.
+
+# Detailed design
+
+The primary goals of this initial expansion are: 
+
+1. restructure the `math` library into distinct sub-packages
+2. provide common `math` data types
```
SeanTAllen: what does this mean? It feels important to understanding this RFC, but I don't know what it means.

text/0000-expand-math.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Structure
+
+I propose a structure of distinct subpackages including the following:
+
++ `math/big`: Arbitrary precision numbers
```
SeanTAllen: why call this big instead of arbitrary? or arbitrary_precision?

text/0000-expand-math.md
SeanTAllen: why should they be removed?

rhagenson: These do not necessarily need to move, but it is what I am suggesting. For me the reason is that we have previously had new Pony users (and experiences Pony users) be surprised by the availability of these constants on `F32` and `F64`. I can expand the motivation to reflect this reasoning.

text/0000-expand-math.md
rhagenson: This was my attempt to state that there are two parts to this RFC. The first is to "expand" math by the creation of subpackages to organize its functionality, while the second is to "expand" math by filling these subpackages with new functionality.

text/0000-expand-math.md
rhagenson: Mostly because I have seen "big" be used more often than "arbitrary precision" but that could be my experience alone. Not against a different name, but needed to pick one that seemed reasonable in the wider landscape of similar libraries.

text/0000-expand-math.md
rhagenson: Same thing I said to Joe:

> These do not necessarily need to move, but it is what I am suggesting. For me the reason is that we have previously had new Pony users (and experiences Pony users) be surprised by the availability of these constants on F32 and F64. I can expand the motivation to reflect this reasoning.

SeanTAllen: Surprise doesn't seem like a good reason to me. I think what is most important would be usability. I'd like to hear how the move will make these constants more easily usable.

rhagenson:
> Is there a reason to tie these different packages together? Would there be something we lose by discussing each individually?

I expect that you mean to suggest creation of discrete RFCs. As I answered for the "primary goals" comment for clarification, this RFC is meant to "expand" math is two ways. The first is to expand by providing structure, the second by providing functionality. If we want to make this RFC only the former, then I open RFCs for each subpackage that is fine by me. I thought opening with the initial additions as discussed (a long time ago) in Zulip was "better" than breaking it up without any discussion as to how we want to move towards a more expanded math package.

text/0000-expand-math.md
rhagenson: It is a flimsy reason merely "it has surprised people before" I understand that.

What I would like to be able to do is have some guaranteed immutable value. As these are constants I would want to disallow mutations (ref and iso) entirely. I know generic constraints are possible, but have never explored a `#read` type return so not sure it is even possible. Is that possible?

rhagenson: Best I can think of with existing knowledge is to "wrap" a Float and make all the mutating operations do nothing. But that seems like a bad idea even if it is possible.

```pony
class Constant [F: FloatingPoint[A] #read] is F
```

I highly doubt the above works, but that would be my idea of how to achieve this without compiler changes.

SeanTAllen: So I think that `F32.pi()` is completely reasonable when you understand the language.

I want an F32 width of pi. 

What would this look like on Constant?

- `fun pi_f32()`
- `fun pi_f34()`

This is looking like a very large compiler change that I don't understand a good reason for.

So, all primitive values like F32 are already immutable. So I'm not sure what the "these are constant so I would want to disallow mutations". All F32's are already immutable.

text/0000-expand-math.md
SeanTAllen: what is a common math data type?

SeanTAllen:
@rhagenson if these aren't tried to one another, I think that an RFC for each would be good.

I would focus on adding support for things like Complex numbers, Rational numbers, and those because the design issues for each of those is worthy of individual RFCs.

I'm not even sure that Rational numbers should go in `math`. Math to be is more about "algos and processes" like "matrix multiplication" etc and not so much about things like complex or rational numbers. I'd seriously consider having rational numbers be their own top-level package. Ditto complex.

And fairly reasonable argument could be made for Arbitrary precision numbers to go into builtin. I think we would decide not to, but the argument could reasonably be made based on consistency of F32 and friends being there.

text/0000-expand-math.md
rhagenson: With the being on primitives they are already immutable so forget my comments on that.

As for what it would look like on Constant, I saw it as generic so for example `Constant[F64].pi() => F64(3.1415926...)`

SeanTAllen: So if I do...

`Constant[U64].pi()` 

do I get 3?

This interface seems really weird to me. Pi is generally accepted "to be a float" (as far as computers are concerned", but there's an interface now to get it for me for non-floating ones.

Additionally, now, constants which will vary by underlying type are no longer on the underlying type but on this new primitive that groups things based on being a constant. So if we want to be able to get pi as a BigDecimal, we need to not only implement BigDecimal but we need to touch all of the code for constant to add all the appropriate constants. That feels wrong. Adding a new number type like BigDecimal should be relatively self-contained. This spreads it around. I'm not in favor of this without some serious convincing.

rhagenson: I have changed it to say "could" not "should" as considering F32 and F64 are immutable by default (an oversight on my part) my concern for keeping them in place was misplaced. Now it is just this "surprised placement" concern, which is not a good enough reason to remove it.

text/0000-expand-math.md
rhagenson: I added examples to this line to address your question. I was using "math" and "numeric" interchangeably here and meant to imply what is listed below in the RFC, namely `Rational`, `BigInt`, and `Complex` types which are common enough in usage to be part of the standard library.

text/0000-expand-math.md
rhagenson: I went against my own RFC here. The parameter is as the call site via `Constant.pi[A: Float](): A(3.14159...)`.

I changed "should" to "could" to address removal of methods on floats. The reason of "surprising placement" is not a good enough reason to remove them, but having a grouping of constants on a `Constant` primitive still seems reasonable to me.

rhagenson:
I am not against breaking this RFC into pieces, but wanted to present all the initial pieces at once. I am personally in favor of the addition functionality discussed here being outside of builtin since it will rely on builtin, but builtin will not rely on any of the suggested additions.

text/0000-expand-math.md
```diff
@@ -0,0 +1,105 @@
...
++ `math/series`: Mathematical series
++ `math/constants`: Mathematical constants
++ `math/rational`: `Rational` data type and related functions
++ `math/complex`: `Complex` data type and related functions
++ `math/(x,exp,etc)`: experimental additions, utilities, and effective "catch-all" for matters that do not neatly fit into other subpackages
```
SeanTAllen: why not "just at the top level aka `math`" along with others that are already there?

rhagenson:
During [Sync](https://sync-recordings.ponylang.io/r/2022_03_01.m4a) on 2022-03-01 I wrote down the following notes for how to improve this RFC:


- [x] Add implementation details; for example, what does a `Rational` look like internally?
- [x] Add motivation for having canonical implementation for these types; addressing having interoperability of numerics across the Pony ecosystem
- [x] Mention intention to be compliant with existing numeric traits such as Real, Integer, etc
- [x] As an external reference, Haskell was noted as having a strong numerics hierarchy

text/0000-expand-math.md
rhagenson: Not sure I understand what you mean, but I suspect you mean dropping the "math" prefix so we have `constants`, `rational`, `complex`, and etc. For that my biggest argument against is that `constants` could be seen as not being "mathematical constants" but rather some way to produce globally immutable "constants" so could be confusing based on that. The alternative for me if we are reducing the subpackaging is that some of these suggestions don't need a package necessary -- a `Constant` primitive can be added to `math` directly without issue.

This gets to the "why" for me including all the piece in one RFC so we can discuss how we want to develop such expansions in the stdlib.

ergl: I understood Sean's comment in the context of `math/(x,exp,etc)`: instead of introducing a catch-all subpackage of `math`, put those types and functions at the root of the `math` package.

rhagenson: I see no issue with `math` itself being the catch-all. If that is the direction we agree is better then I will change the RFC to reflect this decision. I suggested the catch-all subpackage so we are forced to have the conversation about where code that doesn't neatly fit somewhere goes.

rhagenson:
@SeanTAllen Would me switching this to a draft PR prevent the discuss during sync label from being reapplied? I have doubts that I will get all the updates done before next Sync in time for folks to review the changes. I have no issue with more reviews/comments/discussion, but want to prevent issues of this RFC continually reappearing for Sync with only small or in-progress edits.

My current plan is:
- [x] Add top-down mermaid graph for the numeric type hierarchy as we have it today
- [x] Add top-down mermaid graph for the numeric type hierarchy with these new types included
- [ ] Fix the typing errors for at least `Rational` (I have a comment in the example implementation in the most recent commit)

SeanTAllen:
> Would me switching this to a draft PR prevent the discuss during sync label from being reapplied? 

No. 

See https://github.com/ponylang/rfcs/blob/main/.github/workflows/add-discuss-during-sync.yml#L3 for events that cause the label to be added.

rhagenson:
- [x] Add BigFloat numeric to mermaid chart
- [x] Add Complex parameterized on Real
- [x] Add Rational parameterized on Integer
- [x] Discuss breaking this RFC into smaller dedicated RFCs concerning
  - Adding numeric types to builtin
  - Introducing new math functionality as suggested in this RFC (e.g., having a Constant primitive)

rhagenson:
On [Sync 2022-03-08](https://sync-recordings.ponylang.io/r/2022_03_08.m4a), we discussed breaking this RFC into smaller dedicated RFCs with this RFC/PR specifically becoming a "Introduce new numerics; Rational, BigInt, BigFloat, Complex" and splitting the other content off to better allow us to discuss the merits of each addition/change.

SeanTAllen:
I'm in favor of breaking this apart into different RFCs.

rhagenson:
Apologies about temporarily closing this -- I tried changing the name of the `expand-math` branch to `add-numerics` and I guess that is not a thing I can do without forcing this PR to close.

text/0000-expand-math.md
```diff
@@ -1,26 +1,23 @@
...
+The primary motivation for adding these types to the stdlib is to have a single canonical implementation of them which allow interoperability of numeric types across the Pony ecosystem.
 
 # Detailed design
 
-The primary goals of this initial expansion are:
-
-1. restructure the `math` package into distinct subpackages; allowing for separation of concerns over continuing to build a monolithic `math` package
-2. provide common `math` data types; for example, `BigInt`, `Rational`, and `Complex`
+I propose we add the aforementioned numeric types into `builtin` so they exist alongside the other standard numeric types. These introduced numeric types **must** existing within the current numeric type hierarchy by being compliant with existing numeric traits.
```
jemc, line 16: For me, the preference would be to keep `builtin` as minimal as possible, so I'd prefer to put these new types in a new package (or series of packages), unless there is a strong motivation for it to go in `builtin`.

Pretty much all of the existing public types in `builtin` have hard requirements forcing them to be there, with a reason like one of the following:
- they are used by core language constructs (`None`, string literals, numeric literals, etc)
- they need to use raw pointers (`Array`, `String`, etc)
- they use `compile_intrinsic` for one or more function definitions
- they are part of the `Env`, which is passed to the `Main` actor on entry
- they are used by one of the types that meet the above criteria

I don't think these new types have any hard requirements forcing them to go in `builtin`, so I believe they shouldn't be there.

rhagenson: We can discuss this further on Sync and I will update the RFC according to our conversation. I have no particular need for these to be added to `builtin` so no objection to an agreed upon other location such as a newly created `numerics` or adding these to `math` (as was the RFC state prior to my latest changes).

rhagenson:
Notes from [Sync 2022-03-15](https://sync-recordings.ponylang.io/r/2022_03_15.m4a):

- [ ] Update Unresolved Questions w/ details -- this RFC is the high-level overview of how the types fit into the numeric hierarchy, but discrete RFC for each numeric is needed providing complete implementation details
- [ ] Move suggestion out of `builtin` into new `numerics` package

text/0000-expand-math.md
```diff
@@ -0,0 +1,230 @@
...
+## Methods of Concern
+
+```pony
+trait val Real[A: Real[A] val] is
+  (Stringable & _ArithmeticConvertible & Comparable[A])
+  ...
+  new val min_value()
+  new val max_value()
+  ...
+```
```
jasoncarr0, line 119: I understand that this has already been present, but given that this is re-design I'm not clear on the benefit of these constructors, but there's some obvious costs. Is there a use case for max_value() / min_value() that makes sense for all numbers? Given that this doesn't work for BigInt/BigFloat/similar types I'd be hesitant to place it at the top of the hierarchy.

Can we instead make a Bounded interface? This will only break code which relies on these methods being present in any Real[A]

text/0000-expand-math.md
```diff
@@ -0,0 +1,230 @@
...
+
+`Real` will exist above `Rational`, `BigInt`, and `BigFloat` and as such would require defining the above methods for these types. `Rational` can be defined as minimum and maximum of the numerator, however `BigInt` and `BigFloat` by definition have arbitrary precision making defining a minimum and maximum difficult at the least -- if we define them as the minimum and maximum of a machine-sized int and float, or define them as -Inf and Inf -- or impossible at the worst -- if we define them by their possible limits which are arbitrary.
+
+```pony
+trait val Integer[A: Integer[A] val] is Real[A]
+  ...
+  fun op_and(y: A): A => this and y
+  fun op_or(y: A): A => this or y
+  fun op_xor(y: A): A => this xor y
+  fun op_not(): A => not this
```
jasoncarr0, line 129: This can't be implemented by a BigInt, so it's odd that a BigInt can't actually be an Integer

SeanTAllen: Sorry, I don't understand your comment. Can you try explaining in a different way?

text/0000-expand-math.md
SeanTAllen: I don't understand what you are suggesting. What's a "bounded interface"?

text/0000-expand-math.md
jemc: You can't do a bitwise "not" operation on a `BigInt` - at least not by the normal semantics of a bitwise "not" operation.

You can't because the you'd expect that any bits "more significant" than the most significant bit of the value would be set to `1`.

But a `BigInt` has no upper bound, and thus there is no limit to the number of leading `1` bits implied by inverting its bits

text/0000-expand-math.md
```diff
@@ -0,0 +1,230 @@
...
+
+I propose we add the aforementioned numeric types into `builtin` so they exist alongside the other standard numeric types. These introduced numeric types **must** existing within the current numeric type hierarchy by being compliant with existing numeric traits.
+
+## Numeric Hierarchy
+
+Current the Pony numeric type hierarchy is as follows:
+
+```mermaid
+classDiagram
+class Any
```
SeanTAllen, line 24: At the moment, the mermaid diagrams are distracting. Is there a way to see them rendered? If not, then as part of the RFC text, I would prefer something else as a way to express whatever I am supposed to be getting from this.

text/0000-expand-math.md
jemc: I think he means a new interface whose name is `Bounded`, which has the min and max value methods on it.

text/0000-expand-math.md
SeanTAllen: I still don't understand @jemc.

> "This can't be implemented by a BigInt, so it's odd that a BigInt can't actually be an Integer"

Can you explain how what you said is related to Jason's comment?

I don't understand why not being able to be implemented BigInt makes it odd that BigInt can't actually be an Integer.
I don't understand what is being said here.

text/0000-expand-math.md
```diff
@@ -0,0 +1,230 @@
...
+  fun bit_reverse(): A
+    """
+    Reverse the order of the bits within the integer.
+    For example, 0b11101101 (237) would return 0b10110111 (183).
+    """
+
+  fun bswap(): A
+```
+
+`Integer` will exist above `BigInt` and as such would require defining the above methods -- however `BigInt` will be defined via other numerics so these methods could be applied recursively.
```
SeanTAllen, line 140: I don't understand what this means. What are "the above methods"? Those listed on the Integer trait? 

"`BigInt` will be defined via other numerics so these methods could be applied recursively" <-- I don't understand what this means either. There seems to be something that I am supposed to understand about "other numerics" that BigInt will be defined via (did I miss that somewhere, if yes, bringing information forward so that you don't have to hold large chunks of RFC in your head would be good). I'm also not sure what "applied recursively" means in this context.

text/0000-expand-math.md
```diff
@@ -0,0 +1,230 @@
...
+  fun asin(): A
+  fun atan(): A
+  fun atan2(y: A): A
+
+  fun acosh(): A
+  fun asinh(): A
+  fun atanh(): A
+```
+
+`FloatingPoint` will exist above `BigFloat` and as such would require defining the above methods -- many of which are ill-defined under arbitrary precision, or are functions using C-FFI and/or LLVM intrinsics.
```
SeanTAllen, line 204: I'm confused. Why would we do it this way if it has these problems? This seems like an argument against doing it this way. I feel like there's some meaning here that I am missing.

SeanTAllen:
I find this RFC very confusing as written. I don't understand what large sections are attempting to convey. I think I'm going to need someone to walk me through it as at the moment, I'm not really sure how to even phrase my "I don't understand" statements.

text/0000-expand-math.md
SeanTAllen: @jasoncarr0 is that what you meant? re: `Bounded` interface? if yes, did you specifically mean structural typing or were you using "interface" more loosely?

text/0000-expand-math.md
jemc: I'll copy Jason's words and insert parentheticals with my own commentary:

This (the `op_not` function) can't be implemented by a `BigInt` (for the reason mentioned in my previous comment)

So it's odd that a `BigInt` (which is conceptually a kind of integer) can't actually be an `Integer` (it cannot be a subtype of the `Integer` trait, because the `Integer` trait includes the `op_not` function).

SeanTAllen: Thanks @jemc. I understand now.

SeanTAllen:
I'd like to see this broken apart further into an RFC for where Rational and Complex fit with the existing numerics hierarchy and a separate discussion of the best way to implement "Big" numerics. There are rather complicated issues that appear to apply to only "the bigs" that I think will take some time to resolve and during that time we can probably agree to how Rational and Complex will fit in with the existing, do detailed RFCS for each and implement them before the questions around "the bigs" are resolved.

text/0000-expand-math.md
jasoncarr0: I meant a literal Pony interface, that is:
These constructors would be part of an interface named `Bounded`, while the other methods would be part of this trait.

jasoncarr0: Actually there is also an argument for removing Ordered, but only because this trait is really the lowest possible. Mostly that the implementations for things like matrices would be sketch but making them numerics makes sense and is powerful.

adrianboyko:
Any thoughts on the possibility of having builtin complex primitive types? I think the RFC is proposing that all complex values will be object instances, which seems like it would be troublesome in various use-cases. For instance, I'm working on signal processing in Pony and have an actor that produces a stream of `Vector[U8C]`s from RTL-SDR hardware. I'm currently defining `type U8C is U16`, which gives me a complex type that matches C's representation of the same, so I can use C libraries that do vector processing on complex values. 

C11 standard says:

> Each complex type has the same representation and alignment requirements as an array type containing exactly two elements of the corresponding real type; the first element is equal to the real part, and the second element to the imaginary part, of the complex number.

libvolk (C library for vector operations) defines the following:

```C
typedef char complex lv_8sc_t
typedef short complex lv_16sc_t
typedef long complex lv_32sc_t
typedef long long complex lv_64sc_t
typedef float complex lv_32fc_t
typedef double complex lv_64fc_t
```

In Pony, I imagine types like:

```
primitive I8C is SignedIntegerComplex[I8C]
primitive I16C is SignedIntegerComplex[I16C]
primitive I32C is SignedIntegerComplex[I32C]
primitive I64C is SignedIntegerComplex[I64C]
primitive I128C is SignedIntegerComplex[I128C]
primitive U8C is UnsignedIntegerComplex[U8C]
primitive U16C is UnsignedIntegerComplex[U16C]
primitive U32C is UnsignedIntegerComplex[U32C]
primitive U64C is UnsignedIntegerComplex[U64C]
primitive U128C is UnsignedIntegerComplex[U128C]
primitive F32C is FloatingPointComplex[F32C]
primitive F64C is FloatingPointComplex[F64C]
```

rhagenson:
@adrianboyko Since I am the author on this RFC, I will chime in.

> Any thoughts on the possibility of having builtin complex primitive types?

I am not against it outright because I do not know what all that would entail. However, this RFC will be refined (when I have time) to propose the higher level aspects of what numerics we wish to introduce. Which will then be followed by specific RFCs for the structure and implementation of those specific numerics. That is to say, how complex numerics are implemented is critical to their successful adoption and definitely needs to be a part of the RFC on their implementation.

jemc:
The idea of primitive complex types was discussed a bit on the sync call.

The advantage of not having an extra object allocation is a great one.

To make that idea work, they would need to be in the `builtin` package and would likely need many of their methods to be implemented as `compiler_intrinsic`. This would be possible, but in my opinion it's not ideal.

In my opinion, this would be a good use case for struct-like value types, which I have discussed a bit in the past (and implemented in the Savi compiler for prototyping). The basic premise is that it would be a new kind of Pony type declaration that is declared just like a `class`, but can only have `let` fields (no `var` fields) - and at runtime it behaves like a tuple (no object allocation, unless it gets "boxed" into an abstract type).

Having a "struct-like value type" like this would allow for avoiding the extra object allocation for the complex numeric types, as well as anything similar that others wanted to build in user-defined libraries. It also would solve the "newtype" use case discussed in #65. It would also help use cases where performance-sensitive code is currently using tuples for performance reasons where they would otherwise prefer to use classes.

PR #201, opened 2022-05-29
Introduce empty Ranges

stefandd:


text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,255 @@
...
+
+# Drawbacks
+
+* As far as I can tell from public code, Range.is_infinite() is not used in any Pony code other than test code associated with tests for Range.
+* This should not produce any bugs in the compiler since it is a small library change only
+* Maintenance cost should be marginal.
+
+# Alternatives
+
+The impact of not doing this are user code bugs in attempted uses of `Range` as shown in the motivation. At the very least the impact of loops over unintendedly infinite `Range` iterators should be minimized by either encouraging more use of `Range.is_infinite()` or by improved tutorials.
```
SeanTAllen: Your belief that there will be user code bugs is editorialzing at this point. Alternatives should be presented with the best possible case for those alternatives such as "Leaving it as is means that code that might exist that I am unaware from my search on GitHub would continue to work".

Suggesting documentation or tutorial improvements if left as is is certainly reasonable to consider in this section.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,255 @@
...
+
+# Unresolved questions
+
+What parts of the design are still TBD?
```
SeanTAllen: This should be completed. You left the template value here.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,255 @@
...
+
+This can be taught by improving the documentation of `Range` and by discussing a few examples in the tutorial section. Also, as stated above, `Range.is_infinite()` is hardly used on Github and the impact of this change should be considered mild.
+
+Tests can easily be added to `packages/collections/_test.pony` that would add a `_assert_empty` method and test for relevant cases while removing thoses cases from the current `_assert_infinite` tests.
+
+# Drawbacks
+
+* As far as I can tell from public code, Range.is_infinite() is not used in any Pony code other than test code associated with tests for Range.
+* This should not produce any bugs in the compiler since it is a small library change only
+* Maintenance cost should be marginal.
```
SeanTAllen: If you are going to make a statement about maintenance cost, it should have at least some justification. Given this is in drawbacks, it sounds like you believe there is an additional maintenance cost vs the status-quo. If that is the case, it should be detailed here. If there isn't an additional maintenance cost, that should be discussed elsewhere, not in drawback.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,255 @@
...
+
+# How We Teach This
+
+This can be taught by improving the documentation of `Range` and by discussing a few examples in the tutorial section. Also, as stated above, `Range.is_infinite()` is hardly used on Github and the impact of this change should be considered mild.
+
+Tests can easily be added to `packages/collections/_test.pony` that would add a `_assert_empty` method and test for relevant cases while removing thoses cases from the current `_assert_infinite` tests.
+
+# Drawbacks
+
+* As far as I can tell from public code, Range.is_infinite() is not used in any Pony code other than test code associated with tests for Range.
```
SeanTAllen: This isn't a drawback, it is further justification for your change. The point of an RFC is to attempt to present an even-handed argument. This isn't even handed. Saying that it is a breaking change is a reasonable drawback. Stating what you can see as the extent of that breaking change as part of discussing it, but this should be rewritten.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,255 @@
...
+* This should not produce any bugs in the compiler since it is a small library change only
```
SeanTAllen: If it doesn't touch the compiler then there's no point to stating that it won't have bugs in the compiler. Also this isn't a drawback, you are using the drawback section to further your argument.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,255 @@
...
+Range(nan, nan, -inf): .is_empty() = false, .is_infinite() = true
+```
+
+# How We Teach This
+
+This can be taught by improving the documentation of `Range` and by discussing a few examples in the tutorial section. Also, as stated above, `Range.is_infinite()` is hardly used on Github and the impact of this change should be considered mild.
+
+Tests can easily be added to `packages/collections/_test.pony` that would add a `_assert_empty` method and test for relevant cases while removing thoses cases from the current `_assert_infinite` tests.
+
+# Drawbacks
```
SeanTAllen, line 410: None of these drawbacks feel like you've given thought to why not to make the change. They are all further arguments that you are making for the change.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
+Range(nan, nan, 1): .is_empty() = false, .is_infinite() = true
+Range(nan, nan, inf): .is_empty() = false, .is_infinite() = true
+Range(nan, nan, -inf): .is_empty() = false, .is_infinite() = true
+```
+
+# How We Teach This
+
+This can be taught by improving the documentation of `Range` and by discussing a few examples in the tutorial section. Also, as stated above, `Range.is_infinite()` is hardly used on Github and the impact of this change should be considered mild.
+
+Tests can easily be added to `packages/collections/_test.pony` that would add a `_assert_empty` method and test for relevant cases while removing thoses cases from the current `_assert_infinite` tests.
```
SeanTAllen: please remove the `easily`. Its a meaningless word n this content.

Notes about testing are really about testing not about teaching it.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
+Range(-inf, -inf, -inf): .is_empty() = false, .is_infinite() = true
+Range(nan, nan, 0): .is_empty() = false, .is_infinite() = true
+Range(nan, nan, 1): .is_empty() = false, .is_infinite() = true
+Range(nan, nan, inf): .is_empty() = false, .is_infinite() = true
+Range(nan, nan, -inf): .is_empty() = false, .is_infinite() = true
+```
+
+# How We Teach This
+
+This can be taught by improving the documentation of `Range` and by discussing a few examples in the tutorial section. Also, as stated above, `Range.is_infinite()` is hardly used on Github and the impact of this change should be considered mild.
```
SeanTAllen: "Also, as stated above, `Range.is_infinite()` is hardly used on Github and the impact of this change should be considered mild."

has no place in a "how we teach this section". Its not about teaching.

Please elaborate on what should be in the tutorial. An RFC should be a document where if accepted anyone can pick it up and implement it. Further, it should be a document that serves as the guide for when to accept a change. "A few examples in the tutorial" isn't something most people could use as the basis for accepting a PR that implements tutorial changes for this RFC.

SeanTAllen: Please detail how the documentation would be changed in as much detail as makes sense.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
+    _empty = no_progress
+          or ((_min == _max) and is_bounds_finite)
+    _infinite = not _empty
+            and ((_inc == 0)
+                or not is_bounds_finite
+                or not is_step_finite)
+    if _empty then _idx = _max else _idx = _min end // has_next() will return false without additional code
+```
+
+With a so-modified Range class and a new .is_emtpy() function, here is a list of Range examples all of which were previously infinite, and their new classification:
```
SeanTAllen: It seems like everything below should also be included in the "how we test" section.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
+In this RFC, this expression is used to define one portion of the now *empty* Ranges in accordance with the natural expectation and the implementation of Range in other languages. Using this test, `Range(0, 10, -1)` is *empty* and no longer *infinite*, but so is in fact `Range(0, inf, -1)`, even though any finite steps would mathematically never produce progress towards an infinite bound. While one can generally never produce a finite list of elements that would incrementally move from 0 to +Inf, we can know in the case of `Range(0, inf, -1)` that the list of elements must be empty because of the sign of the step parameter.
```
SeanTAllen: please remove the word "natural"

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
...ases in which the current `Range` implementation returns infinite iterators, this particular case seems illogical since the iterator should produce a number of elements of `step` spacing on the *trajectory from min to max*. In the example above, no such elements can exists due to the sign mismatch between min, max, and step, and the returned Range should be empty rather than infinite -- like it is in Rust, Python, and others. The only reason why Pony creates an *infinite* range in this situation is that it, algorithmically, seems to be saying: I know there are no elements (the implementation actually checks for this "no progress possible" situation), but I will walk indefinitely in the wrong direction because you told me so and hence I am infinite, even though the number of elements on the expected tranjectory is zero. The big issue with this, beside the creation of unwanted infinite loops, is that it also precludes or makes very cumbersome the expressive use of the bounds like in the csv file example above in which the programmer *knows* and *makes use of* the fact that if the bounds (and the step) are not compatible sign-wise, the range will have zero iterations, i.e. be *empty*.
```
SeanTAllen: please remove the world "natural".

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
+- Feature Name: Introduction of empty Ranges
+- Start Date: 2022-05-29
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+The RFC carves out a subset of cases that are currently defined as *infinite* by proposing to instead produce an *empty* `Range` in these cases. This makes `Range` behave more similar to the way that Python's range() and Rust's Range() behave, allowing one to write more natural code. The proposal also addresses (fixes) a corner case that is currently unecessarily producing *infinite* Ranges.
```
SeanTAllen: I'm not comfortable with the arguments in here that are "we are changing this because other languages do it this way".

Accepted RFCs set a precedent for making a change. "Other languages do it" isn't a good precedent.

I'd like to see things like "natural expectation" and "because language x does it" removed from the RFC and for the focus to be on Pony and why this change is good for Pony code and why those advantages are worth a breaking change.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
+  let lines = Range(1, f.size()) 
+  if not lines.is_infinite() then
+    for idx in lines do // start at 1 to skip header line
+      ... // perform only if f is not empty and has at least 1 data line
+    end
+  end
+  ```
+  
+  it is fair to mention that currently no `Pony`-labeled code on Github other than test code for Range exists that would make use of .is_infinite(). Also, the above is unnecessarily cumbersome, and again prevents the expressive use of the Range bounds with calculated values.
+
```
SeanTAllen: I think it would help to show before and after code and then explain why you think the after is more expressive. A statement like "more expressive" doesn't really mean much. Everyone has their own idea of what "expressive" might mean.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,267 @@
...
+- Secondly, case 1) above indiscriminately makes a `Range` infinite when *any* of the arguments is a floating point NaN or +-Inf. It does so on the grounds that in all such cases no quantifyable progress could be made from `min` to `max` (quote from range.pony):
+  
+  > If any of the arguments contains `NaN`, `+Inf` or `-Inf` the range is considered infinite as operations on any of them won't move `min` towards `max`.
+  
+  However, this unecessarily renders ranges infinite in which `min == max`. This proposal argues that `Range[F64](5, 5, nan)` or `Range[F64](5, 5, inf)` should be empty rather than infinite since no single iteration is necessary to advance from `min` to `max` regardless of the magnitude of `step`.
```
SeanTAllen: please removed editorializing like "unecessarily"

SeanTAllen:
I see in one of your commits since I first left comments that you added the following:

> "It is argued here that based on using the name Range, it is expected to return an iterator producing a sequence of values that lie between the stated bounds. In the cases in question, Pony creates an iterator that produces a sequence of values that lie outside the provided bounds min and max."

I think that statement should be the core of your argument. That's an excellent point that is "entirely about Pony" that would justify a breaking change. Given you didnt find much code using this, you can't make a good "its easy to create bugs" based on empirical evidence But, you should be able to show code that the average reader of the RFC would expect to do one thing and you can show that it doesnt do that.

If you can demonstrate that, and show that there aren't surprising corner cases with the changes you propose, then I think it would be a change I could support wholeheartedly.

SeanTAllen:
Given that performance is immediately after correctness in the Pony principles, it would be good to note if there are performance implications of the RFC changes.

stefandd:
This needs some more preparation time. I will reopen when I have a significantly improved version

stefandd:
The draft was completely reworked, incorporating as much as possible the comments that were previously given.

stefandd:
This has been re-opened and is ready for more comments.

rhagenson:
@stefandd I have added myself as a reviewer to remind myself to look at it this evening. I think Sean has been busy lately so I want to ensure you have eyes on it in case he is still busy.

SeanTAllen:
I am very busy at work and have had limited time. I apologize for the slowness with which any review might come.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
+- Feature Name: Introduction of empty Ranges
+- Start Date: 2022-05-29
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+The RFC addresses the fact that for certain parameters, Pony currently creates infinite Range iterators that produce infinite numbers of values that lie *outside* the range of values `min` and `max` provided to `Range`. For example the Range iterator `Range(10, 0)` produces the infinite sequence `10, 11, 12, ..` on subsequent calls to `.next()` where all but the first element do not lie within the range 10..0. However, it is argued here that an iterator using the name `Range` can be expected to produce a sequence of values that lie within the `min, max` bounds. This expectation can be met by introducing and in such and similar cases returning *empty* ranges. One unexpected side effect of Pony's current behavior is that even though finite bounds are given, an *infinite* iterator can be created from them which, when used without consideration for this side effect, can produce indefinitely-running actors and "hanging" programs.
```
rhagenson: > This expectation can be met by introducing and in such and similar cases...

Some words have been omitted here and I am not sure what it is meant to say.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+The RFC proposes a delineation of the so-far *infinite* cases into one of two types: those at are now *empty*, and those which remain *infinite* even under this RFC.
```
rhagenson: > ...those at are now *empty*...

I presume this is meant to read "...those that are now *empty*..."

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
...ues of step (note that if either of the `min` or `max` parameters is `NaN` such expressions are always `false` and this special case is discussed further below). Here, besides the obvious sign-incompatibility of the three Range parameters, the cases of equal bounds `min == max` and `step == 0` are included, since the former wouldn't satisfy `[min, min)` (and also doesn't require any progress) and the latter prevents any progress. In the current implementation of Range, a somewhat similar couple of expressions are used with the comment `// no progress` and `// progress into other directions`, but they are treated as a sufficient condition for *infinite*. Here, this criterion is used as part of a test for empty Ranges because it allows us to know in which cases the generation of values between min and max is impossible, irrespective of the magnitude of the bounds or steps, in other words, no valid *trajectory* from min to max does exists with the provided parameters. The second important criterion used here is that for a Range iterators to be not empty, it must only produce finite values even when infinite bounds are specified, and if that is not possible, the Range is again *empty*.
```
rhagenson: I find this paragraph tremendously difficult to follow. For example, you state "two criteria" yet by the time I got to the second criteria I was not clear what in the bulk of the writing the *first* criteria was. Then in stating the second criteria you restate the premise of it being a criteria for the Range to not be empty -- which is an unnecessary restatement -- then state the second criteria and continue the sentence with further conditions of when the criteria does not apply.

I am not sure if this advice will be helpful, but in the past I have asked people to focus on a "topic and stress" approach to rewriting in situations such as this. That is, the start of each sentence introduces the topic at hand which you then connect to the stress at the end of each sentence. For example:

> The second important criterion used here is that for a Range iterators to be not empty, it must only produce finite values even when infinite bounds are specified, and if that is not possible, the Range is again *empty*.

Might become:

> The second criterion is that any non-empty Range must produce finite values -- making these Ranges distinct from both "empty" and "finite" Ranges.

The topic is the second criteria, it is stated without repetition, then *why* it is important (as a criteria) ends the sentence.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+
+Combining these two criteria, for a `Range` to not be *empty*, progress from `min` to `max` must be possible (no-progess is `false`), and the iterator that realizes this progress must produce *finite* values that lie within `[min, max)`.
+
+Using this test, `Range(0, 10, -1)` is *empty* and no longer *infinite*, but so is in fact `Range(0, inf, -1)` that is also considered *infinite* currently. While one can generally never produce a complete list of elements that would incrementally move from say 0 to +Inf, we can know in the case of `Range(0, inf, -1)` that the Range must be *empty* because of the incompatible sign of the step parameter relative to the signs of `min` and `max` (progress impossible).
```
rhagenson: What test?

rhagenson: Using your criteria, what is the argument *against* transforming some of these ranges *into* finite ranges? For example, `Range(0, 10, -1)` could be made to produce: 1) progress from `min` to `max` and 2) values that lie within `[min, max)` by arranging to always enforce that `min <= max` or changing the sign of `step` or by iterating backwards. What is the distinct benefit of having this be your proposed "empty" rather than enforcing any of the above conditions?

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+Another case in which a Range is now *empty* is when the bound equality `min == max` can be evaluated meaningfully, independent of whether `step` equals e.g. `0` or `+-Inf` such as for example `Range(10, 10, 0)`. In the case of equal and finite bounds, no value can exist that satisfies `[min, min)`. In case of infinite bound values, the range is also empty because one `inf` cannot be meaningfully tested for equality to another `inf` and even if it could, the second criterion would be violated in that a sequence of values would contain non-finite values.
```
rhagenson: I do not know what it is that you are trying to say here. There is no value where `min == max` that satisfies `[min, max)` -- I agree that far. But then you state that this allow applies to infinite bound values, which is already addressed by the fact that there is no value where `min == max` that satisfies `[min, max)`. However, you then call to the second criterion as if `inf` could satisfy this condition only to state that it would violate your rule against non-finite values in (non-empty) ranges. This final point is not part of the `min == max` condition you are stating will be empty.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+
+As already stated, the current `Range` implementation treats the occurrence of `+-Inf` or `NaN` parameters in *any* of the `min, max, step` parameters as sufficient condition for *infinite*. This RFC considers the opposite: the no-progress expression cannot be meaningfully evaluated if any one of the 3 tested parameters is a `NaN` value, and one can also not decide whether any .next() iterations which add NaN lie within the given numerical range, or, if `min` or `max` are `NaN`, what that range is in the first place. Any occurrence of `NaN` in the Range parameters therefore produces an *empty* Range. Similarly, while a `step` argument of `+-Inf` may "pass" the inspection by the *no-progress criterion*, no finite points within the mathematical range `[min, max)` can be computed, including when `min, max` themselves are `inf`. Therefore, such Ranges are *empty*, too. A `step` argument of `0` satisfies the *no-progress criterion* and produces an *empty* Range.
```
rhagenson: You jump from the current implementation to your proposed definitions in this paragraph without clear distinction of what currently is and what will be. The purpose here seems to be to discuss how `min`, `max`, and/or `step` being `+Inf`, `-Inf` or `NaN` will now produce empty ranges, but that is not true under your own rules. `Range(0, Inf, 1)` is not empty.

You also restate what you have already stated about the situation when `min == max` producing an empty range -- "...no finite points within the mathematical range `[min, max)` can be computed, including when `min, max` themselves are `inf`." -- this restatement is distracting from your point here.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+
+The cases that would remain *infinite* under this proposal are then only those which fulfill the following condition: they are *not empty* (as per the evaluation above) AND their upper bound is either `+Inf` or `-Inf`. Typical examples would be `Range[F64](0, inf, 1)` or `Range[F64](0, -inf, -1)`. We can't allow a lower bound that is not finite, because then the elements returned on .next() calls would again not be finite.
```
rhagenson: How would `-Inf` ever be an upper-bound in a non-empty range? (I presume you mean `max` not upper-bound?)

rhagenson: Same sort of point with your use of lower bound in the final sentence -- where I presume you mean `min`.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+
+Summarizing this, the proposed algorithm for the Range classification is outlined in a close-to-code form. The following expressions are useful:
+
+```
+no_progress = ((min <= max) and (step <= 0) or (min >= max) and (step >= 0))
+progress = not no_progress = ((min > max) or (step > 0)) and ((min < max) or (step < 0))
+                           = (((min > max ) or (step > 0)) and (min < max)) or (((min > max ) or (step > 0)) and (step < 0))
+                           = ((step > 0) and (min < max)) or ((min > max ) and (step < 0))
+                           = ((min < max) and (step > 0)) or ((min > max ) and (step < 0))
```
rhagenson, line 163: Why are you defining `progress` in terms of `not no_progress` only to provide additional cases where progress can be made? It seems redundant, and I misread it the first time as an instance of destructive read.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+```
+
+The above discussed combined criterion for *empty* was: either the no_progress criterion is true, or the generated values would not be finite. Finite values require both `min` and `step` to be finite, since the values are generated by an iterative addition of `step` to `min`. Therefore, one can state:
+
+```
+not_empty = progress and min.finite() and step.finite() // .finite() => the value is not `+-Inf` and not `NaN`
+infinite = not_empty and max.infinite()
+empty = not not_empty
```
rhagenson: I feel like these three distinctions are what your entire RFC is about and yet once I got here I did not feel like you made a strong case for why we should have these three distinctions.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,425 @@
...
+Range(10, 10, -1), .is_empty() == true, .is_infinite() == false
+Range(10, 10, 0), .is_empty() == true, .is_infinite() == false
+Range(-10, -10, 1), .is_empty() == true, .is_infinite() == false
+Range(-10, -10, -1), .is_empty() == true, .is_infinite() == false
+Range(-10, -10, 0), .is_empty() == true, .is_infinite() == false
+Range(0, 10, -1).is_empty() == true, .is_infinite() == false
+Range(10, 0, 1).is_empty() == true, .is_infinite() == false
+Range(10, 10, 1).is_empty() == true, .is_infinite() == false
+Range(10, 10, -1).is_empty() == true, .is_infinite() == false
+Range(10, 10, 0).is_empty() == true, .is_infinite() == false
```
rhagenson: Some of these ranges are repeated such as `Range(0, 10 -1)` on lines 202 and again on 210.

rhagenson (commented):
I found this RFC quite difficult to follow in places. From what I understood, I am in favor of the core idea of allowing for ranges which have built in guards against unsuspecting infinite running -- however I am not sure I am in favor of this particular approach to that problem. I have made some suggestions for how to clarify meaning and have asked what is maybe a distracting question about why carve out empty ranges at all. This final points is meant along the lines of if we are primarily intending to protect users against unsuspecting infinite behavior, why convert existing infinite ranges to the proposed "empty" rather than transform the range to be finite where we can do so?

`Range(0, 10, -1)` could just as well be a finite range that iterates backwards from 10 down to 0 by the -1 step and that approach is just a reasonable to save users from infinitely iterating down from 0 by the -1 step, never reaching 10.

text/0000-introduce empty ranges.md
stefandd: The argument against transforming `Range(0, 10, -1)` into `Range(10, 0, -1)` is that Range evaluates expressions using the provided parameters in their meaning of `min` and `max`. In conjunction with the `step` parameter there is a directionality (I called trajectory). Even the current implementation uses the term "direction", honoring the meaning of the parameters names. Then, just like an if clause prevents a block of code to be executed if the condition is evaluated as false, the Range should be empty if those conditions (here the *progress* condition) are not met. Re-arranging/swapping the parameters that feed into the underlying expression and thereby basically changing the meaning of the parameters `min` and `max` feels wrong.

A distinct advantage of producing an empty range in those currently infinite cases instead of enforcing `min <= max` or changing the sign of `step` is that the user has control over the order within the set of values in the interval [min, max) that are produced by Range. Within the proposal, Range(-10, 0, 1) produces `-10, -9, ..., -1` in that exact order which is different from Range(-1, -11, -1) which produces the same numbers but in opposite order. With your suggestion, only the ascending order of iteration values can be created which is a significant reduction of functionality even from the current implementation.

text/0000-introduce empty ranges.md
stefandd: No, infinite bounds are *not* already covered by the `min == max` arguments because one cannot meaningfully compare an infinite value to another infinite value! Therefore, the second part of the text you quoted explains that Range(Inf, Inf, 1) is empty for two reasons: the possibility of progress can't be evaluated, thus there is *no progress* (first criterion), and even if that was disregarded, the second criterion is also not fulfilled since this iterator would not produce finite values.

However, again I realize this could really be worded better!

text/0000-introduce empty ranges.md
stefandd: Yes, this should be fixed

text/0000-introduce empty ranges.md
stefandd: I am not defining `progress` here -- it is calculated as `not no_progress` to cleanly show how the expression in the implementation follows from the expression for *no-progress*. In order to discuss what *empty*-- the core of the proposal--means, `no-progress` is more useful. However, its positive `progress` form is used for the implementation and is thus introduced here.

text/0000-introduce empty ranges.md
stefandd: Yes, this should be shortened and simplified.

text/0000-introduce empty ranges.md
stefandd: These three distinctions are not what the RFC is about -- they are merely used to explain how all that is explained above is implemented. I did not mean to and might indeed not have made a strong case for how I would implement this, although I think it is logical.
Like the original implementation evaluates the infinite cases, my discussion also uses `no-progress` as one of the elements based on which the distinction empty or not empty is made. However, implementation-wise I liked it better to use a positive `progress` (possible) expression.

text/0000-introduce empty ranges.md
stefandd: Yes. c&p errors that need to be fixed

text/0000-introduce empty ranges.md
stefandd: > The purpose here seems to be to discuss how min, max, and/or step being +Inf, -Inf or NaN will now produce empty ranges

No, Range(0, Inf, 1) is not empty under the proposed criteria/rules, since progress is possible and only finite values will be created (in the first 1.7976931348623157e+308 iterations). I think the sentence
>Similarly, while a `step` argument of `+-Inf` may "pass" the inspection by the *no-progress criterion*, no finite points within the mathematical range `[min, max)` can be computed, including when `min, max` themselves are `inf`

is poorly worded. I meant this strictly for the case of `step` being `+-Inf` in which case the range is *empty*, no matter whether `min, max` are finite or infinite.

text/0000-introduce empty ranges.md
stefandd: Yes, the language needs to be improved

text/0000-introduce empty ranges.md
stefandd: Yes, your point is exposing a flaw in the parameter naming. `min` and `max` are poor choices when descending Range iterators such as Range(-10, -20, -1) are allowed, where clearly -20 is the *minimum* value. A much better choice of naming the parameters in the draft and the code would be `start`, and `end`.

text/0000-introduce empty ranges.md
stefandd: Yes, I should try to simplify it a lot more. I realize it is too wordy and in places too hard to follow.

rhagenson: Thank you. I think there is a general tendency by most folks when describing mathematical concepts in plain language to try to keep caveats in the reader's mind which eventually leads to wordiness. Simplifying and reducing this wordiness can significantly help readability -- even if we sometimes have to remind ourselves of the caveats that otherwise would have been kept in place.

text/0000-introduce empty ranges.md
rhagenson: I see what you mean, but still do not think you have made a case of "empty" ranges over "parameter rearrangement" as a solution. I put more thought into this and the critical element is whether we want to keep an implied directionality to `min` and `max` which then interacts with the direction of `step` or if directionality should *only* be determined by `step` with `min` and `max` simply setting the interval.

In the latter, `Range(-1, -11, -1)` would indeed iterate from -1 to -11, while `Range(-1, -11, 1)` would iterate from -11 to -1. I think the biggest argument against a parameter rearrangement is the half-closed interval of range, which should not change. If I provide two values (`min` and `max` as the are named now), I should know which "side" it will stop at. Empty ranges approach this better, but there might be a good use case for some `Interval` class which works via parameter rearrangement which is *not* half-closed. Changing the half-closed nature of Range is far too breaking of a change for parameter rearrangement to be viable.

Once you have read this message, please resolve this conversation thread. I am sufficiently on the side of empty range over parameter rearrangement now. It only took me sitting on the idea for a day to realize the problem with parameter rearrangement.

text/0000-introduce empty ranges.md
rhagenson: Okay, I understand from the point of not being able to compare infinites. I was thinking about it from both a representation standpoint (inf is the same bits as inf) and from an inductive `x +1 == x +1` for any value of x up to inf.

text/0000-introduce empty ranges.md
rhagenson: > No, Range(0, Inf, 1) is not empty under the proposed criteria/rules, since progress is possible and only finite values will be created (in the first 1.7976931348623157e+308 iterations). I think the sentence

Correct, I missed this distinction and think that was due mostly to me being tired, not to wording.

text/0000-introduce empty ranges.md
rhagenson: I agree completely! I think it is a small change to make the parameters `start` and `end` but that is more accurate than `min` and `max`, the latter of which suggests the parameter rearrangement that I showed above in another comment is a poor fit for the purpose of Range.

text/0000-introduce empty ranges.md
rhagenson: > This expectation can be met by introducing and in such and similar cases returning *empty* ranges. One unexpected side effect of Pony's current behavior is that even though finite bounds are given, an *infinite* iterator can be created from them which, when used without consideration for this side effect, can produce indefinitely-running actors and "hanging" programs.

The definitions I commented on here seem like the representation of your central point in the introduction. That is, we are carving out three distinct categories of Range to be finite, infinite, or empty. Although, my comment here was made under the "why empty, why not parameter rearrangment" line of thinking which I have now convince myself does not fit with the purpose of Range anyhow. Had I not been in that line of thinking, I would not have made the comment I did here.

I would suggest adding language to the introduction that we are making changes to prevent unsuspecting infinite ranges when, according to construction, progress cannot be made and therefore the range should be empty, not infinite. Then add in alternatives, that preventing such unsuspecting infinite ranges could be done via parameter rearrangement, but due to the half-closed interval that defines Range, such rearrangement is likely to introduce more errors, not fewer. Keeping Range to always move from `min` to `max` and having it be empty when progress cannot be made is a clearer design.

text/0000-introduce empty ranges.md
rhagenson: Been there!

rhagenson:
@stefandd I have added some comments to continue the conversation. Biggest update for my thinking is, after thinking on it more, convincing myself a parameter rearrangment approach is a bad fit for solving the same underlying problem that empty ranges intends to address (i.e., preventing unsuspecting infinite ranges).

SeanTAllen:
I haven't had a chance to review this yet. Sorry, very busy with non-pony things. Please do not take that as a lack of interest.

text/0000-introduce empty ranges.md
stefandd: I rewrote quite a bit of this in the hope to make the discussion more concise.

rhagenson:
@stefandd Excellent to see the rewrite! I will review this when I am home from work today. I really like/support the changes you are putting forward so I hope the rewrite clarifies points where I was previously confused.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+  ...
+  let f: Array[Array[String]] = csv.read("data.csv", ",") // reads whatever conforms to csv and returns a (possibly empty) array
+  for idx in Range(1, f.size()) do // start at 1 to skip header line
+    ... // perform only if f is not empty and has at least 1 data line
+  end
+```
+
+This currently leads to a hanging actor/memory crash because the Range(1, f.size()) instead of being empty iterates infinitely in cases when the file "data.csv" exists but is empty, leading to an empty array in f. This code isn't particularly elegant or safe (a guard regarding f.size() should be put here), but it can still serve an example for how a programmer who would expect this Range to be empty in case of an empty f Array could utilize the Range parameters directly to ensure that the for loop is not executed by writing `Range(1, f.size())`.
+
+Another problematic example of the current implementation is the response to any of the parameters being `Nan`.
```
rhagenson:
```suggestion
Another problematic example of the current implementation is the response to any of the parameters being `NaN`.
```

Simple typo. :smile:

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+-nan(ind)
+-nan(ind)
+-nan(ind)
+-nan(ind)
+-nan(ind)
+-nan(ind)
+-nan(ind)
+```
+
+`NaN` is not a value and can therefore not possibly lie within any valid range. The same is true when one of the bound parameters is `NaN`. Under this RFC Range iterators that use `NaN` in any parameter are *empty*. Pony currently also considers those Ranges *infinite* were the `min, max` bounds are finite and identical but the `step` parameter is `+-Inf`. These cases too would be treated as empty since no iteration is necessary to advance from `min` to `max` regardless of the magnitude of `step`.
```
rhagenson:
```suggestion
`NaN` is not a value and can therefore not possibly lie within any valid range. The same is true when one of the bound parameters is `NaN`. Under this RFC Range iterators that use `NaN` in any parameter are *empty*. Pony currently also considers those Ranges *infinite* where the `min, max` bounds are finite and identical but the `step` parameter is `+-Inf`. These cases too would be treated as empty since no iteration is necessary to advance from `min` to `max` regardless of the magnitude of `step`.
```

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Proposed reclassifications
+
+This RFC partitions the set of currently *infinite* `Range` cases into those now considered *empty*, and a very small set of those that remain *infinite*. To discuss this discrimination, we first list three criteria. All three must be met for a currently *infinite* Range to *not* now be reclassified as empty -- if either of these criteria is violated, the Range is *empty*; if all three are met, a currently *infinite* Range remains so under this proposal.
+
+  **Criterion 1** - progress from `min` to `max` must be possible. This necessitates the "no-progress expression" (discussion below) to be `false`.
+  
+  **Criterion 2** - the iterator that realizes the progress must produce *finite* values that lie within `[min, max)`.
+  
+  **Criterion 3** - none of the parameters `min, max, step` can be float `NaN`. (discussion below)
```
rhagenson: I am confused here. 

> ...if all three are met, a currently *infinite* Range remains so under this proposal.

Suggests to me that the criteria are what we will not consider all infinite iterators to be. However, criterion 2 is

> **Criterion 2** - the iterator that realizes the progress must produce *finite* values that lie within `[min, max)`.

Which is decidedly finite.

---

I think what happened here is that we have a sort of excluded middle situation where the criterion are set up to differentiate empty from infinite, but we also need to differentiate both of those from finite.

Are there any "currently *infinite* Range[s]" which meet criterion 2?

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+
+### Criterion 1 -- no-progress expression
+
+The first criterion is related to the question whether progress from `min` to `max` is possible or needed. Numerical values including the floating point values `+Inf` and `-Inf` can be meaningfully compared in expressions like `min > max` or `max > min`. Based on this, an obvious absence of *progress* of the iteration from min towards max based on the sign of the Range parameters can be tested by the following *no-progress expression*: `((min <= max) and (step <= 0)) or ((min >= max) and (step >= 0))`, for both finite and infinite bounds.
```
rhagenson, line 133: You add the "or needed" part here, which is not in the original definition you provide above. I suspect you mean to guard for when `min == max`, correct?

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+-3
+-4
+-5
+-6
+-7
+-8
+-9
+```
+
+All of these but the first value lie outside the range 0..10 that the user requested the Range class to generate. However, not the least based on the word *Range* -- defined by Merriam Webster as ".. 7a: a sequence, series, or scale between limits" -- the user can expect this to produce an iterator that generates all values along the *trajectory* from `min` to `max`, interspaced by `step`. Therefore, no sequence value should lie *outside* `[min, max)`.
```
rhagenson, line 42: This definition excludes "infinitely" producing the same value. I agree that Range should not include such a use case. We might want to add a suggestion -- not necessarily where I am making this comment -- that if someone needs a infinite production of the same value to use "itertools" `Iter[A].repeat_value(value)`.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+> ```pony
+> let infinite_range1 = Range(0, 1, 0)
+> infinite_range1.is_infinite() == true
+> 
+> let infinite_range2 = Range[I8](0, 10, -1)
+> for _ in infinite_range2 do
+>   env.out.print("will this ever end?")
+>   env.err.print("no, never!")
+> end 
+> ```
```
rhagenson: I do not think any of this is necessary for this RFC. None of it is relevant to Alternatives and by its own admission discussed in the Motivation already.

rhagenson (commented):
I am in favor of this, but would like someone else to review it. I see no major flaws with the ideas presented here.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,443 @@
...
+    ... // perform only if f is not empty and has at least 1 data line
+  end
+```
+
+This currently leads to a hanging actor/memory crash because the Range(1, f.size()) instead of being empty iterates infinitely in cases when the file "data.csv" exists but is empty, leading to an empty array in f. This code isn't particularly elegant or safe (a guard regarding f.size() should be put here), but it can still serve an example for how a programmer who would expect this Range to be empty in case of an empty f Array could utilize the Range parameters directly to ensure that the for loop is not executed by writing `Range(1, f.size())`.
+
+Another problematic example of the current implementation is the response to any of the parameters being `Nan`.
+
+```pony
+  let nan = F64 = F64(0) / F64(0)
```
redvers, line 62: Typo?
```
let nan: F64 = F64(0) / F64(0)
```

stefandd: Yes, thanks

text/0000-introduce empty ranges.md
stefandd: To clarify -- by finite I mean not `+-Inf`. But an infinite upper bound does equate to the production of infinite values! Since an `+-Inf` step parameter is ruled out in criterion 2, a Range such as [finite_start, Inf, finite_step) will only produce *finite* values as required by criterion 2, but is still an *infinite* Range as per current Pony implementation (and would remain so).

For example, this range is currently *infinite*:

```pony
use "collections"

actor Main
  new create(env: Env) =>
    let p_inf: F64 = F64.max_value() + F64.max_value()
    env.out.print(Range[F64](0, p_inf).is_infinite().string())
```
This code prints out `true`, since currently Pony makes a Range infinite if any of the bounds is infinite, and this Range [0, +inf) is an example of those that remain *infinite* under this RFC because they pass Criterion 2. Criterion 2 is needed, even though I agree with you that all of the *normal* Ranges met it, too.

Of course a more radical approach would be to axe *all* occurrences of Nan, +Inf, -Inf in Ranges and make all of them empty. The implementation would become simpler and requires only criteria 1 and a modified 3. The reason why I haven't proposed that is that I wanted to conserve as much of the original implementation intent as possible or plausible.

text/0000-introduce empty ranges.md
stefandd: Yes. It is convenient to roll the `min==max` case into the expression by adding the "needed". I will adjust the phrase in the first description of the criterion. Thanks for catching this!

text/0000-introduce empty ranges.md
stefandd: Will do.

text/0000-introduce empty ranges.md
stefandd: >Your belief that there will be user code bugs is editorialzing at this point. Alternatives should be presented with the best >possible case for those alternatives such as "Leaving it as is means that code that might exist that I am unaware from >my search on GitHub would continue to work".

Sean reviewed an earlier draft and sort-of insisted on discussing the status quo.

Or do you literally mean lines 428 - 443? - these could surely be deleted. Which parts would you remove?

text/0000-introduce empty ranges.md
rhagenson: I was just about to disagree and state that there are infinite values in `[0, +inf)` however upon more consideration you are correct since the infinite exists on the open bound.

text/0000-introduce empty ranges.md
rhagenson: I highlighted the portion that I thought was not needed, but it does not look like that is visualized in the review properly. I was referring to this:

> Nonetheless, as laid out in the summary and motivation sections, the use of the word Range suggests a behavior that is different from the current one. Therefore, if this RFC is not accepted, I'd suggest to better point out how Pony's use of `Range` is contrary to both the mathematical meaning and the use of iterator objects of that name in many other programming languages, and how this can lead to infinite loops over unintendedly infinite `Range` iterators. The current Range docstring does state this behavior and provides an example (infinite_range2), but this is possibly not prominent enough given the gravity of the discrepancy between reasonable expectation and behavior.
> 
> >  If the `step` is not moving `min` towards `max` or if it is `0`,
> >  the Range is considered infinite and iterating over it
> >  will never terminate:
> > 
> > ```pony
> > let infinite_range1 = Range(0, 1, 0)
> > infinite_range1.is_infinite() == true
> > 
> > let infinite_range2 = Range[I8](0, 10, -1)
> > for _ in infinite_range2 do
> >   env.out.print("will this ever end?")
> >   env.err.print("no, never!")
> > end 
> > ```

I understand addressing the status quo from the purpose of explicitly stating what we lose by making these changes, but doubling back to the summary and motivation as you do here is not contributing towards addressing the status quo.

The status quo does more than merely not break existing code. You could also discuss how, after these changes, would have a new confusion around the reasonable assumption that a Range will produce values, but by adding empty Ranges this will no longer be true. The suggestion here is to move from having two categories (finite and infinite) of Ranges to having three categories (empty, finite, infinite). This can do more than break existing code, it could cause future confusion similar to what you had for expectations of Range to create this RFC. Your proposed solution to teaching this change is to modify the docstring...but you argue in this Alternatives section that the docstring was "possibly not prominent enough given the gravity of the discrepancy between reasonable expectation and behavior" -- something which is also possible for someone with an assumption that Ranges produce values.

text/0000-introduce empty ranges.md
stefandd: I was editing while you responded ... but yes, the key point is that the infinite upper bound does not change the fact that all iterations create by a finite start and iterations of `+ finite step` will be finite.

text/0000-introduce empty ranges.md
stefandd: You raise good points. 

> You could also discuss how, after these changes, would have a new confusion around the reasonable assumption that a Range will produce values, but by adding empty Ranges this will no longer be true.

With this I only agree partially because Pony *had* empty Ranges but provided no means to test for that like .is_empty(). For example `Range(0, 0)` or `Range(10, 10, 10)` produce currently no iteration so that there should not be a reasonable user expectation for a Range to always produce values.

Should this potential confusion still be added to Drawbacks then? Also, I will delete the passage you quoted..

stefandd: > You could also discuss how, after these changes, would have a new confusion around the reasonable assumption that a Range will produce values, but by adding empty Ranges this will no longer be true.

I only agree partially because Pony *had* empty Ranges but provided no means to test for that like .is_empty(). For example `Range(0, 0)` or `Range(10, 10, 10)` produce currently no iteration so that there should not be an user expectation for a Range to always produces values.

rhagenson: I would add it to Drawbacks, yes.

rhagenson (approved):
I see nothing objectionable.

SeanTAllen:
I haven't had time to look at this at all. Apologies, long drawn out work project. I might have time in November, certainly not before. I will try to give this the time in deserves soon.

jemc (approved):

SeanTAllen:
Actually, during sync, we covered this slightly. Joe gave me an overview, said he was in favor, and based on the overview, and his support, I am hereby giving my "I'm fine with this going through as others have said it is good".

Side note, Joe read the Drawbacks section out loud and the call and kudos on the drawbacks section. It was very very well written.

redvers (approved):

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,431 @@
...
+
+Another fair alternative is to leave everything as it is. Pony has been in fairly wide-spread use academically and to some degree even commercially, without anyone reporting the issues that for the author were the reason to start working on this RFC -- which was hanging (eventually running out of memory) Pony code. One could therefore speculate that the side effects and instances in which currently infinite Ranges are created are sufficiently documented and known by the majority of users of Pony.
+
+# Unresolved questions
+
+The function `.is_empty()` was added to the proposal as a counterpart to `.is_infinite()`. However, it would be fair to ask whether such an `.is_empty()` is really necessary since there already is a `.has_next()`. While for infinite loops it makes sense to have an `.is_infinite()` which informs that `.has_next()` will *remain* `true`, in the opposite case of empty Ranges, `.has_next()` will be `false` right away, and an `.is_empty()` is therefore somewhat redundant.
```
pmetras, line 431: The difference in semantic between `has_next` and `is_empty` is that `is_empty` describes a static property of the `Range` that you can know in advance before using it, while `has_next` is a dynamic property while you are executing the `Range`. The `is_empty` function allows to take decisions about a `Range`, for instance when its `start`, `end` or `inc` parameters are results of calculation, before using it. As such, I don't see them as redundant.

For instance
```pony
let r = Range(0, 10, 1)
// Before using the range
// r.is_empty() --> false
// r.has_next() --> true
for i in r do
  // When i == 10
  // r.is_empty() --> false
  // r.has_next() --> false
end
```

text/0000-introduce empty ranges.md
pmetras: Typo is still present

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,431 @@
...
+
+It is argued here that an iterator using the name `Range` should produce a sequence of values that lie within the `min, max` bounds. Therefore, most of what are now *infinite* ranges should instead be *empty* ranges (where `.has_next() == false`). The RFC proposes a delineation of the so-far *infinite* cases into one of two types: those that are now *empty*, and those which remain *infinite*.
+
+A smaller issue proposed to be fixed by this RFC is the naming of the Range bounds `min` and `max`. Especially when an implementation allows for descending Ranges like Pony's does -- e.g. `Range(-10, -20, -1)`, naming the first parameter `min` and the second `max` is clearly mathematically wrong. Therefore if accepted, in the actual implementation these will be renamed to `start` and `end`. However, to not confuse things, they will still be named `min, max` throughout here.
```
pmetras, line 12: Renaming parameters could be a breaking change (when named arguments are used) but in the present case, current parameter names are misleading and this renaming is expected.

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,431 @@
...
+
+Besides the cases with obvious sign-incompatibility of the three Range parameters, this expression also covers the cases of equal bounds `min == max` and `step == 0`. Equal bounds correspond to `[N, N)` which cannot contain any value, and the case of `step == 0` to a situation where no progress is possible. In the current implementation of Range, a somewhat similar couple of expressions are used with the comment `// no progress` and `// progress into other directions`, but they are treated as a sufficient condition for *infinite*. Here, this criterion is used to test for empty Ranges because no valid *trajectory* from min to max does exists with the provided parameters.
+
+#### Examples for empty Ranges that violate criterion 1
+
+`Range(0, 10, -1)` is *empty* nd no longer *infinite*, but so is in fact `Range(0, inf, -1)` that is also considered *infinite* currently. While one can generally never produce a complete list of elements that would incrementally move from say 0 to +Inf, we can know in the case of `Range(0, inf, -1)` that the Range must be *empty* because of the incompatible sign of the step parameter relative to the signs of `min` and `max`.
```
pmetras: Typo: `Range(0, 10, -1)` is *empty* and no longer *infinite*

stefandd: What is the typo?

text/0000-introduce empty ranges.md
stefandd: Agreed.

text/0000-introduce empty ranges.md
rhagenson: "and" is missing an "a"

text/0000-introduce empty ranges.md
```diff
@@ -0,0 +1,431 @@
...
+Furthermore, having empty Ranges when using `Range` with calculated, non-literal bounds also has an *expressive* use (see CSV reader example in Motivation section) which the reinterpretation of the parameters to always return a non-empty Range would preclude.
+
+Another fair alternative is to leave everything as it is. Pony has been in fairly wide-spread use academically and to some degree even commercially, without anyone reporting the issues that for the author were the reason to start working on this RFC -- which was hanging (eventually running out of memory) Pony code. One could therefore speculate that the side effects and instances in which currently infinite Ranges are created are sufficiently documented and known by the majority of users of Pony.
+
+# Unresolved questions
```
pmetras, line 429: Also, perhaps this RFC should also apply these changes to the [`Reverse`](https://stdlib.ponylang.io/collections-Reverse/) class that has the same problems as `Range`, and updating its documentation too.

@stefandd, the suggested changes are very good. The fact that pony has a `Reverse` iterator may question why we allow negative `inc`, but at least your proposal will prevent some type of bugs.

text/0000-introduce empty ranges.md
stefandd: Ah - thx.

text/0000-introduce empty ranges.md
rhagenson: My initial thought was "surely this would affect `Reverse` implicitly" as I thought `Reverse` was implemented via `Range` but that is not true. I am in favor of having `Range` and `Reverse` mirror one another since even the docstring for `Reverse` states itself as "(i.e. the reverse of `Range`)." Whether that should be rolled into this RFC or not, I have no strong opinion on. The one caveat for why `Reverse` is not a direct call to `Range` underneath is that the `Number` parameter can be unsigned. There would have to have a `direction` argument to `Range` which switches `+ increment` or `- increment`.

rhagenson (approved):
I believe there is still an open conversation about how/if we should change `Reverse` to match the changes here to `Range` (I am in favor of such consistency) -- but the changes to the RFC since opening for final comment I agree with.

SeanTAllen:
@rhagenson FYI, I would be wary of assuming that any docstrings are correct rather than someone (a volunteer) attempting to explain what they are seeing. I think consistency is good, if that happens to match the docstring, great! If not, I think we can also reasonably decide the doc string is wrong.

rhagenson:
Sean, I presume your comment was due to my talking about the docstring for `Reverse`. I took the docstring to mean "`Reverse` is meant to be `Range` in reverse" I was not so much assuming it to be correct as deciding I agreed with the docstring as written that it **should be** that way.

SeanTAllen:
@rhagenson gotcha

rhagenson:
If we go with the changes here as written without any change to `Reverse` then `Range` and `Reverse` would no longer be related in the same way and the docstring should change so as to not suggest a relationship that we know does not apply any longer.

SeanTAllen:
We discussed this during sync and Joe and I both believe that if we agree with the statement that "reverse should be the opposite of range" then changes to Reverse should be part of this RFC.

SeanTAllen:
I agree with Ryan's statement that Reverse should be the opposite of Range. Therefore, I think changes to reverse should be added to this RFC. That said, I think that it might turn out when trying to make Reverse the opposite of Range that the statement that it should be opposite is in fact, non-sensical.

SeanTAllen:
@jemc noticed during sync that in terms of inclusivity, Range and Reverse are not opposites of each other. In general, it appears that Reverse isn't at all the opposite of Range. Across many factors

@kulibali raises the interesting point, "if Range can take a negative step to go backwards, why do we need Reverse"?

SeanTAllen:
I think it would be reasonable to consider removing reverse as part of this RFC.

SeanTAllen:
After additional conversation during sync, Joe, Gordan, and I have arrived on the following:

This RFC should be updated to remove Reverse.

If we find that using Range instead of Range turns out to be problematic, then an RFC specifically for adding a new Reverse. As part of that RFC, we can decide what characteristics we want from Reverse.

stefandd:
I am also in favor of retiring `reverse` as it is almost the same, e.g. `Reverse(max, min, 2)` is similar to `Range[ISize](max, min-1, -2)`. For the sake of a best-outcome discussion, I still want to offer one argument for why someone could possibly want it to remain. Unlike `Range` it plays more nicely with the default unsigned integer types for descending series. For example, if you did:

```pony
use "collections"
actor Main
  new create(env: Env) =>
    for e in Reverse(10, 2, 2) do
      env.out.print(e.string())
    end
    env.out.print("\n")
    
    for e in Range(10, 1, -2) do
      env.out.print(e.string())
    end
```
it outputs

```
10
8
6
4
2

10
8
6
4
2
0
18446744073709551614
18446744073709551612
```
The reason is of course that `-2` gets silently converted to `USize`. Still, I'd be in favor of a single Range class, especially since the current one and the one proposed here can deal with ascending and descending Ranges.

Note however that under the current proposal and because of this silent conversion to the default `USize`, the Range in the code example above would be empty, whereas currently `Range(10, 1, -2)` is one of these nasty indefinitely iterating Ranges I try to get rid of herewith. The above example nearly hung the Playground runtime...

SeanTAllen:
There's no silent type conversions in Pony @stefandd. What do you mean by the term?

Ah, looking more I think you mean that Range is generic and it defaults to USize which seems somewhat problematic given that it is unsigned and we are going to have people using negative values. I think as part of this RFC, it would make sense to chanğe the default from USize to ISize. 

Or perhaps, removing the default entirely. My first thought though is that ISize as a default is good.

@jemc thoughts?

I think there is a decent argument in favor of reverse being useful because it can represent a wider range of numbers to reverse through than is allowed by Range.

stefandd:
> There's no silent type conversions in Pony @stefandd. What do you mean by the term?

I mean that a negative step parameter like `-2` above is cast to a very large `USize` since `USize` is the default type of the generic Range type parameter ```[A: (Real[A] val & Number) = USize]```. You are right that it isn't really silent. But because that cast happens and the way that `Range` *adds* the step whereas `Reverse` substracts it, one cannot get the same descending series as type `USize`. Instead, one would have to write `Range[ISize](10, 1, -2)` whereas `Reverse(10, 2, 2)` actually produces `USize` values.

However, I don't argue against the removal -- I wanted to merely point out that subtle difference!

SeanTAllen:
There's no casting there @stefandd. It is actually a USize from the get-go, but it wraps around. As I detailed in my updated comment, I think that is quite problematic and that either there should be no default or it should be an ISize as the default. I agree that USize SHOULDNT be the default type for range and would not argue against there being no default.

SeanTAllen:
@stefandd Even if we keep range, I think it is very important as part of this PR to either

- Not have a default type for Range so that you have to specify the type in question or...
- Set it to ISize, because USize is a horrible choice given the existence of negative step values as an option.

stefandd:
@SeanTAllen I agree - having a default `ISize` for the `Range` type would be a logical remedy though it may then break more existing code since it is used heavily for indexing of Arrays which usually needs `USize`.

Yes, I am afraid this would be a very breaking change...

SeanTAllen:
@stefandd I think that is reasonable break. The current default is, I think, downright awful. Really really bad. I think this is a good breakage.

A giant PITA breakage but it removes a serious issue with the API that you pointed out. I think that not having any default would be a good thing here as well.

SeanTAllen:
I think the other thing that could be argued from this, if we don't want the breaking change is that Range shouldn't support negative decrementing and that should be the sole domain of an updated Reverse.

You are perhaps unintentionally bringing me around to that point of view @stefandd.

stefandd:
> You are perhaps unintentionally bringing me around to that point of view @stefandd.

Yes, that is why I felt it was important to point out the type assymetry, no matter what the final conclusion for this RFC will be. Also it is a good discussion to have...

rhagenson:
I am on the side of having a `Range` that increments and a `Reverse` that decrements. Changing the default parameter to be `ISize` to allow `Range` to be used for both increment and decrement sounds like a change more out of convenience than principle.

SeanTAllen:
> I am on the side of having a `Range` that increments and a `Reverse` that decrements. Changing the default parameter to be `ISize` to allow `Range` to be used for both increment and decrement sounds like a change more out of convenience than principle.

I dont think it is convenience. I think the combination of API and that default choice is error prone. And "dont create error prone defaults" is a principle of mine.

SeanTAllen:
Ok, so thinking about this. I think @stefandd / @rhagenson is something like this.

- Remove the footguns. All choices of type to Range or Reverse should give you unsurprising results. To do this we need to make the `inc` value not be of type `A` which could lead to weird footguns, but instead `inc` should always be positive and probably of a fixed type like `ISize`. 

Except... well that doesn't work for floats. I can't use ISize with a Float that I want to increment by a non-integer value.

I dont offhand know how do support for integer types and floating point types without also having the foot guns except by having 2 different Range and Reverse implementations or... by having lots of runtime errors if you set things up poorly.

I'm very open to hearing ideas on how to address and if it is possible to remove the footguns or if the best we can do is have ISize rather than USize as the default.

So after looking at `Reverse` for a while, I think a way we could address the footguns is to "not allow negative increments" and then have interesting "if you do X" rules of getting some side effect. Something like... what reverse is currently doing:

```
  If `dec` is 0, produces an infinite series of `max`.

  If `dec` is negative, produces a range with `max` as the only value.
```

SeanTAllen:
Another oddity of `Reverse` is that unlike `Range`, calling it when `has_next` is false, doesn't create an error.

rhagenson:
> > I am on the side of having a `Range` that increments and a `Reverse` that decrements. Changing the default parameter to be `ISize` to allow `Range` to be used for both increment and decrement sounds like a change more out of convenience than principle.
> 
> I dont think it is convenience. I think the combination of API and that default choice is error prone. And "dont create error prone defaults" is a principle of mine.

I think the error prone default you had in mind is the one Stefan showed where, since the default parameter is USize, a -2 in effect is really adding USize.max_value()-2. Right?

If so, that was not what I was thinking about for the major error we were intending to solve. I was working from the idea of the type parameter determines the range of values that are possible and the method determines the direction. So if I want to have a range that includes signed integers I need an ISize, but if I only want positives then I want USize. I then assume the latter is more common based on how I have seen Range being used.

SeanTAllen:
> > > I am on the side of having a `Range` that increments and a `Reverse` that decrements. Changing the default parameter to be `ISize` to allow `Range` to be used for both increment and decrement sounds like a change more out of convenience than principle.
> > 
> > 
> > I dont think it is convenience. I think the combination of API and that default choice is error prone. And "dont create error prone defaults" is a principle of mine.
> 
> I think the error prone default you had in mind is the one Stefan showed where, since the default parameter is USize, a -2 in effect is really adding USize.max_value()-2. Right?
> 
> If so, that was not what I was thinking about for the major error we were intending to solve. I was working from the idea of the type parameter determines the range of values that are possible and the method determines the direction. So if I want to have a range that includes signed integers I need an ISize, but if I only want positives then I want USize. I then assume the latter is more common based on how I have seen Range being used.

Right but having documentation that says "USE NEGATIVE VALUES, its cool" and a default parameter that makes that negative value actually an integer underflow is just an awful confluence of items. We can't in good conscience drop Reverse if that is the API we are giving folks.

rhagenson:
Absolutely.

I will have to sit down later today and collect my thoughts on the API that is "best" in my opinion. Not to get too far away from the original point of this RFC any change must include the empty ranges as that was fixing one possible footgun in the existing API.

SeanTAllen:
> Absolutely.
> 
> I will have to sit down later today and collect my thoughts on the API that is "best" in my opinion. Not to get too far away from the original point of this RFC any change must include the empty ranges as that was fixing one possible footgun in the existing API.

I think we should entertain getting away from the RFC. If this RFC is impetus for something different than it is right now, I think that is ok.

I also think that we could say, we are ok with this RFC "just as it is" and we approve it and then ask @stefandd or something else to tackle the "larger issues" we are addressing now.

pmetras:
What about, like initially suggested by @stefandd, having the following signature:
```pony
Range[A: (Real[A] val & Number) = USize] is Iterator[A]

  new create(from: A, to: A, inc: A = 1, backward: Bool = false)
```
Advantages
* Compatible with existing code. Can still be used to index arrays and collections that use default `USize`.
* Can still be used with floats.
* `Reverse` becomes useless. Less code to maintain.
* No questions about increasing or decreasing ==> look at the `backward` value.

There's still a question though about the interval of the ranges' values: is it closed or open? If it is open, counting down from 10 to 1 with `Range[USize](10, 0, 1, true)` does not give the value `0`. How would you write a `Range[USize]` to include the end value? The meaning of `Range(10, USize.max_value(), 1, true)` is not obvious.

rhagenson:
@ponylang/committer @stefandd @pmetras 

I do not see a clear consensus on a direction for unifying `Range`/`Reverse` functionality. Is anyone opposed to accepting this RFC as-is and having a discussion about further changes to `Range`/`Reverse` elsewhere (Zulip or open office hours both seem sensible to me) so we do not delay the closing of one source of error (unintended infinite ranges) which this RFC addresses?

pmetras:
I agree that the RFP as it is solves a first problem. We can open another one later to decide about `Reverse` status.

jemc:
My suggestion for resolving the `Reverse` asymmetry (and the fact that you can't currently go backward with an unsigned integer type in `Range`) would be:

- Remove the `Reverse` type
- Add a `Range.reverse(...)` constructor, which is symmetrical to the main constructor. This is how you'd go backward when using an unsigned integer type for the increment amount.

jemc:
On the sync call we discussed that we want to leave the `Reverse` topic for another RFC. If anyone objects to this, feel free to comment within the next week. We think this is a reasonable iterative improvement without introducing any changes to `Reverse`.

We will vote on this during next week's sync call, unless the RFC author wants to make any further changes between now and then, or delay the vote based on feedback.

pmetras:
I assume that acceptance of this RFP will invalidate this one: [added range_redesign rfc](https://github.com/ponylang/rfcs/pull/147/files).

SeanTAllen:
This was accepted at the sync meeting. We still would like to see another RFC to address Reverse and various other issues.

SeanTAllen:
RE: additional RFC for Reverse and Range, I've opened an "official" issue for that: https://github.com/ponylang/rfcs/issues/204

PR #202, opened 2022-06-21
Enhance `term` package with mouse input and other enhancements

nogginly:
Here's a video recording of my test app as a GIF where you can see 
* the cursor tracking the mouse, 
* button up/down for left and right buttons, 
* mouse dragging, 
* mouse "wheel" up/down scroll, 
* cursor on/off while tracking mouse, 
* capture of `^C` and `^Z` as input,
* alt buffer use restoring original contents on exit.

term-enhancements-video

### DIFF
--- a/text/0000-stdlib-term-mouse-input-plus.md
+++ b/text/0000-stdlib-term-mouse-input-plus.md
@@ -7,41 +7,16 @@
 
 Enhance ANSI terminal support in package `term` with mouse input handling, additional ANSI escape codes to support alt buffer and cursor switching, and capture of SIGINT and SIGTSTP to handle `^C` and `^Z` as input.
 
-**Table of contents**
-- [Summary](#summary)
-- [Motivation](#motivation)
-- [Detailed design](#detailed-design)
-  - [New `ANSINotify` methods](#new-ansinotify-methods)
-  - [New `ANSI` escape codes](#new-ansi-escape-codes)
-  - [Enhance `ANSITerm`](#enhance-ansiterm)
-    - [Mouse input handling](#mouse-input-handling)
-    - [Ctrl-C and Ctrl-Z input handling](#ctrl-c-and-ctrl-z-input-handling)
-  - [ANSI terminal options](#ansi-terminal-options)
-- [How We Teach This](#how-we-teach-this)
-- [How We Test This](#how-we-test-this)
-- [Drawbacks](#drawbacks)
-- [Alternatives](#alternatives)
-- [Reference Implementation](#reference-implementation)
-- [Questions](#questions)
-- [References](#references)
-
 ## Motivation
 
-> Why are we doing this? 
-
-Improve ANSI terminal application capabilities when implementing text-based UIs.
-
-> What use cases does it support? 
-
-Mouse input handling, alt buffer switching, cursor hiding and capturing missing control keys.
-
-> What is the expected outcome?
-
-Enable richer text UI applications with both key and mouse input and improved user experience.
+The purpose of this proposal is to improve terminal application capabilities when implementing text-based UIs to enable richer text UI applications with both key and mouse input and improved user experience.
+The proposal add the following funcationality to the existing `term` package:
+* mouse input handling), 
+* screen switching between normal and alternate buffer [^1] 
+* cursor hiding and 
+* capturing missing control keys.
 
 ## Detailed design
-
-> This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with the language to understand, and for somebody familiar with the compiler to implement. This should get into specifics and corner-cases, and include examples of how the feature is used.
 
 ### New `ANSINotify` methods
 
@@ -150,12 +125,9 @@ Via changes to the package's generated documentation.
 
 ## How We Test This
 
-> How do we assure that the initial implementation works? How do we assure going forward that the new functionality works after people make changes? Do we need unit tests? Something more sophisticated? What's the scope of testing? Does this change impact the testing of other parts of Pony? Is our standard CI coverage sufficient to test this change? Is manual intervention required?
+Most of this proposal is additive. The one area where code is modified is the input stream processing; yet even then the changes there are additive in that a new character is matches when in `_EscapeCSI` state after which all handling is via new states.
 
-_Currently there is an example for `ANSITerm` and `Readline`; otherwise there are no tests in the repo._
-
-* Add a new example that surfaces the new functionality to enable mouse input, screen switching.
-* Any suggestions?
+This will require manual testing to ensure that (a) the new functionality works as advertised and (b) the changes to not adversely affect the existing functionality. 
 
 ## Drawbacks
 
@@ -165,13 +137,7 @@ None other than that this introduces new, albeit a small amount of, changes that
 
 ## Alternatives
 
-> What other designs have been considered? 
-
-This extends the extending design in the `term` package.
-
-> What is the impact of not doing this?
-
-Richer terminal apps will need to rely on implementing their own terminal input handling. 
+As an alternative, instead of extending `term` someone could directly integrate with `ncurses`, for example, to implement rich terminal UIs.
 
 ## Reference Implementation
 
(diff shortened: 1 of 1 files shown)


stefandd:
The functionality shown in the gif -- which is really nice -- can this be achieved uniformly across all Pony-supported platforms (Linux, MacOs, Windows, etc)? If this cannot likely be accomplished, then term should maybe remain *as is* and only cover the subset of functionality that can be achieved across platforms.

text/0000-stdlib-term-mouse-input-plus.md
```diff
@@ -0,0 +1,196 @@
...
+### New `ANSINotify` methods
+
+`fun ref mouse_down(button: U8, x: U32, y: U32)`
+* This method is called whenever a mouse button is pressed
+* `button` is either 0 (left), 1 (middle), 2 (right)
+* `x` and `y` are the character coordinates below the mouse cursor
+
+`fun ref mouse_up(button: U8, x: U32, y: U32) => None` 
+* This method is called whenever a mouse button is released
+* `button` is either 0 (left), 1 (middle), 2 (right)
```
rhagenson, line 30: Can we make button be enumerated primitives that use 0/1/2 underneath? See https://tutorial.ponylang.io/types/type-aliases.html#enumerations if you do not know what I am suggesting.

text/0000-stdlib-term-mouse-input-plus.md
```diff
@@ -0,0 +1,196 @@
...
+* `x` and `y` are the character coordinates below the mouse cursor
+
+`fun ref mouse_drag(button: U8, x: U32, y: U32) => None`
+* This method is called whenever the mouse is moved while a button is pressed
+* `button` is either 0 (left), 1 (middle), 2 (right)
+* `x` and `y` are the character coordinates below the mouse cursor
+
+`fun ref mouse_wheel(direction: U8, x: U32, y: U32) => None`
+* This method is called whenever the mouse wheel is rolled in either direction. 
+* `direction` is either 0 or 1
```
rhagenson, line 44: Same point about enumerated primitives.

text/0000-stdlib-term-mouse-input-plus.md
```diff
@@ -0,0 +1,196 @@
...
+The proposed changes have been implemented by making a copy of the `term` package and creating a temporary derivative with the enhancements. When/if the RFC is approved, I will be able to supply the implementation via a PR.
+
+## Questions
+
+> Some possible questions and draft answers on which I am open to be convinced otherwise. :)
+
+1. ✅ Should the mouse coordinates use `U32` to match the coordinate types used by `ANSI.cursor()` method?
+     * I was surprised to see `U32` used for text cursor coordinates as the terminal codes barely support 16-bit coordinate values. 
+     * Nonetheless, I think the answer here is YES to avoid unnecessary conversions across methods for text coordinates.
+2. ❓ Should on/off method pairs instead be defined as a single method with a boolean parameter? E.g. `cursor_visibility(show: Bool)` instead of `cursor_hide()/cursor_show()`
```
rhagenson: I prefer the separate methods.

rhagenson (commented):
I offered a quick review while on a much needed coffee break. I am certain I will have more to say later, but am excited to see these additions being suggested!

SeanTAllen:
The prompts in the template aren't meant to be a "please answer these as Q&A". Can you rewrite this to not have the prompt of Q & A in the format?

We went over this for a moment at sync but no one had read yet and we moved on. 

Folks said they would try and take a look soon. There was a small amount of discussion on the topic of "do we have a rule to only include things in the stdlib that work on all platforms"- answer no, but, it depends. See audio for more details.

text/0000-stdlib-term-mouse-input-plus.md
```diff
@@ -0,0 +1,196 @@
...
+    - [Mouse input handling](#mouse-input-handling)
+    - [Ctrl-C and Ctrl-Z input handling](#ctrl-c-and-ctrl-z-input-handling)
+  - [ANSI terminal options](#ansi-terminal-options)
+- [How We Teach This](#how-we-teach-this)
+- [How We Test This](#how-we-test-this)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Reference Implementation](#reference-implementation)
+- [Questions](#questions)
+- [References](#references)
```
SeanTAllen: you can remove this. It isn't really needed.

SeanTAllen:
How much of this is "mouse input" vs "other enhancements"? If they aren't tied to one another, I'd like to see separate RFCs. Note, this is merely prompted by the title. I haven't had time to read.

nogginly:
@stefandd FYI

> The functionality shown in the gif -- which is really nice -- can this be achieved uniformly across all Pony-supported platforms (Linux, MacOs, Windows, etc)? If this cannot likely be accomplished, then term should maybe remain _as is_ and only cover the subset of functionality that can be achieved across platforms.

I've tested this using macOS `Terminal` app and therefore I believe this will work as shown in Linux terminals.

However I do not know about Windows. From what I've heard, it "should" work via WSL2 terminal, but I don't think it will if run in a DOS or PowerShell window. Either way, I don't have evidence on Windows.

text/0000-stdlib-term-mouse-input-plus.md
nogginly: @rhagenson, while this proposal addresses the three mouse buttons, in the Xterm specs it's is possible for the button number to reflect more than left, middle, right (because of advanced mice with title buttons etc). My thinking was that for now I'll limit it to the known three button numbers, but over time if we handle additional buttons then naming will become hard as the high buttons don't always mean the same thing across fancy mice. 😄  

I don't like to "predict the future" often; in this case since the API maps underlying xterm/ANSI concepts 1 for 1, I propose using numeric values.

text/0000-stdlib-term-mouse-input-plus.md
nogginly: @rhagenson, ideally direction is UP or DOWN, but given that we now have two kinds of up/down because of "natural" vs what we had before swiping happened, up or down is relative. Unlike the buttons, in this case there are only two values so we could define `Up` and `Down` ... though those might be too common to put in the global namespace. Do you have a recommendation for the names?

text/0000-stdlib-term-mouse-input-plus.md
nogginly: Me too. Thanks.

text/0000-stdlib-term-mouse-input-plus.md
nogginly: Will do.

SeanTAllen:
@nogginly  please attempt to recruit testers on the various platforms understanding what works on what platforms would be an important part of accepting the RFC. 

I suspect there's no automated tests for some of this functionality and that detailed manual testing should be part of the acceptance criteria detailed in this PR.

Is there anything in this RFC that precludes this starting as a library that could be enhanced over time and various questions of testing and platform support could be answered there?

text/0000-stdlib-term-mouse-input-plus.md
SeanTAllen: What do you mean by global namespace?

nogginly:
@SeanTAllen 

> How much of this is "mouse input" vs "other enhancements"? If they aren't tied to one another, I'd like to see separate RFCs. Note, this is merely prompted by the title. I haven't had time to read.

Good point, this originally started out as adding mouse support and that led to switching screens and the other things I've listed. If we decide to split it, I'd recommend keeping screen switching and mouse support together since I think one without the other, which technically possible, is not super useful.

nogginly:
> please attempt to recruit testers on the various platforms understanding what works on what platforms would be an important part of accepting the RFC.

I'll setup my reference implementation in a public repo and post that here and Zulip to get some help.

> I suspect there's no automated tests for some of this functionality and that detailed manual testing should be part of the acceptance criteria detailed in this PR.

Yes. I'll update the proposal section on testing to reflect that.

nogginly:
> Is there anything in this RFC that precludes this starting as a library that could be enhanced over time and various questions of testing and platform support could be answered there?

@SeanTAllen, by "library" do you mean independent  of `stdlib` packages that someone could pull in and use via `corral` for example?

In fact I was originally surprised to find `term` in the standard library, but of course once I wanted to do more it seemed beneficial to enhance it. If however you think it would be better to publish a derivative of `term` with enhancements, I am open to that as well. I can start with the contents of `term` and then add to it while keeping it backwards compatible so anyone who starts with `term` could switch to a library.

SeanTAllen:
@nogginly yes, I mean it's own library you install via corral. 

Given the question of support on various platforms saved difficulty in providing automated tests, "seasoning" as a library outside the stdlib might make sense for things like the mouse support.

text/0000-stdlib-term-mouse-input-plus.md
nogginly: I just meant that after a `use "term"` statement, any primitives defined within would become accessible to the rest of the code in that file, and thus a `primitive Up fun apply() : U8 => 1` would put the name `Up` into the namespace. (It's how I think about it, though in hindsight "global" is likely not the right word since I assume that's reserved for the built-in objects.)

SeanTAllen: Don't take that into account when naming things. You can always alias names when importing so using a common name shouldn't be something you consider.

nogginly:
Ha! I really like how you call it  "seasoning" [a standard package] as a library. My wife's a chef and every time she used that word it never occured to me to make that conceptual leap and apply it to derived software.

@SeanTAllen, some questions:

1. Any concerns with me using the contents of `term` as the start for such a new library?
2. Regarding the license, is it sufficient that I start with the one in `ponyc` and add the additional copyright to the existing two statements?
3. Any other considerations (e.g. attributions) when/if I do this?

SeanTAllen:
1. No concerns 

2. Definitely no concerns of keeping the same license and maintaining copyright

3. Nothing pops to mind.

nogginly:
Sounds to me like we have a decision ... I'm happy to take this proposal and work on an external library with the enhancements. Thanks for taking the time to go through the proposal and providing feedback. 

Closing this PR.

PR #203, opened 2022-10-26
Add RFC for the inclusion of all erase codes in the Term package

SeanTAllen:


text/0000-ansi-erase.md
```diff
@@ -0,0 +1,84 @@
...
+
+Three new public primitives will be created in the `term` package that will be
+used to control which direction `erase` erases aka, what code it will return.
+
+```pony
+primitive EraseLeft
+primitive EraseLine
+primitive EraseRight
+
+type _EraseDirection is (EraseLeft | EraseLine | EraseRight)
```
jemc, line 28: What's the motivation for making this a private type alias?

SeanTAllen: To make reading the erase method a little easier so that there isn't the full union listed there but also not exposing this type outside of the package if someone does `use "term"`. That is all.

I'd be fine with getting rid of the alias and having the full union in the method definition, but, it makes it a little easier to read this way.

rhagenson (commented):
I like the approach taken with 3 primitives and 1 method. The alternatives all strike me as strange.

Not explicitly "approving" in this review because I have not looked at the ANSI term functionality we are adding/aligning ourselves with, only at the suggested changes here.

SeanTAllen:
@rhagenson @jemc I have a better implementation idea. Same shape of things but without the need for the match statement.

I prefer this and think it is easier to follow.

Thoughts?

```pony
primitive EraseLeft
  fun apply(): String =>
    "\x1B[1K"

primitive EraseLine
  fun apply(): String =>
    "\x1B[2K"

primitive EraseRight
  fun apply(): String =>
    "\x1B[0K"

type _EraseDirection is (EraseLeft | EraseLine | EraseRight)

primitive Ansi
  fun erase(direction: _EraseDirection = EraseRight): String =>
    """
    Erases content. The direction to erase is dictated by the `direction`
    parameter. Use `EraseLeft` to erase everything from the cursor to the
    beginning of the line. Use `EraseLine` to erase the entire line. Use
    `EraseRight` to erase everything from the cursor to the end of the line.
    The default direction is `EraseRight`.
    """
    direction()
```

rhagenson:
I like that a lot more. I was trying to think of a way to get rid of the match because I did not like one condition being the fall through in else when we know all the states that are possible.

SeanTAllen:
@rhagenson yeah. I didn't like it either. I dashed it off yesterday to get it open.

SeanTAllen:
I've updated again because "dur", exhaustive match, the else wasn't needed. I think I like the version in the RFC better than the one in my comment now. It fits the existing API better.

rhagenson (approved):
Using the exhaustive match is clean. I do not see anymore necessary changes. I looked up the ANSI standard to see if there are any further "holes" in this method that we need to support, but right/left/line is exhaustive for erase on a single line.

I am indifferent on Joe's point of a private alias versus the union being inlined on the method. I generally lean towards the idea of not creating more names, but in this case I think the union makes it clearer to any reader that the three options are exhaustive.

chalcolith (approved):

jemc (approved):

text/0000-ansi-erase.md
```diff
@@ -0,0 +1,84 @@
...
+
+# How We Teach This
+
+The only teaching will be in the method documentation for `erase` on the `Ansi`
+primitive. The current documentation will be updated to reflect the changes
+detailed in this RFC.
+
+# How We Test This
+
+Just about every approach we as an option is a very simple implementation where
```
ergl: Missing word here

ergl (approved):
I like the version in the RFC text more, too. Also, how does the generated documentation look like if the parameter to `erase` is a private type alias? Does it show all the different types?

SeanTAllen:
@ergl Type aliases dont appear in any generated documentation.

jemc:
Approved during today's sync call :heavy_check_mark:

PR #216, opened 2024-12-10
`String.find`/`String.rfind` `USize` proposal

hovsater:
The current return type of `String.find` and `String.rfind` is `ISize`, which stems from its historical behavior of returning `-1` when no match was found. However, the methods now return `error` for such cases. Returning `USize` makes more sense because indices are always non-negative, and it aligns with `Array.find`'s method signature.

### DIFF
--- a/text/0000-usize-indexing
+++ b/text/0000-usize-indexing
@@ -1,0 +1,50 @@
+- Feature Name: string_find_usize
+- Start Date: 2024-12-09
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Change the return type of `String.find`  and `String.rfind` from `ISize` to `USize`, aligning its return type with `Array.find`.
+
+# Motivation
+
+The current return type of `String.find` and `String.rfind` is `ISize`, which stems from its historical behavior of returning `-1` when no match was found. However, these methods now return `error` for such cases. Returning `USize` makes more sense because indices are always non-negative, and it aligns with `Array.find`'s method signature.
+
+# Detailed design
+
+The method signature of [String.find](https://github.com/ponylang/ponyc/blob/736220a34364a864fe0fd1f091a85852ded84d23/packages/builtin/string.pony#L640) and [String.rfind](https://github.com/ponylang/ponyc/blob/736220a34364a864fe0fd1f091a85852ded84d23/packages/builtin/string.pony#L669) would change to return `USize` instead of `ISize`.
+
+Specifically, the method signatures would be changed from this:
+
+```pony
+fun find(s: String box, offset: ISize = 0, nth: USize = 0): ISize ?
+fun rfind(s: String box, offset: ISize = -1, nth: USize = 0): ISize ?
+```
+
+to this:
+
+```pony
+fun find(s: String box, offset: ISize = 0, nth: USize = 0): USize ?
+fun rfind(s: String box, offset: ISize = -1, nth: USize = 0): USize ?
+```
+
+# How We Teach This
+
+N/A
+
+# How We Test This
+
+Standard CI and the compiler should identify most areas that need to be addressed. Additionally, any documentation, code examples, and guides should be reviewed and updated to reflect the new behavior.
+
+# Drawbacks
+
+It's a breaking change as the method signature has to change. Existing code relying on `String.find` and `String.rfind` will need to be migrated.
+
+# Alternatives
+
+Keep the existing method signature to avoid breaking changes introduced by this proposal.
+
+# Unresolved questions
+
+N/A
--- a/text/0000-usize-indexing
+++ b/text/0000-usize-indexing
@@ -35,7 +35,7 @@ Through the standard library documentation and the associated release notes.
 
 # How We Test This
 
-Standard CI and the compiler should identify most areas that need to be addressed. Additionally, any documentation, code examples, and guides should be reviewed and updated to reflect the new behavior.
+Standard CI and the compiler should identify any additional code areas that are needed. Sean T. Allen did a review of the tutorial, pony patterns, and the website and found no changes that would be needed to reflect the new behavior.
 
 # Drawbacks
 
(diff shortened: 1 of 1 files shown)


jemc (approved):
Agreed!

SeanTAllen:
@hovsater can you rebase this against main?

hovsater:
> @hovsater can you rebase this against main?

Done!

text/0000-usize-indexing
```diff
@@ -0,0 +1,50 @@
...
+to this:
+
+```pony
+fun find(s: String box, offset: ISize = 0, nth: USize = 0): USize ?
+fun rfind(s: String box, offset: ISize = -1, nth: USize = 0): USize ?
+```
+
+# How We Teach This
+
+N/A
```
SeanTAllen: We are teaching this in the same way we are currently teaching. Via the standard library documentation. This N/A should be changed. It is applicable and there is a means that we are already teaching such things. 

Small change but important for not setting an incorrect precedent going forward.

text/0000-usize-indexing
```diff
@@ -0,0 +1,50 @@
...
+
+# How We Test This
+
+Standard CI and the compiler should identify most areas that need to be addressed. Additionally, any documentation, code examples, and guides should be reviewed and updated to reflect the new behavior.
```
SeanTAllen: Such a review should happen before this RFC is accepted as fixing anything that this breaks that isn't programmatically caught should be part of the acceptance criteria for merging the change to the return type.

text/0000-usize-indexing
```diff
@@ -0,0 +1,50 @@
...
+
+It's a breaking change as the method signature has to change. Existing code relying on `String.find` and `String.rfind` will need to be migrated.
+
+# Alternatives
+
+Keep the existing method signature to avoid breaking changes introduced by this proposal.
+
+# Unresolved questions
+
+N/A
```
SeanTAllen: This is applicable. 

Can you change this to "none at this time". I believe that is what you meant.

However, given the state of this RFC, unresolved at this time is anything that needs to be updated that this might break in documentation.

SeanTAllen: Another thought. Given that this will break existing code that might require changes in other ponylang org projects, I'd like to have encoded into this RFC or at least an agreement amongst committers that we will not merge this change without an assurance from the author of the PR that implements this to quickly address all breakages in ponylang org libraries that our nightly breakage tests identify (if any exist).

I've often been left to pick up that important piece and I'd like to have a discussion that about how we address that isn't solely "Sean labor".

text/0000-usize-indexing
SeanTAllen: There's no usage in the tutorial that I can find.

SeanTAllen: There's no usage in pony-patterns that I can find.

SeanTAllen: The only usage I can find on the website is a reference to this RFC from the most recent LWIP.

text/0000-usize-indexing
SeanTAllen: I think this could be that we already teach the usage of rfind and find via the method documentation and that nothing needs to change regarding that. And that otherwise, inclusion in the release notes plus examples of how to update code where a type annotation is given and the lack of a change if type implication is used would suffice here.

We do however, I think want release notes that detail the change, that it is breaking and how to fix it if the break impacts on you. The fix in this case might be "obvious" but I think it behooves us to practice giving explanations for fixing any breaking changes we introduce.

SeanTAllen:
I'm in favor of this RFC with the changes I have requested.

jemc:
The next few sync calls where we would vote on this are cancelled due to holidays, so we decided to approve this RFC today, conditional on Sean's concerns above being addressed to his satisfaction. So we don't need to wait for a future vote - this can get merged as soon as those changes have come in and been approved by Sean.

text/0000-usize-indexing
hovsater: So what should I write here instead, @SeanTAllen?

text/0000-usize-indexing
hovsater: Thanks for the feedback. I've addressed that in https://github.com/ponylang/rfcs/pull/216/commits/674631057dcd605563dbbc1156460b02064dc97d.

text/0000-usize-indexing
hovsater: > This is applicable.
> 
> Can you change this to "none at this time". I believe that is what you meant.
> 
> However, given the state of this RFC, unresolved at this time is anything that needs to be updated that this might break in documentation.

See https://github.com/ponylang/rfcs/pull/216/commits/674631057dcd605563dbbc1156460b02064dc97d.

text/0000-usize-indexing
SeanTAllen: That a review has been done so the tutorial, paterns, and website and that there's no changes needed to reflect the new behavior.

SeanTAllen:
I just noticed the fixups that kevin did on this. He didn't ping me so I didn't notice. I'll do the book keeping for this and merge it soon-ish.

hovsater:
I can manage the fixups if we're happy with them. I tend to push fix ups first, and rebase after approval. I'll get to it this evening. 🙂

text/0000-usize-indexing
```diff
@@ -0,0 +1,50 @@
...
+fun rfind(s: String box, offset: ISize = -1, nth: USize = 0): USize ?
+```
+
+# How We Teach This
+
+Through the standard library documentation and the associated release notes.
+
+# How We Test This
+
+Standard CI and the compiler should identify most areas that need to be addressed. Additionally, any documentation, code examples, and guides should be reviewed and updated to reflect the new behavior.
```
SeanTAllen:
```suggestion
Standard CI and the compiler should identify any additional code areas that are needed. Sean T. Allen did a review of the tutorial, pony patterns, and the website and found no changes that would be needed to reflect the new behavior.
```

PR #218, opened 2025-09-10
Updating String.copy_cstring() and String.copy_cpointer() to return iso^

redvers:
The act of writing this RFC made me question my life choices.

I'm now leaning more towards just adding an additional constructor (from the Alternatives section)

```pony
new iso copy_cstring_iso(ptr: Pointer[U8] val)
```

An RFC for discussion!

### DIFF
--- a/text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
+++ b/text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
@@ -1,0 +1,111 @@
+- Feature Name: Modify String.copy\_cpointer() and String.copy\_cstring() to return iso^
+- Start Date: 2025-09-10
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
+The two current constructors `String.copy_cpointer()` and `String.copy_cstring()` both take a `Pointer[U8] box` and return a `String ref^`. Given that these two constructors make copies, returning a `String iso^` would be preferable and safe.
+
+Note: In the interests of clarity of writing, I will use `String.copy_cstring()` throughout the rest of this RFC. All of the commentary below should be read to apply to both `String.copy_cstring()`, and `String.copy_cpointer()`.
+
+# Motivation
+
+In many cases, as a library author I want to provide my end-users a mutable and sendable copy of a String received via C-FFI. In other words, a `String iso^`. With the way that the String library is currently written, there are two ways to do this:
+
+## Create a String ref^ using String.from\_cstring() or from\_cpointer() and clone().
+
+```pony
+  var str: String iso = String.from_cstring(ptr).clone()
+```
+
+The main issue with this is that it directly violates the documentation which states: "This must be done only with C-FFI functions that return pony\_alloc'd character arrays".
+
+## Create a String ref^ using String.copy\_cstring() and clone()
+
+```pony
+  var str: String iso = String.copy_cstring(ptr).clone()
+```
+
+This does not violate the documentation and does a clean copy into a new `String ref^`. Unfortunately, the clone() that follows to generate the `String iso^` we need results in a second copying of the data.
+
+Since the `String.copy_cstring()` does a clean copy, it would be a safe operation to return a `String iso^`, making the second copying redundant.
+
+Ideally, the change would result in the following use:
+
+```pony
+  var str: String iso = String.copy_cstring(ptr)
+```
+
+# Detailed design
+
+The existing `String.copy_cstring()` and `String.copy_cpointer()` have the following signatures:
+
+```pony
+new ref copy_cstring(str: Pointer[U8 val] box)
+new ref copy_cpointer(str: Pointer[U8 val] box, len: USize val)
+```
+
+The reason that we cannot just change the return type to `String iso^` is that `str: Pointer[U8] box` is not sendable. The ideal outcome would be to not change the `Pointer[U8] box` to `Pointer[U8] val` in order to not create a breaking change.  Here are two proposed ways to achieve this:
+
+## Convert from constructors to functions
+
+The main disadvantage to this approach is that it does cause an additional creation of an (empty) `String` in addition to the `String iso^` it would return. That feels "impure"™. An implementation might look something like this:
+
+```pony
+  fun copy_cpointer(ptr: Pointer[U8] box, len: USize): String iso^ =>
+    """
+    Create a string by copying a fixed number of bytes from a pointer.
+    """
+    let str: String iso = recover iso String(len + 1) end
+    if not ptr.is_null() then
+      ptr._copy_to(str._ptr._unsafe(), len)
+      str._set(len, 0)
+      str.recalc()
+    end
+    consume str
+```
+
+## Retain the constructor, but change the receiver to Pointer[U8] tag
+
+```pony
+new iso copy_cstring(str: Pointer[U8 val] tag)
+```
+
+As there is a copying mechanism this doesn't break safety, but it does break the expectation that you cannot get data out of a `Pointer[U8] tag`.
+
+# How We Teach This
+
+The aim of this RFC is for there to be no user impact, so no new information will need to be shared.
+
+# How We Test This
+
+There are currently no tests for `String.copy_cstring()` or `String.copy_cpointer()` in builtin_tests. My initial PR included some.
+
+# Drawbacks
+
+The first proposal doesn't feel "correct", creating an object via a function, and not using the calling object.
+
+The second proposal although not breaking safety, does break expectations that have been set over time.
+
+# Alternatives
+
+## Change the signature to only accept `Pointer[U8] val`
+
+This would be a trivial but breaking change. A review of stdlib and the libraries in the github ponylang repositories finds no clear cases where I think this would break.
+
+```pony
+new iso copy_cstring(str: Pointer[U8 val] val)
+```
+
+## Leave copy_cstring() as is, add copy_cstring_iso()
+
+This would be a trivial non-breaking change, and there is existing precedent for including refcaps in function names: `String.from_iso_array()`.
+
+```pony
+new iso copy_cstring_iso(str: Pointer[U8 val] val)
+```
+
+# Unresolved questions
+
+Although the ideal of not making a breaking change is strong, it may result in being the most correct way forward.


text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+In many cases, as a library author I want to provide my end-users a mutable and sendable copy of a String received via C-FFI. In other words, a `String iso^`. With the way that the String library is currently written, there are two ways to do this:
+
+## Create a String ref^ using String.from\_cstring() or from\_cpointer() and clone().
+
+```pony
+  var str: String iso = String.from_cstring(ptr).clone()
+```
+
+The main issue with this is that it directly violates the documentation which states: "This must be done only with C-FFI functions that return pony\_alloc'd character arrays".
```
SeanTAllen, line 22: side note: that documentation should probably be updated as if it doesn't state a reason. to be safe, it must be assured that the allocator of the memory won't free it for as long as we hold on to it.

what should really be said is that it should only be done with a pointer to memory that we can take ownership of.

SeanTAllen, line 22: I disagree with this being "the main issue". It is important to highlight taht the pattern you have requires two string allocations. That is wasteful, especially if you do it a lot.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Create a String ref^ using String.copy\_cstring() and clone()
```
SeanTAllen, line 24: again, here you are doing a clone and that is wasteful. You create a string solely to clone it to get a new string. eeewww. unclean! unclean!

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+```pony
+  var str: String iso = String.copy_cstring(ptr).clone()
+```
+
+This does not violate the documentation and does a clean copy into a new `String ref^`. Unfortunately, the clone() that follows to generate the `String iso^` we need results in a second copying of the data.
```
SeanTAllen, line 30: as i note above, it worse than the copying of the data.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
redvers: Agreed - this is the motivation for the change…  To avoid this second copy.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+    Create a string by copying a fixed number of bytes from a pointer.
+    """
+    let str: String iso = recover iso String(len + 1) end
+    if not ptr.is_null() then
+      ptr._copy_to(str._ptr._unsafe(), len)
+      str._set(len, 0)
+      str.recalc()
+    end
+    consume str
+```
```
SeanTAllen, line 67: this is no better what you are already having to do AND it adds extra allocations for everyone. do not like.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Retain the constructor, but change the receiver to Pointer[U8] tag
+
+```pony
+new iso copy_cstring(str: Pointer[U8 val] tag)
+```
+
+As there is a copying mechanism this doesn't break safety, but it does break the expectation that you cannot get data out of a `Pointer[U8] tag`.
```
SeanTAllen, line 75: what expectation that you can't get data out of a Pointer[U8] tag. `tag` is identity. If I have the identity of a pointer, I can copy it. It is why Pointer is only available to manipulate from inside `builtin`.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+# How We Teach This
+
+The aim of this RFC is for there to be no user impact, so no new information will need to be shared.
```
SeanTAllen, line 79: If this could break existing code, it needs to be taught in the sense of we must inform people this is happening. At minimum that is in release notes, perhaps more.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+# How We Test This
+
+There are currently no tests for `String.copy_cstring()` or `String.copy_cpointer()` in builtin_tests. My initial PR included some.
+
+# Drawbacks
+
+The first proposal doesn't feel "correct", creating an object via a function, and not using the calling object.
```
SeanTAllen, line 87: and it adds a tax on everyone even those who are fine with the existing functionality.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+The second proposal although not breaking safety, does break expectations that have been set over time.
```
SeanTAllen, line 89: please show where these expectations have been set.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Alternatives
+
+## Change the signature to only accept `Pointer[U8] val`
```
SeanTAllen, line 93: I dont see any reason to make a breaking change of this sort when the same can be accomplished with a new constructor.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+This would be a trivial but breaking change. A review of stdlib and the libraries in the github ponylang repositories finds no clear cases where I think this would break.
```
SeanTAllen, line 95: trivial can mean a lot of things. i would not be in favor of an RFC that that's this is trivial without defining what is trivial about it.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+This would be a trivial non-breaking change, and there is existing precedent for including refcaps in function names: `String.from_iso_array()`.
+
+```pony
+new iso copy_cstring_iso(str: Pointer[U8 val] val)
+```
+
+# Unresolved questions
+
+Although the ideal of not making a breaking change is strong, it may result in being the most correct way forward.
```
SeanTAllen, line 111: correct is a loaded term. i would not be in favor of an rfc that uses that language without detailing exactly what is meant by "correct".

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
redvers: I think my expectation came from the fact that there is no (current) way to get the content of a Pointer[U8] tag without resorting to C-FFI because all the functions to create Array[U8] or Strings require box/ref/val.

So, creating the first constructor which allowed you to create a String from a Pointer[U8] tag would be a change in expectation.

It comes down to the motivation as to why these constructors don't exist.

Is it *only* to prevent creation of other objects that use the same underlying Pointers (thus breaking all guarantees…), or was there something more?

I don't know - so I err'd on the side of caution.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
```diff
@@ -0,0 +1,111 @@
...
+
+As there is a copying mechanism this doesn't break safety, but it does break the expectation that you cannot get data out of a `Pointer[U8] tag`.
+
+# How We Teach This
+
+The aim of this RFC is for there to be no user impact, so no new information will need to be shared.
+
+# How We Test This
+
+There are currently no tests for `String.copy_cstring()` or `String.copy_cpointer()` in builtin_tests. My initial PR included some.
```
SeanTAllen, line 83: this isnt sufficient for an RFC. If we get close to accepting a version of the RFC, then there needs to be detailed information about how we test. This section must be fulfilled for a PR implementing the RFC to be merged so a sufficiently detailed answer will be required here.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
redvers: … and define what "we can take ownership of" means.

I always assumed (I know) that meant that when the Pointer[U8] refcount reached zero that ponyrt would attempt to free it via GC. So if not that, it means more that the library author wrote "You are responsible for calling free() on this allocation." or equivalent?

redvers: Yup - the double allocation is motivation for the change.

SeanTAllen: Sorry what I meant was "won't be freed while we hold a pointer to". Freeing it is an entirely different issue.  

If it is pony allocated then we know it won't be freed under us.

redvers:
In my original comment / commit I highlighted that my preferred solution was the addition of a set of `create_c*_iso` constructors but pushed I still pushed the RFC as written to start a discussion.

The more I look at the problem now, the more I think that even the addition of this function is unnecessary as the existing functions can be coaxed to return an iso as long as you provide a `Pointer[U8] val` - my proposed change.

I think I focused too hard on the documented function signature which, in my reading meant that it could *only* provide a ref^.

My understanding missed the fact that the documentation cites the most permissive refcaps for inputs and the most restrictive on outputs.

I think I'll write-up my thoughts on this subtlety later today on Zulip and take a look through the documentation to see if I can work out where I missed this subtle detail.

The RFC was a very useful process for me in getting my thoughts on the subject to ground, but I no longer believe this is a useful RFC to pursue.

text/0000-string-copy_cpointer-copy_cstring-returns-iso.md
redvers: Found it - this (from the tutorial).

[This one is easy: tag variables are opaque! They can’t be read from.](https://tutorial.ponylang.io/reference-capabilities/combining-capabilities.html?h=tag#reading-from-a-tag-variable)

So in my mind, if I could pass a Pointer[U8] tag into a function and get a _copy_ of that string, it counted in my mind as "reading" from it.

PR #220, opened 2026-02-28
Redesign signal handling

### LINKED ISSUE #170: When Multiple Signal Handlers Register, They Step on One another
If you try to register signal handlers multiple times for the same signal, the last one wins. This can lead to undefined behaviour if libraries use signal handlers.

SeanTAllen:
Proposes redesigning signal handling to fix the inconsistent cross-platform behavior (first-handler-wins on Linux vs last-handler-wins on macOS) and add capability security via a new SignalAuth type. The new design supports multiple subscribers per signal, with the runtime dispatching to all registered handlers in undefined order. SignalRaise also gains an auth requirement for consistency with other I/O primitives.

Closes #170

### DIFF
--- a/text/0000-redesign-signal-handling.md
+++ b/text/0000-redesign-signal-handling.md
@@ -60,13 +60,12 @@ actor SignalHandler is AsioEventNotify
     Create a signal handler.
     """
 
-  be raise() =>
+  be raise(auth: SignalAuth) =>
     """
-    Raise the signal. No additional auth is required because the
-    capability was already verified when this handler was created.
+    Raise the signal.
     """
 
-  be dispose() =>
+  be dispose(auth: SignalAuth) =>
     """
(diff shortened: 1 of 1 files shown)


jemc (commented):
I think the design thinking here is good, the only open question is if the runtime architecture changes will work as described or if they'll need to make different choices or end up exposing different properties

SeanTAllen:
Here's an idea of what this would look like: https://github.com/ponylang/ponyc/pull/4984/changes

jemc:
API design seems fine - no issues from me.

As discussed in sync today, we will need a careful, detailed round of review and testing of the runtime code that involves atomics.

SeanTAllen:
Rename ValidSignal and related to HandleableSignal. In case we add sending of signals in the future.

PR #223, opened 2026-03-13
Json.print() api in stdlib json package

mfelsche:
[Rendered](https://github.com/ponylang/rfcs/blob/c67abf7aafd7010eb733b65a08408f27fa21e20a/text/0000-json-encode-jsonvalue.md)

### DIFF
--- a/text/0000-json-encode-jsonvalue.md
+++ b/text/0000-json-encode-jsonvalue.md
@@ -1,0 +1,75 @@
+- Feature Name: json-encode-jsonvalue
+- Start Date: 2026-03-12
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Adapt the stdlib [`json` package][json-package] in order to add means to encode arbitrary [`JsonValue`][JsonValue]s into `String` or `Array[U8]`. And while we are at it, provide a matching API for decode `String` or `Array[U8]` into [`JsonValue`] for completeness.
+
+# Motivation
+
+The basic functionality of a json library should be to provide a representation for structured json values (which is [JsonValue][JsonValue]) and methods to decode a string or bytes as JSON into this [JsonValue][JsonValue] and vice versa, to encode all possible representations of [JsonValue][JsonValue] into valid JSON bytes (or string). The current [`json` package][json-package] is lacking the means of encoding all [JsonValue][JsonValue] instances into valid JSON.
+
+`JsonObject` and `JsonArray` provide methods for encoding them as valid JSON with the `.string()` and `.pretty_string()` methods, but for all other possible JSON values (string, number, bool, null) there is no such method. The `.string()` method on `None` is producing `"None"` (without the quotes), which is invalid JSON. Another example: the String `"\""` is copying itself when `.string()` is called on it, without any escaping, which is also invalid JSON. The method `.pretty_string()` is missing from all those other [JsonValue][JsonValue] instances.
+
+This package needs a consistent and convenient way to encode and decode json bytes from and to all possible [JsonValue][JsonValue] instances.
+
+# Detailed design
+
+This PR suggests adding a new primitive called `Json`:
+
+```pony
+primitive Json
+  fun print(value: JsonValue, pretty: Bool = false): String =>
+    if pretty then
+      _JsonPrint.pretty(value)
+    else
+      _JsonPrint.compact(value)
+    end
+
+  fun parse(source: String): (JsonValue | JsonParseError) =>
+    JsonParser.parse(source)
+```
+
+All the bits and pieces for encoding arbitrary values are there already, they just aren't exposed. `_JsonPrint` is private.
+
+The naming of the functions is not settled yet. The dual of `encode` and `decode` would be ideal, but the `JsonParser` already returns `JsonParseError`, hence `parse` makes more sense for the decoding step, but it lacks a proper dual name for the process of decoding. For further discussion of the naming, see [Alternatives](#alternatives).
+
+Proper docstrings are missing, obviously. They should explain what can be expected, under which conditions errors are raised etc. and also list usage examples. They were left out for brevity.
+
+We might want to remove the `JsonObject.string()` and `JsonArray.string()` (and `.string_pretty()`) functions. As all [JsonValue][JsonValue] members implement `ToString` and thus all have a `.string()` method, but don't all produce valid JSON, this might be a source of confusion. This would also shrink the amount of different ways to achieve one thing to one: `Json.print()`.
+
+The `JsonParser` primitive should also be removed and the contents of its `parse()` method should be moved to the `Json.parse()` method, to have only one way of decoding a string as JSON.
+
+# How We Teach This
+
+The package documentation of the `json` package should show usage examples of these functions most prominently and then show all other components of the crate, like `JsonValue`, `JsonPath` or `JsonNav`.
+
+The intend of this primitive is to expose an easy-to-use interface for turning strings into structures JSON values and vice versa. This should answer the question of how to get from bytes or strings from the network or from files to structured JSON that can be analyzed (Pass the string to `Json.parse()`), and how to serialize Pony objects and data-structures as JSON (build a `JsonValue`, then pass it those `Json.print()`).
+
+
+# How We Test This
+
+Additional property-based tests will be added for encoding arbitrary [JsonValue][JsonValue]s via `Json.encode(...)` and parsing them again via `Json.parse(...)` and ensuring equivalence of the initial values and the ones produces by the roundtrip.
+
+# Drawbacks
+
+While exposing a way of printing/encoding/serializing [JsonValue][JsonValue]s is necessary, removing `JsonParser` and `JsonObject.string` etc. might not be. These changes are breaking.
+
+# Alternatives
+
+Function naming alternatives:
+
+- `parse` / `print` - this would follow the existing naming in the package the closest, hence considered as the best candidate.
+- `encode` / `decode` - this pair would make sense if the decoding would not expose `JsonParseError`. It feels inconsistent to create a `ParseError` from a `decode` function call.
+- `serialize` / `deserialize` - these terms are not used anywhere yet in pony, so didn't favor them.
+- `to_string` / `from_string` - The most humble variant of them all, getting along without any fancy terminology.
+
+# Unresolved questions
+
+- Should a partial version of parsing be also exposed?
+- Should `Json.print()` expose finer controls than just the `pretty` flag?
+
+[json-package]: https://github.com/ponylang/ponyc/tree/main/packages/json
+[JsonValue]: https://github.com/ponylang/ponyc/blob/main/packages/json/json.pony#L243
--- a/text/0000-json-encode-jsonvalue.md
+++ b/text/0000-json-encode-jsonvalue.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-Adapt the stdlib [`json` package][json-package] in order to add means to encode arbitrary [`JsonValue`][JsonValue]s into `String` or `Array[U8]`. And while we are at it, provide a matching API for decode `String` or `Array[U8]` into [`JsonValue`] for completeness.
+Adapt the stdlib [`json` package][json-package] in order to add a `JsonPrinter` primitive exposing functions to encode arbitrary [`JsonValue`][JsonValue]s into `String iso^`, in a similar manner as the existing `JsonParser` for parsing [`JsonValue`][JsonValue] from String.
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,75 @@
+- Feature Name: json-encode-jsonvalue
+- Start Date: 2026-03-12
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Adapt the stdlib [`json` package][json-package] in order to add means to encode arbitrary [`JsonValue`][JsonValue]s into `String` or `Array[U8]`. And while we are at it, provide a matching API for decode `String` or `Array[U8]` into [`JsonValue`] for completeness.
```
SeanTAllen: does JsonParser.parse not do what you are looking for with decode?

It takes a String and gives you a JsonValue. I'm unclear why we would wrap that.

text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Motivation
+
+The basic functionality of a json library should be to provide a representation for structured json values (which is [JsonValue][JsonValue]) and methods to decode a string or bytes as JSON into this [JsonValue][JsonValue] and vice versa, to encode all possible representations of [JsonValue][JsonValue] into valid JSON bytes (or string). The current [`json` package][json-package] is lacking the means of encoding all [JsonValue][JsonValue] instances into valid JSON.
+
+`JsonObject` and `JsonArray` provide methods for encoding them as valid JSON with the `.string()` and `.pretty_string()` methods, but for all other possible JSON values (string, number, bool, null) there is no such method. The `.string()` method on `None` is producing `"None"` (without the quotes), which is invalid JSON. Another example: the String `"\""` is copying itself when `.string()` is called on it, without any escaping, which is also invalid JSON. The method `.pretty_string()` is missing from all those other [JsonValue][JsonValue] instances.
+
+This package needs a consistent and convenient way to encode and decode json bytes from and to all possible [JsonValue][JsonValue] instances.
```
SeanTAllen, line 16: I believe the decode already exists. What we are talking about is taking arbitrary JsonValue items and encoding to Json rather than starting from a JsonArray or JsonObject, yes?

text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,75 @@
...
+
+The naming of the functions is not settled yet. The dual of `encode` and `decode` would be ideal, but the `JsonParser` already returns `JsonParseError`, hence `parse` makes more sense for the decoding step, but it lacks a proper dual name for the process of decoding. For further discussion of the naming, see [Alternatives](#alternatives).
+
+Proper docstrings are missing, obviously. They should explain what can be expected, under which conditions errors are raised etc. and also list usage examples. They were left out for brevity.
+
+We might want to remove the `JsonObject.string()` and `JsonArray.string()` (and `.string_pretty()`) functions. As all [JsonValue][JsonValue] members implement `ToString` and thus all have a `.string()` method, but don't all produce valid JSON, this might be a source of confusion. This would also shrink the amount of different ways to achieve one thing to one: `Json.print()`.
+
+The `JsonParser` primitive should also be removed and the contents of its `parse()` method should be moved to the `Json.parse()` method, to have only one way of decoding a string as JSON.
```
SeanTAllen: What happens to its current dual, JsonPathParser? This trade one asymmetry for another?

I quite like the JsonPathParser.parse() and JsonParser.parse()

Ideally I'd rather not lose that.

I think I'd veer more towards having a JsonPrinter, perhaps as you say, exposing _JsonPrint in some fashion.

I'm not quite sure but we if did something like JsonPrinter that was exposing _JsonPrint modulo a little bit of changes, would that fill your desire?

text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,75 @@
...
+
+# How We Teach This
+
+The package documentation of the `json` package should show usage examples of these functions most prominently and then show all other components of the crate, like `JsonValue`, `JsonPath` or `JsonNav`.
+
+The intend of this primitive is to expose an easy-to-use interface for turning strings into structures JSON values and vice versa. This should answer the question of how to get from bytes or strings from the network or from files to structured JSON that can be analyzed (Pass the string to `Json.parse()`), and how to serialize Pony objects and data-structures as JSON (build a `JsonValue`, then pass it those `Json.print()`).
```
SeanTAllen: longer term, I'd rather be able to have a different method to serialize pony objects to json rather than having to round-trip through a JsonValue. I've been looking at general serialization options.

text/0000-json-encode-jsonvalue.md
jemc: I agree with Sean about this.

Conceptually I have no issue with "more than one way" of parsing, since I see one as just a "convenience wrapper" of the other - having a convenience wrapper that is simpler than the underlying mechanism is a common design pattern.

text/0000-json-encode-jsonvalue.md
jemc: Agreed with Sean.

text/0000-json-encode-jsonvalue.md
mfelsche: `JsonParser.parse` does exactly what i am looking for with decode. My suggestion is to have both decode and encode apis to be similar and expectable from a users point of view. They should follow the same pattern. 

So it is either go with something that matches the pattern already in use with `JsonParser.parse` or find something new. I was going for `Json.something` because the `JsonParser` is not actually a thing that does the parsing, but delegates to `JsonTokenParser` hence i found it weird to call that primitive a `Parser`. This is just about naming things, nothing else. In the end I am quite indifferent about under what name to expose the decoding api, but since this an RFC, i wanted to come up with a reasonable decision.

text/0000-json-encode-jsonvalue.md
mfelsche: yes

text/0000-json-encode-jsonvalue.md
mfelsche: We can totally go for `JsonParser.parse()` and `JsonPrinter.print()`. I don't prefer this choice, but I would go for it, if this is the preferred way. My main pain point is to have this functionality exposed.

And yes, the `encoding` part should expose the functionality that is already there in `_JsonPrint`, as shown in the first code example of the RFC.

text/0000-json-encode-jsonvalue.md
mfelsche: This RFC is not about the longer term or a generalized serialization framework or whatnot, but about exposing JSON encoding from `JsonValue` to `String` right now. We can have the long-term discussion in another RFC or at another place any time.

The paragraph above is about how the usage of the current json package (modulo the change of this RFC) is to be documented.

SeanTAllen:
@mfelsche our reading of this is that you would be good with updating to our preferred approach. is that a reasonable reading?

mfelsche:
@SeanTAllen Could you summarize, what your preferred approach is? Is it using `JsonPrinter.print()` and `JsonParser.parse()`? As long as it includes exposing a means to turn any `JsonValue` into a JSON `String`, I am up for it.

What is your preference about removing `JsonObject.string()` and `JsonArray.string()` as requested in the RFC? I would vote for it, as it might be confusing.

SeanTAllen:
To make things more concrete. I am imagining https://github.com/ponylang/ponyc/pull/5084/changes

mfelsche:
I am going to rephrase this RFC to add a JsonPrinter as per your draft PR and rename `JsonArray.string()`, `JsonObject.string()` to `print()` and `pretty_string()` to `pretty_print()`, in order to stay consistent with the terms used for serialization throughout the package. 

Side-effect of this change will be that `JsonArray` and `JsonObject` and thus `JsonValue` no longer implement `Stringable`.

text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,69 @@
...
+# How We Teach This
+
+The package documentation of the `json` package should show usage examples of the new `JsonPrinter` functions as prominently as `JsonParser` usage and then show all other components of the crate, like `JsonValue`, `JsonPath` or `JsonNav`.
+
+The intend of this primitive is to expose an easy-to-use interface for turning JSON values into strings as a dual to `JsonParser`. This should also answer the question of how to serialize Pony objects and data-structures as JSON (build a `JsonValue`, then pass it to `JsonPrinter.print(value)`).
+
+
+# How We Test This
+
+Additional property-based tests will be added for encoding arbitrary [JsonValue][JsonValue]s via `JsonPrinter.print(...)` and parsing them again via `JsonParser.parse(...)` and ensuring equivalence of the initial values and the ones produces by the roundtrip.
```
SeanTAllen: what additional tests? this is loose enough that it would be hard to know when the RFC has implemented this. Do you have any thoughts here? I'd like to get a "bare minimum" into the RFC.

text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Drawbacks
+
+Renaming `.string()` and `.string_pretty()` on `JsonObject` and `JsonArray` to `.print()` and `.pretty_print()` is a breaking change. It will cause libraries and applications using the `json` package to be changed in order to be compiled with the ponyc version containing this change.
```
SeanTAllen, line 53: Given the early age, I'm not worried about this being a breaking change.

text/0000-json-encode-jsonvalue.md
```diff
@@ -68,8 +63,7 @@ Function naming alternatives:
 
 # Unresolved questions
 
-- Should a partial version of parsing be also exposed?
-- Should `Json.print()` expose finer controls than just the `pretty` flag?
+- Should the options, like pretty-printing or not, be exposed as optional parameters to `JsonPrinter.print()` instead of exposing `.print()` and `pretty()`?
```
jemc: Can we anticipate what other options we may want to expose?

The main one that comes to mind for me is to add configurable indentation amount/style (i.e. spaces vs tabs) to pretty-printing (and to document what the default indentation amount/style is)

Are there other options we'd want? We probably want to try to consider them for design in this RFC rather than adding more potentially-breaking changes later.

text/0000-json-encode-jsonvalue.md
```diff
@@ -68,8 +63,7 @@ Function naming alternatives:
```
jemc, line 63: Sean got confused by this section in the sync call. Adding the below would clarify the confusion point.

```suggestion

Note that this RFC only includes `print` in scope, so renaming the `parse` would be out of scope unless we decided to pursue one of the above alternatives for a broader change.

```

text/0000-json-encode-jsonvalue.md
```diff
@@ -0,0 +1,69 @@
...
+- `encode` / `decode` - this pair would make sense if the decoding would not expose `JsonParseError`. It feels inconsistent to create a `ParseError` from a `decode` function call.
+- `serialize` / `deserialize` - these terms are not used anywhere yet in pony, so didn't favor them.
+- `to_string` / `from_string` - The most humble variant of them all, getting along without any fancy terminology.
+
+# Unresolved questions
+
+- Should the options, like pretty-printing or not, be exposed as optional parameters to `JsonPrinter.print()` instead of exposing `.print()` and `pretty()`?
+
+[json-package]: https://github.com/ponylang/ponyc/tree/main/packages/json
+[JsonValue]: https://github.com/ponylang/ponyc/blob/main/packages/json/json.pony#L243
```
SeanTAllen, line 87: Would the optional parameter be an indent? IE if you give an indentation it pretty prints?

SeanTAllen:
@mfelsche I'm liking this. I have a couple questions but definitely liking. Thanks for your time on this.

text/0000-json-encode-jsonvalue.md
mfelsche: I just saw that we already have such a test here: https://github.com/ponylang/ponyc/blob/main/packages/json/_test.pony#L191

So this test just needs to be updated to use the new public api and we need some more tests verifying `JsonValue` printing produces correct results i.e. that we don't have a common bug between encoding and decoding which makes the roundtrip tests work but actually produces invalid json (e.g. bool, ints, various floats, strings including escape sequences, such tests for arrays and objects already exist).

text/0000-json-encode-jsonvalue.md
mfelsche: While there is not much options to add to `.print()`, as its behaviour is to not insert any indents, spaces or newlines whatsoever.

The `.pretty()` path is adding a configurable indent. For each level of nesting, the provided indent string is repeated times the current level. Additionally, linebreaks are inserted, after each object or array member.
This could also be controlled via an exposed option like `add_newlines: Bool = true`, if anyone would want pretty json on a single line. Also spacings are added after each object members delimiter `:`.

We could distill the difference between `.pretty()` and `.print()` into three options:

- add_newlines: after the opening `[`/`{` and after each object and array member
- add_spacings: adds a whitespace after each object member delemiter `:` 
- add_indent: adds indent based on the level of nesting

with these one could model `.print()` as a new common printing function:

```pony
JsonPrinter.print_with_options(value where add_newlines=false, add_spacings=false, indent="")
```

and `.pretty()` could be expressed as:

```pony
// indent is two spaces
JsonPrinter.print_with_options(value where add_newlines=true, add_spacings=true, indent="  ")
```

So with this we could have `JsonPrinter.print_with_options()` as above and `.print()` and `.pretty()` both forward to .`print_with_options()` with the above arguments added, without exposing them. I do like this approach, as `.pretty()` communicates the intent of getting a nice readable json and `.print()` is for compact single-line json. This works without knowing any internal knobs and whistles of the concrete implementation. For anyone who wants to tweak how the output should look like in detail, `JsonPrinter.print_with_options()` is the way to go.

If we'd say we condense it all down to just one `JsonPrinter.print()` function with the above arguments, we make it harder for anyone who just wants readable json.

So, if we would consider all possible knobs in the current implementation, my suggestion would now look lile this:

```pony
primitive JsonPrinter
  """
  Serialize any `JsonValue` to a JSON string.
  """
  fun print(value: JsonValue): String iso^ =>
    """Compact JSON serialization of any `JsonValue`."""
    JsonPrinter.print_with_options(value where indent = "", add_newlines = false, add_spaces = false)

  fun pretty(value: JsonValue, indent: String = "  "): String iso^ =>
    """Pretty-printed JSON serialization of any `JsonValue`."""
    JsonPrinter.print_with_options(value, indent where add_newlines = true, add_spaces = true)

  fun print_with_options(value: JsonValue, indent: String = "", add_newlines: Bool = false, add_spaces: Bool = false): String iso^ =>
    """JSON serialization with all possible formatting options exposed. Defaults to compact formatting."""
    // TBD
```
But this would require a slight rework of the current `_JsonPrint` mechanics.

mfelsche: After considering it all, I would argue that `.print(value)` and `.pretty(value, indent: String = "  ")` are sufficient for almost all situations, so the changes suggested above are most likely overkill and should'nt be considered, imho.

text/0000-json-encode-jsonvalue.md
mfelsche: There are several options, one I explored in this comment: https://github.com/ponylang/rfcs/pull/223#discussion_r3178807420

I still like exposing `.print()` and `.pretty()` the most. These are rather alternatives to be considered, than open questions. I will rephrase.

SeanTAllen:
@mfelsche is this ready for another round of review? it appears so. "just checking".

mfelsche:
@SeanTAllen yes, please. Lets move this along.

SeanTAllen:
I have some slight hesitation around not being stringable. Slight. I would be willing to accept this RFC. And I am definitely ok with this going to final comment period.

jemc:
To that point, I'd be in favor of having a `string` method that calls `print` internally, to match the `Stringable` interface. But not a hard requirement, and I'd approve the RFC either away.

SeanTAllen:
I'm in favor.

jemc:
:+1:

redvers:
thumbs up

PR #224, opened 2026-03-17
Add --shuffle option to PonyTest

SeanTAllen:
Proposes a `--shuffle` flag for PonyTest that randomizes test execution order to detect hidden test coupling. Accepts an optional U64 seed for reproducible orderings.

Applies to all scheduling modes: sequential (the primary use case for CI environments already using `--sequential`), concurrent (randomizes dispatch order), and within exclusion groups.

Uses `Time.cycles()` for seed generation and the existing `Random.shuffle` (Fisher-Yates) for the shuffle itself. Implementation buffers `(_TestRunner, _Group)` pairs and shuffles before dispatch, requiring no changes to the group implementations.

### DIFF
--- a/text/0000-shuffle-test-ordering.md
+++ b/text/0000-shuffle-test-ordering.md
@@ -1,0 +1,172 @@
+- Feature Name: shuffle_test_ordering
+- Start Date: 2026-03-16
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


jemc (approved):

SeanTAllen:
This was accepted. I will merge and do the shuffle to get it into the stdlib.

PR #225, opened 2026-03-27
Remove serialise package from stdlib

SeanTAllen:
The serialise package is a security footgun that's only safe with fully trusted data. Deserializing untrusted data can crash the program or give hostile code unauthorized access to the machine.

It also depends on runtime code that uses pony_error, which is being removed (ponylang/ponyc#5002). Updating serialise to work without pony_error is an unknown amount of effort, and given the security issues, that effort isn't justified.


jemc (approved):

SeanTAllen:
Approved during sync.

PR #226, opened 2026-04-04
Explain the \do_not_use\ source code annotation

pmetras:
The `\do_not_use\` Pony compiler annotation is used to output a warning at compilation-time to the developer that she is using a feature with unpredictable results. She should check the code...

### DIFF
--- a/text/0000-do-not-use-annotation.md
+++ b/text/0000-do-not-use-annotation.md
@@ -1,0 +1,84 @@
+- Feature Name: `\do_not_use\` annotation
+- Start Date: 2026-04-03
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Make the compiler emit a "Do not use!" warning when compiling code with the `\do_not_use\` annotation.
+
+# Motivation
+
+Sometimes, implementing a trait or interface can only be done partially: some methods don't make sense. For example, in an infinite-precision integer class that implement `SignedInteger` trait, the methods `min_value` and `max_value` have no sense: there is no minimal or maximal values possible, because the trait contract applies only to fixed-width integers. But the `SignedInteger` trait features cover 95% of the infinite-precision implementation, and it isn't worth adding a new type like `SignedInfiniteInteger` to cover these cases. Reusing an existing type makes the class more useful to client programmers that can substitute it to existing types with limited changes in their code.
+
+# Detailed design
+
+When the Pony compiler encounters a `\do_not_use\` annotation in the code, it emits a "Do not use" warning message at compilation time with information (source file, line number, method or other [contexts](https://tutorial.ponylang.io/appendices/annotations#what-can-be-annotated) where annotations can be used) where this information has been encountered. The programmer can see that her code uses features that must not be used because results are impredictable.
+
+Similarly, the compiler or pony-doc must injects the warning message into the documentation of the feature. The text of the message in the documentation must warn the reader not to use the feature, perhaps with an icon or a visual hint like found in othe languages documentation (Random example from https://doc.rust-lang.org/stable/std/primitive.u128.html#method.bit_width).
+
+## Where can it be used?
+
+Usually this annotation only makes sense when applied to global symbols like:
+
+* `actor`
+* `class`
+* `struct`
+* `primitive`
+* `trait`
+* `interface`
+* `new`
+* `fun`
+* `be`
+
+But it must also be accepted in other contexts like `if` or `match` because it can occur in platform-dependant code (i.e. within `ifdef` blocks). 
+
+## Example
+
+To continue with the infinite-precision integer example and the `SignedInteger` trait. Here are methods that must be annotated with `\do_not_use\`:
+
+* `min_value`: There is no minimal value in an infinite integer, as we can't represent -∞ with `SignedInteger` trait.
+* `max_value`: Similarly, there is no maximal value and +∞ can't be represented as we can always create a new value greater than any existing instance.
+* `clz`: No way to count the number of leading `0` bits. Only valid for fixed-size integer representations.
+* `clz_unsafe`: Idem.
+* `rotl` and `rotr`: Rotating bits can't happen if the width of the integer representation is not fixed.
+* `bit_reverse` and `bswap`: These methods could be implemented with an infinite-precision integer representation, but their meaning is not clear and results probably would surprise the user. 
+
+The remaining >100 methods of `SignedInteger` have a valid implementation in this variable-width integer class. More than 95% of the trait featurures have sense while a few methods produce results that can surprise the programmer who uses them.
+
+# How We Teach This
+
+This new compiler annotation will be added to the [list of recognized annotations](https://tutorial.ponylang.io/appendices/annotations#annotations-in-the-pony-compiler) with explanations on what to do when encountered.
+
+It can be mentioned in the tutorial, in a place discussing compiler error messages (to be written?). It is not mandatory as it is an advanced feature, but useful for completeness of the documentation. 
+
+# How We Test This
+
+Tests of the `\do_not_use\` annotations are similar to those of the other annotations. Minimally, we must test that a warning message is output during the compilation when the annotation is present in the source.
+
+# Drawbacks
+
+Not implementing such feature, a client programmer could use API that is not *reliable* and get surprising/not-expected results. In absence of such annotation, client can create bugs when they use these API. This annotation is used to prevent programmers from using these API inadvertently.
+
+# Alternatives
+
+## Poor man alternative
+
+Annotations are implementation-specific hints to the compiler. One can run `grep` in the `Makefile` and print the warnings if this annotation is found in the source files. The problem with such solution is that you get the warning message as soon as you compile a source where the `\do_not_use\` annotation is present, even if you don't use a forbidden method. In our infinite-precision integer example, as soon as a programmer uses this class, the warning is always issued. Vague warnings are not useful.
+
+## Change types ontology
+
+Taking again the example of the infinite-precision integer, one can change the types structure. Like we've seen, the variable-size integer can't implement some methods of the `SignedInteger` trait. We can define new traits like `VariableWidthInteger` and `FixedWidthInteger` and change the type dependencies.
+
+```pony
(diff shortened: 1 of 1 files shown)


text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Make the compiler emit a "Do not use!" warning when compiling code with the `\do_not_use\` annotation.
+
+# Motivation
+
+Sometimes, implementing a trait or interface can only be done partially: some methods don't make sense. For example, in an infinite-precision integer class that implement `SignedInteger` trait, the methods `min_value` and `max_value` have no sense: there is no minimal or maximal values possible, because the trait contract applies only to fixed-width integers. But the `SignedInteger` trait features cover 95% of the infinite-precision implementation, and it isn't worth adding a new type like `SignedInfiniteInteger` to cover these cases. Reusing an existing type makes the class more useful to client programmers that can substitute it to existing types with limited changes in their code.
```
jemc, line 12: I personally don't find this to be a compelling motivation for this RFC.

It seems like a narrow use case for what would be a broad feature, and furthermore, I think this particular use case would be better solved by either creating a new trait (which has downsides as you mentioned, but is reasonable), or implementing the methods on the trait (as described in my other advice below).

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+* `fun`
+* `be`
+
+But it must also be accepted in other contexts like `if` or `match` because it can occur in platform-dependant code (i.e. within `ifdef` blocks). 
+
+## Example
+
+To continue with the infinite-precision integer example and the `SignedInteger` trait. Here are methods that must be annotated with `\do_not_use\`:
+
+* `min_value`: There is no minimal value in an infinite integer, as we can't represent -∞ with `SignedInteger` trait.
```
jemc, line 40: Not related to this RFC's proposal, but to comment on what a reasonable approach would be for your specific use case:

---

The minimum value of an infinite signed integer would be negative infinity.

I would expect any infinite-precision integer class to be able to represent both positive and negative infinity - instances of the class that would always compare as greater or less than (respectively) every other value.

We have this for floating point values using a particular bit pattern, so I see no reason why an arbitrary-precision integer library couldn't include an inner flag which marks a value as infinite, and handle it appropriately throughout the implemented methods, including this one.

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+* `max_value`: Similarly, there is no maximal value and +∞ can't be represented as we can always create a new value greater than any existing instance.
```
jemc, line 41: Same as above, I'd expect an "infinite signed integer" library to be able to represent an infinite value.

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+* `clz`: No way to count the number of leading `0` bits. Only valid for fixed-size integer representations.
+* `clz_unsafe`: Idem.
```
jemc, line 43: If I called `clz` on a variable-width integer implementation, I'd expect the number of leading zero bits reported to be zero, since conceptually there are never any leading bits.

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+* `rotl` and `rotr`: Rotating bits can't happen if the width of the integer representation is not fixed.
+* `bit_reverse` and `bswap`: These methods could be implemented with an infinite-precision integer representation, but their meaning is not clear and results probably would surprise the user.
```
jemc, line 45: It's possible to implement rotation and reversal among the bits that exist in the particular variable-bit-width instance you're calling it on.

Again, I'd expect these both to treat the bit width as being the current bit width of the current value, with a mental model of having no leading zero bits, conceptually.

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+# Motivation
+
+Sometimes, implementing a trait or interface can only be done partially: some methods don't make sense. For example, in an infinite-precision integer class that implement `SignedInteger` trait, the methods `min_value` and `max_value` have no sense: there is no minimal or maximal values possible, because the trait contract applies only to fixed-width integers. But the `SignedInteger` trait features cover 95% of the infinite-precision implementation, and it isn't worth adding a new type like `SignedInfiniteInteger` to cover these cases. Reusing an existing type makes the class more useful to client programmers that can substitute it to existing types with limited changes in their code.
+
+# Detailed design
+
+When the Pony compiler encounters a `\do_not_use\` annotation in the code, it emits a "Do not use" warning message at compilation time with information (source file, line number, method or other [contexts](https://tutorial.ponylang.io/appendices/annotations#what-can-be-annotated) where annotations can be used) where this information has been encountered. The programmer can see that her code uses features that must not be used because results are impredictable.
```
jemc, line 16: This doesn't give sufficient detail to describe what it means "that her code uses features". What does "use" mean here?

Need a pedantic level of detail on what cases would trigger a warning to judge the merits of an RFC like this.

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
+- Feature Name: `\do_not_use\` annotation
+- Start Date: 2026-04-03
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Make the compiler emit a "Do not use!" warning when compiling code with the `\do_not_use\` annotation.
```
jemc, line 8: I think if we were to add this kind of advisory annotation we would also want to take a step back and consider the broader set of use cases (e.g. things like `deprecated`, and cases I'm not thinking of off the top of my head) rather than accept just this one annotation in the RFC - thinking of the bigger picture would inform a more coherent overall design and user experience.

pmetras: I considered `\deprecated\` too, but I thought it was too much specialized. And what is missing to compiler annotations is the ability to have parameters. I would like to explain why a method is deprecated and which other method of the API to use instead.
The reason for `\do_not_use\` was to prevent the user with a minimal warning. And let the user find by herself what is the reason in the documentation.

text/0000-do-not-use-annotation.md
pmetras: I've tried to find an example on public code.

text/0000-do-not-use-annotation.md
pmetras: Please let me disagree. No infinite precision library that I know of (GMP https://gmplib.org/manual/Integer-Logic-and-Bit-Fiddling, num https://docs.rs/num-integer/0.1.46/num_integer/trait.Integer.html or https://en.wikipedia.org/wiki/List_of_arbitrary-precision_arithmetic_software) implement those bit operations nor the infinity on integers. One of the reasons is because the notions make sense only on fixed-width 2-complement representation, and most of the time arbitrary precision libraries don't use this internal representation for arbitrary precision integers.

text/0000-do-not-use-annotation.md
```diff
@@ -0,0 +1,84 @@
...
+
+```pony
+trait VariableWidthInteger is (FixedWidthIntger & BitOperable)
+    ...
+
+type SignedInteger is (FixedWidthInteger | VariableWidthInteger)
+```
+
+Building a correct types ontology is not easy. This would render the base types of the stdlib more complex to users. Additional complexity is not a good thing.
+
```
pmetras, line 81: After thinking of it for a week, an particularly for the integer example, I think that the `\do_not_use\` compiler annotation is a work-around for a deeper problem. Pony's numeric types are wrongly defined.
`Integer[A]` mixes together:
* Algebraic operations — `add`, `sub`, `mul`, `div`...
* Overflow-aware operations — `addc`, `subc`, `mulc`...
* Bit manipulation — `op_and`, `op_or`, `op_xor`, `bit_reverse`, `bswap`, `popcount`, `clz`, `ctz`...

These are completely orthogonal capabilities. An arbitrary precision integer `MPInt` is an integer that wants algebra and overflow-awareness but has no meaningful `bit_reverse` or `popcount`. A hypothetical `BitMap` type wants bit manipulation but not integer algebra. Forcing all three into one trait means either every type implements irrelevant methods, or every type that wants any one concern must implement all three.

Instead of keeping the existing `Integer[A]` interface, it would be more future-proof to rebuild it based on sound mathematical concepts (mathematicians have created such classifications for the last centuries) like fields, rings, sign, order and bitwise operations.

The current hierarchy is flat and monolithic because it was designed around a closed world of machine types (`I8`…`U128`, `F32`, `F64`). Every method that is true of any of those types ended up in a shared super-trait, because there was no cost to doing so — all types need all the methods anyway.
The moment you add an unbounded type, that assumption collapses.

In the end, I think that my proposal of `\do_not_use\` was trying to solve the wrong problem: trying to have a new type fit into a badly-designed closed trait and giving a tool to the developer to signal this bad design. The correct solution is to go back to the origin of the problem and effectively correct it.

If the community is interested, I can suggest a new type hierarchy that would correct Pony's numeric types hierarchy. This redesign would separates what is genuinely universal (algebra, ordering) from what is incidental to machine-word representations (fixed width, bit twiddling, overflow, wrapping).

jemc: It sounds like you're thinking you should close this RFC and open a new one for adjusting the type hierarchy. I'm open to reviewing a PR for the type hierarchy, though I still personally believe that it would be a reasonable alternative to implement those methods for infinite-precision integers. But if the new type hierarchy makes sense and doesn't break too much existing code, I'm open to that as well.

jemc: From Adrian in the sync call today: he's interested to discuss a redesigned type hierarchy, and also thinks it would be interesting to discuss Complex number types.

jemc:
There was also some discussion in the sync call today about the "warnings" aspect of this RFC. As Red noted, this would be "breaking through a new wall", as we don't currently have any concept of warnings in the compiler.

There seemed to be some interest from the group about "deprecated" annotations, but more design is needed in how to surface those (e.g. via a lint command?) and how to silence particular warnings/lint violations to clear away noise to show new warnings/lint violations that appear.

PR #227, opened 2026-04-30
Withdraw RFC 53: Compile-Time Expressions

SeanTAllen:
RFC 53 was accepted in 2018 and never implemented. Reading it now, it leaves several core questions about the feature open: capability semantics, behavior of `error` inside compile-time expressions, floating point host/target semantics, and which subset of the language is actually permitted. These are not edge cases. They define what the feature does.

The idea is worth pursuing. A future RFC on this topic should pin these questions down and come with a working implementation that can be evaluated alongside the design.

If accepted, the follow-on actions are to add a withdrawal note to the top of `text/0053-compile-time-expression.md` and close ponyc#2591 with a pointer back to this RFC.

### DIFF
--- a/text/0000-withdraw-compile-time-expressions.md
+++ b/text/0000-withdraw-compile-time-expressions.md
@@ -1,0 +1,49 @@
+- Feature Name: withdraw-compile-time-expressions
+- Start Date: 2026-04-29
+- RFC PR:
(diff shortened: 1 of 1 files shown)


jemc (approved):

SeanTAllen:
Approved.

PR #229, opened 2026-05-02
Type Layout functions added to compiler to better support C-FFI.

redvers:
Proposes a TypeInfo primitive with four compile-time intrinsics — size_of, stride_of, align_of, offset_of — exposing layout information LLVM already computes. Covers concrete nominal types (primitives, structs, classes, actors) and tuples; unions and other abstract types are compile-time errors.

### DIFF
--- a/text/0000-type-layout-intrinsics.md
+++ b/text/0000-type-layout-intrinsics.md
@@ -1,0 +1,437 @@
+- Feature Name: type-layout-intrinsics
+- Start Date: 2026-05-01
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add a small, cohesive set of compiler intrinsics that expose the in-memory
+layout of types: store size, array stride, alignment, and field offsets.
+These are exposed as methods on a new `TypeInfo` primitive and are direct
+projections of information LLVM already computes during code generation.
+
+# Motivation
+
+Pony users today have no first-class way to ask basic layout questions
+about a type, oft required for C-FFI binding:
+
+- "How many bytes do I need to allocate to hold one of these?"
+- "How many bytes between consecutive elements in an array of these?"
+- "What alignment does this type require?"
+- "Where does field `_x` start within this struct?"
+
+These questions come up every time a Pony program touches raw memory:
+allocating buffers via `@pony_alloc`, packing or unpacking C-ABI
+structs across the FFI boundary, building custom serializers, implementing
+allocators, computing addresses for pointer arithmetic, or interfacing
+with hardware/protocol layouts that pin field positions.
+
+Today, users hand-compute these numbers, hard-code them, or copy them out
+of C headers. All three approaches drift silently when fields are
+reordered, types are widened, or the target ABI changes. The information
+is already known to the compiler — there is no good reason for users to
+recompute it by hand.
+
+There is also a related family of questions about deep memory accounting
+("how much memory does this `Array[String]` actually consume, transitively?").
+Those are explicitly **out of scope** for this RFC; they will be proposed
+separately as a separate interface.
+
+Intrinsics in this RFC are the primitives that such a follow-up may choose
+build upon.
+
+# Detailed design
+
+A new primitive `TypeInfo` is added to the standard library. It exposes
+four methods, all `compile_intrinsic`. Each takes the type of interest as
+a type parameter.
+
+```pony
+primitive TypeInfo
+  fun tag size_of[T](): USize        => compile_intrinsic
+  fun tag stride_of[T](): USize      => compile_intrinsic
+  fun tag align_of[T](): USize       => compile_intrinsic
+  fun tag offset_of[T](field: String): USize => compile_intrinsic
+```
+
+Each intrinsic maps to a specific LLVM query. Where two LLVM functions
+both plausibly fit, the split below picks the one whose semantics match
+the method's documented contract.
+
+## `size_of[T]()`
+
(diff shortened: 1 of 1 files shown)


text/0000-type-layout-intrinsics.md
```diff
@@ -0,0 +1,437 @@
...
+covering:
+
+- Numeric primitives (`U8`, `U32`, `F64`, `I128`) — exact known sizes.
+- Structs with mixed-alignment fields where `size_of` and `stride_of`
+  must differ.
+- Classes including ones with `embed`ded fields.
+- `offset_of` for every field of a representative struct, asserting
+  that consecutive offsets are non-overlapping and respect alignment.
+- `offset_of` with an invalid field name — must produce a compile
+  error, not a runtime failure.
```
jemc: This isn't really possible, generally. You could pass any dynamic string.

text/0000-type-layout-intrinsics.md
```diff
@@ -0,0 +1,437 @@
...
+A new primitive `TypeInfo` is added to the standard library. It exposes
+four methods, all `compile_intrinsic`. Each takes the type of interest as
+a type parameter.
+
+```pony
+primitive TypeInfo
+  fun tag size_of[T](): USize        => compile_intrinsic
+  fun tag stride_of[T](): USize      => compile_intrinsic
+  fun tag align_of[T](): USize       => compile_intrinsic
+  fun tag offset_of[T](field: String): USize => compile_intrinsic
```
jemc: We discussed on the sync call. The "field string must be checked at compile time" angle here causes problems, but an alternative design proposed by Adrian was to make this return an array of tuples with name, offset pairs.

text/0000-type-layout-intrinsics.md
```diff
@@ -0,0 +1,437 @@
+- Feature Name: type-layout-intrinsics
+- Start Date: 2026-05-01
+- RFC PR: (leave this empty)
+- Pony Issue: (leave this empty)
+
+# Summary
+
+Add a small, cohesive set of compiler intrinsics that expose the in-memory
+layout of types: store size, array stride, alignment, and field offsets.
+These are exposed as methods on a new `TypeInfo` primitive and are direct
```
jemc: It's important to get this name specific enough to avoid name collisions, since this is proposed to go in `builtin`, so it will risk collisions with every single Pony source file.

Name suggestions from the sync call:
- `ABITypeInfo` (suggested by Joe)
- `MemoryLayout` (suggested by Adrian. motivated by Rust/Swift comparison)
- `LLVMMemoryLayout` (suggested by Adrian)
- `ABIMemoryLayout` (suggested by Joe, agreed to by Adrian)

We also discussed that Red is looking to add other type introspection things, which is why he originally called it `TypeInfo`

SeanTAllen: is "type introspection things" a baby-step to "type-level programming"?

text/0000-type-layout-intrinsics.md
```diff
@@ -0,0 +1,437 @@
...
+
+Not implementing this RFC leaves Pony users hand-computing layout
+numbers and copying them from C headers. The code keeps working until
+something gets reordered or the target ABI changes, at which point it
+silently breaks.
+
+# Unresolved questions
+
+- **Should `offset_of` accept nested paths** like
+  `"outer_field.inner_field"` in this RFC, or punt to a follow-up?
```
jemc: In the sync call we said we should exclude any special handling of embedded fields.

text/0000-type-layout-intrinsics.md
```diff
@@ -0,0 +1,437 @@
...
+let p_i = base.offset(TypeInfo.stride_of[Entry]() * i)
+```
+
+Sizing the backing allocation for an array-style data structure:
+
+```pony
+let bytes_needed = TypeInfo.stride_of[T]() * capacity
+```
+
+Round-tripping through a C API that takes `(void*, element_size, count)`:
```
adrianboyko: Not sure what "round tripping" means here. The example is about a C function that takes a pointer, and element size, and a count and then presumably performs some function on all the elements. That doesn't imply any "round trip".

text/0000-type-layout-intrinsics.md
redvers: We could limit it to string literals, but arguably we would be creating a new class of check in the compiler.  So, let's do the Array[(String, USize)] val offsets_of[T]()

text/0000-type-layout-intrinsics.md
redvers: Done

text/0000-type-layout-intrinsics.md
redvers: Renamed.

@SeanTAllen - I don't think so. I do have some plans that have the functionality in this RFC as a requirement, but nothing that I would descrive as "type-level programming".

text/0000-type-layout-intrinsics.md
redvers: Done

text/0000-type-layout-intrinsics.md
redvers: done

SeanTAllen:
@redvers please ask a robot to reformat this to have no breaking lines except for paragraphs etc. or suffer and do it yourself. i think you should "ask a robot".

redvers:
This rfc will likely become redundant when the shim compiler is introduced.  Consider this RFC on hold until that merges and I can verify with the other interested parties.

PR #235, opened 2026-07-04
Add noalloc annotation RFC

SeanTAllen:
This RFC proposes `noalloc`, a compile-time annotation that verifies a method, and everything it transitively calls, performs no heap allocation. You mark a hot method; in release builds the compiler fails the build if anything under it allocates, including code you didn't write and didn't annotate. The check runs after optimization on the shipped binary, so whatever stays on the stack passes. It also goes on interface and trait methods, where it forces every implementer to be `noalloc`, which is what lets a `noalloc` method dispatch through an interface.

The open questions are about the compiler error: whether it can point at the specific allocation, and what it reports when more than one allocation caused the failure.

### DIFF
--- a/text/0000-noalloc.md
+++ b/text/0000-noalloc.md
@@ -1,0 +1,183 @@
+- Feature Name: noalloc
+- Start Date: 2026-07-03
+- RFC PR:
+- Pony Issue:
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
we discussed. 

allow noalloc on trait methods. disallow on interface.

like the no supertype annotation's interaction with the type system, we can require any trait method that is \noalloc\ to require its implementers to have \noalloc\ on the same methods. however it isn't required. that's an open question.

at the time that we are doing our check for allocations, we can see all trait possibilities at virtual dispatch call sites and verify that none will allocate because by the time we are past heap2stack and optimizations, we have a fully closed world.

for interfaces, we have to use the same closed world at any virtual dispatch site to see if any implementers would alloc and if yes, disallow.

